PR #48, opened 2014-05-05
Support for graph output on console/shell

paulegan:
This patch adds support for `cpugraph`, `memgraph`, etc to `out_to_console` and `out_to_ncurses`. This is particularly useful when used in conjunction with a text-only status bar common on tiling window managers (e.g. i3bar, dzen2, xmobar).

### DIFF
--- a/doc/config_settings.xml
+++ b/doc/config_settings.xml
@@ -75,6 +75,17 @@
     <varlistentry>
         <term>
             <command>
+                <option>console_graph_ticks</option>
+            </command>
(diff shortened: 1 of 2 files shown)


paulegan:
@marcpayne Sorry for the slow response to this. I didn't see the issue with my configuration but I've just submitted a pull request that makes the documentation encoding explicit.

marcpayne:
@paulegan No problem, and thanks for the fix!

PR #81, opened 2015-02-19
Specify diskio by device label

troxor:
This might be helpful for someone. Block device labels are provided in /dev/disk/by-label/, and can be more reliable than specifying the actual device name.


brndnmtthws:
Has it been tested with and without the labels?

troxor:
Yep, it just resolves the label to a device, and passes that device along to the rest of the module. I've been using this patch myself for a while now, and haven't noticed any terrible behavior so far. Any suggestions/critiques are welcome.

PR #82, opened 2015-02-27
Freetype include path

troxor:
According to comments in /usr/share/cmake-3.1/Modules/FindFreetype.cmake, recent versions of FreeType use some #include trickery that complicate finding headers. The proper way to do this is to use the `freetype-config` application or a recent version of upstream CMake's FindFreetype.cmake module, but this patch worked on all the distros I tested it on (Gentoo, Archlinux, Ubuntu 14.04, Ubuntu 12.04, Fedora 20).


brndnmtthws:
Thanks!

PR #83, opened 2015-03-08
Fix segfault on invalid label

troxor:
when an invalid disk label is provided, realpath returns NULL. This checks for that and keeps conky from segfaulting reporting the following:

`conky: diskio device 'label:aoeu' does not exist`


brndnmtthws:
Thanks!

PR #96, opened 2015-07-01
Avoid clobbering CFLAGS/CXXFLAGS set by user

Vincent-C:
This allows the end user to specify CFLAGS or CXXFLAGS as desired (e.g. `cmake -DCMAKE_C_FLAGS=-foo -bar`); an use case for this would be for package maintainers whose distribution mandates the use of build hardening flags (e.g. Debian/Ubuntu). Thanks for considering this PR!

(Also see http://voices.canonical.com/jussi.pakkanen/2013/03/26/a-list-of-common-cmake-antipatterns/)


brndnmtthws:
Looks great, thanks!

PR #104, opened 2015-07-02
Fix battery

Stebalien:
The real bug fix is the last commit (`battery_short` was displaying `Uknown 99` instead of `U 99`). The other two changes just avoid having different names for the same states.


brndnmtthws:
Looks good, thanks!

Issue #105, opened 2015-07-03
Segfaults if ipv6 switched off at kernel level


mcfadzgit:
Using the conky-1.10.0-2-x86_64.pkg.tar.xz package from Arch's Extra repo.
I have ipv6 switched off at the kernel level by using the parameter: `ipv6.disable=1`
Conky segfaults immediately, regardless of configuration.
Can only be solved by modifying PKGBUILD cmake command to include:
`-D BUILD_IPV6=OFF \`
makepkg can then build the package and once installed, Conky runs as expected.

marcpayne:
I've tracked down the cause of this segfault. Until I figure out how to use git without breaking anything, I'll just describe the fix. In `linux.cc:647`, you call `fclose(file)`, but `file` might be `NULL`. Moving the `fclose()` call up into the `if` statement fixes the problem for me.

marcpayne:
@mcfadzgit Fixed in 1.10.1, so this issue can be closed now.

PR #107, opened 2015-07-04
Move fclose(file) since file may be NULL

marcpayne:
When IPV6 support is enabled in conky, but the user has explicitly used the `ipv6.disable=1` kernel parameter, conky will segfault. The code tries to open `/proc/net/if_inet6` into `file`, but this fails. Subsequently, `fclose(file)` fails. This commit simply moves the `fclose()` call up into the `if` statement where it belongs. See #105.

### DIFF
--- a/doc/config_settings.xml
+++ b/doc/config_settings.xml
@@ -80,7 +80,7 @@
         </term>
         <listitem>A comma-separated list of strings to use as the bars of a graph output
         to console/shell. The first list item is used for the minimum bar height and the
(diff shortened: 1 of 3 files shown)


marcpayne:
I managed to turn this simple pull request into a mess with my ignorance of git. Next time, I'll make my changes in a separate branch (not master!) and submit a pull request from that.

marcpayne:
Closing for now. I'd rather have one clean commit for you to merge.

PR #108, opened 2015-07-06
Close file only if it was successfully opened

marcpayne:
If BUILD_IPV6=ON (default), but the user has disabled ipv6 support
in the kernel using the parameter ipv6.disable=1, then conky fails
to open /proc/net/if_inet6. This leads to a segfault since conky
calls fclose(file) regardless. This fix simply moves the fclose call
into the preceding if statement. See #105.


brndnmtthws:
Awesome, thanks!

Issue #93, opened 2015-06-30
conky: obj->data.i 1 info.cpu_count 0


rtoijala:
I am using the following conkyrc:

````lua
conky.config = {
	lua_load = 'test.lua',
	out_to_console = true,
	out_to_x = false
};

conky.text = [[
${lua pad 3 ${cpu cpu1}}
]];
````

and the following lua file:

````lua
function conky_pad(str, length)
	str = tostring(conky_parse(str));
	length = tonumber(conky_parse(length));

	return string.format("%"..length.."s", str);
end
````

Conky exits with the following output:

````
$ conky -DD -c conkyrc
DEBUG(0) [/build/conky/src/conky-1.10.0/src/conky.cc:2671]: reading contents from config file 'conkyrc'
DEBUG(1) [/build/conky/src/conky-1.10.0/src/core.cc:1913]: no templates to replace
DEBUG(1) [/build/conky/src/conky-1.10.0/src/core.cc:1913]: no templates to replace
DEBUG(1) [/build/conky/src/conky-1.10.0/src/conky.cc:886]: evaluated '3' to '3'
DEBUG(1) [/build/conky/src/conky-1.10.0/src/core.cc:1913]: no templates to replace
DEBUG(1) [/build/conky/src/conky-1.10.0/src/core.cc:576]: Adding $cpu for CPU 1
conky: obj->data.i 1 info.cpu_count 0
conky: attempting to use more CPUs than you have!
````

Changing the line `${lua pad 3 ${cpu cpu1}}` to simply `${cpu cpu1}` fixes the problem.
`/proc/stat` shows 4 cpus.

Running `gdb --args conky -DD -c conkyrc` and adding a breakpoint at `get_cpu_count` reveals that the function is never called (`strace` also shows that `/proc/stat` is never opened).

rtoijala:
Fix merged long ago, closing.

PR #111, opened 2015-07-10
Call get_cpu_count() when needed

rtoijala:
get_cpu_count() is not called before trying to use cpus so conky exits with

conky: obj->data.i 1 info.cpu_count 0
conky: attempting to use more CPUs than you have!

Fixes issue #93.


brndnmtthws:
Thanks!

PR #115, opened 2015-07-21
Add top_name_verbose option

### LINKED ISSUE #113: top displays full command
Hi,

Since version 1.10.0-2, the "top" variable in Conky displays full commands names. Before, if Deluge was running for example, I had the deluged process written like "deluged".

Now, I have "/usr/bin/python2 /usr/bin/deluged --port=54566 --config=/home/me/.config/deluge".

I think it's too verbose.

Do you know how to come back to the last format ? Downgrading to 1.9 branch solved the issue.

Sincerely

marcpayne:
Since commit 749083a, the output of `${top name <num>}` contains the full command line of each process, including arguments. While this feature can be very useful, it changes the default behavior of Conky.

The present commit adds a new `top_name_verbose` option that allows the user to toggle between basenames with no arguments (the old behavior) and full command lines with arguments. To remain consistent with past versions of Conky, the default value of `top_name_verbose` is `false`.

Now, I used the string class, rather than C str* functions. This is a personal preference, but I can change it if necessary. Also, I chose to set the default behavior back to what it was prior to version 1.10.0, which might annoy some people (again). I'm open to input.

Thanks for your consideration of this request.

Fixes #113.

### DIFF
--- a/doc/config_settings.xml
+++ b/doc/config_settings.xml
@@ -962,6 +962,17 @@
     <varlistentry>
         <term>
             <command>
+                <option>top_name_verbose</option>
+            </command>
+        </term>
+        <listitem>If true, top name shows the full command line of each
+        process, including arguments (whenever possible). Otherwise,
+        only the basename is displayed. Default value is false.
+        <para /></listitem>
+    </varlistentry>
+    <varlistentry>
+        <term>
+            <command>
                 <option>top_name_width</option>
             </command>
         </term>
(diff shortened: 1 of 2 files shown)


stardiviner:
+1

Nooby4Ever:
Since https://github.com/brndnmtthws/conky/issues/121 is very similar to this, i was wondering if it would be possible for you to make an ```if_running_any``` command which checks on the short name (as in Conky 1.9) instead of the verbose name as ```if_running``` does now?

brndnmtthws:
Looks reasonable, thanks!

Issue #97, opened 2015-07-01
"conky -i 5" has no effect (since update to 1.10.0)


fdietze:
```
conky: Invalid value of type 'table' for setting 'total_run_times'. Expected value of type 'number'.
```

achilleas-k:
Same happens for other command line arguments that accept numerical values:
```
$ conky -c ~/.conky/panel.conky -y 100
conky: Invalid value of type 'table' for setting 'gap_y'. Expected value of type 'number'.
```

The result is the conky panel being created at an arbitrary Y location (~60), regardless of specified value. If `gap_y` is specified in the config file, it behaves as expected (assuming no `-y` is specified on the command line).

marcpayne:
I can confirm this. The problem is that `config_setting_base::lua_set` in `setting.cc` seems to misinterpret what is actually on the stack. Here is a short breakdown of what is happening. I should note that I am using Lua 5.3.

Suppose we are processing `total_run_times`. On entry into `lua_set`, the stack looks like this:
```
| ... total_run_times |
```

After `l.getglobal("conky");` we have
```
| ... total_run_times  some_table  conky |
```

By hand tracing through `getglobal`, it seems that `some_table` is `registry[LUA_RIDX_GLOBALS]`. Whether or not this is the intended behavior of `getglobal`, an easy fix is to add `l.replace(-2);` after the call to `getglobal`. Doing so solves the problem for me.

marcpayne:
@fdietze Just doing some housekeeping. This issue has been fixed, so it can be closed now.

PR #119, opened 2015-07-22
Ensure getglobal pushes only one value onto stack

marcpayne:
Currently, `state::getglobal` pushes two values onto the stack: the global environment and the requested global. For example, if you call `getglobal("conky")`, the stack ends up with the following:

```
| ... global, conky |
```

The function `config_setting_base::lua_set` in `setting.cc:108` does not take this behavior into account, resulting in #97. To correct this, we call `replace(-2)` at the end of `state::getglobal` where `LUA_VERSION_NUM >= 502`.

I checked every call to `getglobal` (to the best of my abilities) to be sure there are no regressions. I used the incredibly useful [inspect.lua](https://github.com/kikito/inspect.lua) to dump the top entries of the stack after each stack-changing line in `lua_set` to confirm the fix. I'd be happy to provide before/after stack dumps, or the code I used to generate them.

**Edit:** added some details on how I debugged this.

### DIFF
--- a/src/luamm.cc
+++ b/src/luamm.cc
@@ -357,6 +357,7 @@ namespace lua {
 		pushinteger(LUA_RIDX_GLOBALS);
 		gettable(REGISTRYINDEX);
 		getfield(-1, name);
+		replace(-2);
 #else
 		getfield(LUA_GLOBALSINDEX, name);
 #endif


brndnmtthws:
Interesting, thanks for the details!

PR #122, opened 2015-07-24
Added missing build flags to the print_version function

Nooby4Ever:
Added:
BUILD_ICAL
BUILD_ICONV
BUILD_MYSQL
BUILD_LUA_RSVG
BUILD_XSHAPE

BUILD_LIBXML2 is defined in <code>/cmake/config.h.in</code> but as far as i'm aware not used any more,  i did not remove it because i wasn't sure.


brndnmtthws:
Lovely, thanks!

Issue #12, opened 2013-05-16
master branch fails when "border_inner_margin = 0, border_outer_margin = 0, and border_width = 0".


arclance:
The master branch fails with
```
conky: desktop window (201) is root window
X Error of failed request:  BadValue (integer parameter out of range for operation)
  Major opcode of failed request:  1 (X_CreateWindow)
  Value in failed request:  0x0
  Serial number of failed request:  1303
  Current serial number in output stream:  1308
```
when 
```
border_width = 0 
```
is used in conky.config.

I believe this is because it is trying to create the initial window with dimensions of 0x0 which is not allowed by X.

arclance:
I have been told by someone who uses one of my Lua scripts that this affects Conky v1.10.x (the newly released version).

Nooby4Ever:
I can confirm this still exists in the current master branch

arclance:
I probably just needs a sanity check (are we trying to make a invalid window size window?) to handle this possibility (
I would guess you would just make it use 1x1 instead of 0x0 if 0x0 is asked for.
I don't know the language so I can't do it either.

marcpayne:
@arclance Is this fixed for you in 1.10.1?

arclance:
I can't configure it to build it so I can't test it.
```
CMake Error at cmake/ConkyPlatformChecks.cmake:144 (check_include_file):
   Unknown CMake command "check_include_file".
 Call Stack (most recent call first):
   CMakeLists.txt:37 (include)
```
Judging by #169 you need some kind of fallback behavior for cmake versions older than 3.4.0.
That is currently in Experimental in Debian and I am not installing any compiler from Experimental.

arclance:
I figured out why it would not configure (missing include in "ConkyPlatformCheck.cmake") but I am now getting a linking failure (#174) so I still can't test it.

arclance:
I was able to get it to build and can confirm that this fix worked.

PR #124, opened 2015-07-25
Added sanity check to avoid making 0x0 window (resulting in Segementation Fault)

Nooby4Ever:
This fixes the bug explained in: https://github.com/brndnmtthws/conky/issues/12 as far as i know the 1x1 instead of 0x0 window has no side effects and thus fixing the issue.

### DIFF
--- a/src/x11.cc
+++ b/src/x11.cc
@@ -643,6 +643,11 @@ static void init_window(lua::state &l __attribute__((unused)), bool own)
 
 		int b = border_inner_margin.get(l) + border_width.get(l)
 			+ border_outer_margin.get(l);
+			
+		/* Sanity check to avoid making an invalid 0x0 window*/
+		if (b == 0){
+			b=1;
+		}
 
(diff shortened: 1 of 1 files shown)


src/x11.cc
```diff
@@ -643,6 +643,11 @@ static void init_window(lua::state &l __attribute__((unused)), bool own)
 
 		int b = border_inner_margin.get(l) + border_width.get(l)
 			+ border_outer_margin.get(l);
+			
+		/* Sanity check to avoid making an invalid 0x0 window*/
+		if (b == 0){
```
brndnmtthws: Minor formatting nitpick: space before the `{`.

src/x11.cc
```diff
@@ -643,6 +643,11 @@ static void init_window(lua::state &l __attribute__((unused)), bool own)
...
+			b=1;
```
brndnmtthws: Space before & after `=`.

Nooby4Ever:
Al right i will close this one and make a new one so the commit has only 1 commit message.

PR #126, opened 2015-07-31
Added sanity check to avoid making 0x0 window (resulting in Segementation Fault)

Nooby4Ever:
This fixes the bug explained in: https://github.com/brndnmtthws/conky/issues/12 as far as i know the 1x1 instead of 0x0 window has no side effects and thus fixing the issue.

With properly formatted code. (see https://github.com/brndnmtthws/conky/pull/124)


brndnmtthws:
Thanks for the patch.

Issue #100, opened 2015-07-01
Conky 1.10.0 fails to compile without these patches, please commit...


ghost:
Hi,

I found these patches from Arch, but noticed on my distro I can't compile conky also unless I use these patches;

https://projects.archlinux.org/svntogit/packages.git/tree/trunk/ascii.patch?h=packages/conky

https://projects.archlinux.org/svntogit/packages.git/tree/trunk/lua53.patch?h=packages/conky

I hope you'll please commit these, otherwise conky fails to compile on some distros.

On my system without the `ascii.patch` I'll get this error at compile time;

[  6%] Proccessing man page for docs
/usr/bin/iconv: illegal input sequence at position 7845
/usr/bin/db2x_manxml: program in pipeline exited with an error
make[2]: **\* [../doc/docs.html] Error 1
make[1]: **\* [doc/CMakeFiles/docs.dir/all] Error 2
make: **\* [all] Error 2

On my system without the `lua53.patch` I'll get this error at compile time;

[ 58%] Building CXX object src/CMakeFiles/conky.dir/luamm.cc.o
In file included from /usr/include/lua.hpp:6:0,
                 from /tmp/foo/conky-1.10.0/src/luamm.hh:30,
                 from /tmp/foo/conky-1.10.0/src/luamm.cc:24:
/tmp/foo/conky-1.10.0/src/luamm.cc: In member function ‘void lua::state::gettable(int)’:
/tmp/foo/conky-1.10.0/src/luamm.cc:370:3: error: no matches converting function ‘safe_misc_trampoline’ to type ‘lua_CFunction {aka int (_)(struct lua_State_)}’
   lua_pushcfunction(cobj.get(), (&safe_misc_trampoline<&lua_gettable, 1>));
   ^
/tmp/foo/conky-1.10.0/src/luamm.cc:168:7: note: candidate is: template<void (* misc)(lua_State*, int), int nresults> int lua::{anonymous}::safe_misc_trampoline(lua_State_)
   int safe_misc_trampoline(lua_State *l)
       ^
make[2]: *_\* [src/CMakeFiles/conky.dir/luamm.cc.o] Error 1
make[1]: **\* [src/CMakeFiles/conky.dir/all] Error 2
make: **\* [all] Error 2

marcpayne:
Fixed by #137 and #138, so this issue can be closed.

Issue #91, opened 2015-06-29
db2x_manxml fails on non-ascii chars


vesath:
In maintainer mode, db2x_manxml from docboox2x-0.8.8 fails to convert conky-1.10.0's docs.mxml to a man page because it contains non-ascii characters. In fact, db2x_manxml runs `iconv -futf8 -tascii` as a subroutine and iconv itself fails on invalid characters.

I'm not sure what should be done about this. Removing the characters (see patch below) allows the manpage to be generated.

vesath:
Here is the error message:

/usr/bin/iconv: illegal input sequence at position 7849
/usr/bin/db2x_manxml: program in pipeline exited with an error
doc/CMakeFiles/docs.dir/build.make:51: recipe for target 'doc/docs.html' failed
make[2]: *** [doc/docs.html] Error 1
CMakeFiles/Makefile2:218: recipe for target 'doc/CMakeFiles/docs.dir/all' failed

And the obvious patch:

```
diff -aur old/doc/config_settings.xml new/doc/config_settings.xml
--- old/doc/config_settings.xml	2015-06-21 14:49:06.000000000 -1000
+++ new/doc/config_settings.xml	2015-06-29 10:05:00.147318479 -1000
@@ -80,7 +80,7 @@
         </term>
         <listitem>A comma-separated list of strings to use as the bars of a graph output
         to console/shell. The first list item is used for the minimum bar height and the
-        last item is used for the maximum. Example: " ,_,▁,▂,▃,▄,▅,▆,▇,█".
+        last item is used for the maximum.
         <para /></listitem>
     </varlistentry>
     <varlistentry>
```

andreas--e:
That's more like a workaround than a real fix. Think of manuals in complex scripts like Japanese or so. These would most likely still fail because what you did is simply *remove* the non-ascii part (see "patch" snippet), but not properly *handle* it.

marcpayne:
@andreas--e The commit by @paulegan (which is the one that got merged) properly addresses the problem without removing the non-ascii characters.

andreas--e:
Whoops...looks it was late then. Thanks for the heads-up. The "that" in my statement *"That's more like a workaround than a real fix IMO"* was (obviously)  referring to vesath's patch, which suggested that the "fix" was just to remove those "blocky" characters after the *"Example:"* text. Sorry for the noise. In any case, it would be better if the person who does a proper fix also *participated* in this report, instead of doing things stealthily like a street cat. ;-)

[...]

PR #133, opened 2015-08-19
Fixed db2x_manxml failure on non-ascii chars

Nooby4Ever:
This is just a pull request with the patch mentioned in #91, i see no reason not to add it to Conky. It was also mentioned here: #100

The original author of the commit never responded to the question how he did manage to compile the documentation with the non ascii characters. Reference: https://github.com/brndnmtthws/conky/pull/48


paulegan:
I've submitted a pull request for an alternative fix at #138.

Nooby4Ever:
Since your alternative is generally better and more future proof i will close this pull request.

PR #134, opened 2015-08-19
Fixed Curl segmentation fault

Nooby4Ever:
This is a pull request form the patch mentioned in #109 the original patch is contributed by Jay: https://github.com/bagder/curl/issues/342#issuecomment-120579447

Important note from the original author: This may not be the correct solution, it depends on the author's intention. Maybe he wanted to keep the headers from the previous call (in other words if there are no new headers use the old ones) and didn't realize they weren't copied in the handle, in which case he'll have to save the headers instead of above.

This patch has been used in the arch packages for a while now and seems to fix the problem so I've made a pull request for it.


brndnmtthws:
Looks good, thanks!

Issue #116, opened 2015-07-21
Lua 5.3 support


ghost:
Is there a specific reason why there is no native Lua 5.3(.1) support in Conky?

In the arch repositories there is currently a patch to make conky compile with Lua 5.3(.1) but this does not work when enabling the Conky Lua bindings: cairo, imlib2 and rsvg. (I assume they need to be updated)

---

The patch to get Conky 1.10 to compile with Lua 5.3(.1) (no Conky Lua bindings): https://projects.archlinux.org/svntogit/packages.git/tree/trunk/lua53.patch?h=packages/conky

Edit: The Lua 5.3 patch was actually already posted here: https://github.com/brndnmtthws/conky/issues/90 and here https://github.com/brndnmtthws/conky/issues/100

---

The error i get while trying to compile Conky 1.10 (with the Lua patch, and some others) and Lua 5.3.1 with the Conky Lua bindings cairo, imlib2 and rsvg enabled:

```
Linking CXX executable conky
/usr/lib/gcc/x86_64-unknown-linux-gnu/5.1.0/../../../../lib/libtolua++.a(tolua_map.o): In function `tolua_bnd_setpeer':
(.text+0x76): undefined reference to `lua_setfenv'
/usr/lib/gcc/x86_64-unknown-linux-gnu/5.1.0/../../../../lib/libtolua++.a(tolua_map.o): In function `mapinheritance':
(.text+0x468): undefined reference to `lua_insert'
/usr/lib/gcc/x86_64-unknown-linux-gnu/5.1.0/../../../../lib/libtolua++.a(tolua_map.o): In function `mapsuper':
(.text+0x5d6): undefined reference to `lua_insert'
/usr/lib/gcc/x86_64-unknown-linux-gnu/5.1.0/../../../../lib/libtolua++.a(tolua_map.o): In function `tolua_bnd_getpeer':
(.text+0x62a): undefined reference to `lua_getfenv'
/usr/lib/gcc/x86_64-unknown-linux-gnu/5.1.0/../../../../lib/libtolua++.a(tolua_map.o): In function `tolua_dobuffer':
(.text+0x10b7): undefined reference to `luaL_loadbuffer'
/usr/lib/gcc/x86_64-unknown-linux-gnu/5.1.0/../../../../lib/libtolua++.a(tolua_map.o): In function `tolua_dobuffer':
(.text+0x10d3): undefined reference to `lua_pcall'
[...]

PR #136, opened 2015-08-20
Fix compilation on systems with Lua 5.3.0+

### LINKED ISSUE #90: lua_gettable's return type is int, not void in lua-5.3
conky-1.10.0 fails to compile against lua-5.3; the error is:

In file included from /usr/include/lua.hpp:6:0,
                 from /build/conky/src/conky-1.10.0/src/luamm.hh:30,
                 from /build/conky/src/conky-1.10.0/src/luamm.cc:24:
/build/conky/src/conky-1.10.0/src/luamm.cc: In member function 'void lua::state::gettable(int)':
/build/conky/src/conky-1.10.0/src/luamm.cc:370:3: error: no matches converting function 'safe_misc_trampoline' to type 'lua_CFunction {aka int (*)(struct lua_State*)}'
   lua_pushcfunction(cobj.get(), (&safe_misc_trampoline<&lua_gettable, 1>));
   ^
/build/conky/src/conky-1.10.0/src/luamm.cc:168:7: note: candidate is: template<void (* misc)(lua_State*, int), int nresults> int lua::{anonymous}::safe_misc_trampoline(lua_State*)
   int safe_misc_trampoline(lua_State *l)
       ^
src/CMakeFiles/conky.dir/build.make:675: recipe for target 'src/CMakeFiles/conky.dir/luamm.cc.o' failed

It turns out lua_gettable returns an int in lua-5.3. The attached patch fixes this issue, though as currently written it is specific to lua-5.3.

marcpayne:
Since Lua 5.3.0, the return type of `lua_gettable` and similar functions is int, rather than void. Thus, `safe_misc_trampoline<&lua_gettable, 1>` does not compile (the first template parameter is a pointer to function with return type void). A solution was provided by vesath in issue #90, which I modified slightly with Lua version checks. It compiles and works on my system, but I can't test with older Lua versions. If travis can compile it, we should be good.

Fixes #90 and half of #100.

### DIFF
--- a/src/luamm.cc
+++ b/src/luamm.cc
@@ -165,11 +165,28 @@ namespace lua {
 		}
 
 		template<void (*misc)(lua_State *, int), int nresults>
+#if LUA_VERSION_NUM >= 503
+		// Since Lua 5.3.0, the return type of lua_gettable and similar functions
+		// is int, so we need to handle get and set variants separately.
+		int safe_misc_trampoline_set(lua_State *l)
+		{
+			misc(l, 1);
+			return nresults;
+		}
+
+		template<int (*misc)(lua_State *, int), int nresults>
+		int safe_misc_trampoline_get(lua_State *l)
+		{
+			misc(l, 1);
+			return nresults;
+		}
+#else
 		int safe_misc_trampoline(lua_State *l)
 		{
 			misc(l, 1);
 			return nresults;
 		}
+#endif
 
 		int safe_next_trampoline(lua_State *l)
 		{
@@ -326,7 +343,7 @@ namespace lua {
 		if( rawequal(index1, index2) )
 			return true;
 
-		return safe_compare(&safe_compare_trampoline<lua_equal>, index1, index2);
+		return safe_compare(&safe_compare_trampoline<&lua_equal>, index1, index2);
 	}
 
 	int state::gc(int what, int data)
@@ -368,7 +385,11 @@ namespace lua {
 		checkstack(2);
 		pushvalue(index);
 		insert(-2);
+#if LUA_VERSION_NUM >= 503
+		lua_pushcfunction(cobj.get(), (&safe_misc_trampoline_get<&lua_gettable, 1>));
+#else
 		lua_pushcfunction(cobj.get(), (&safe_misc_trampoline<&lua_gettable, 1>));
+#endif
 		insert(-3);
 		call(2, 1, 0);
 	}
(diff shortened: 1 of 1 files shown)


Nooby4Ever:
The build flags corresponding to the: cairo, imlib2 and rsvg will cause errors as stated in #116 . So you have to disable them when Lua 5.3 is used, this should also be clearly documented somewhere to avoid confusion.

marcpayne:
TL;DR: I will probably close this pull request. Some input on my last code snippet below is welcome. In the meantime, I'll look at the C++ standards.

---

Yeah, until someone can update the bindings you mention (which I can't at this point), `ConkyPlatformChecks.cmake` could be changed to output an error if the user tries to build them with any version of Lua past 5.1. This should probably also be mentioned in the LUA API section of the man page.

Regarding `pkg_search_module(...)`, the only reason it works is that it contains `lua>=5.1`. This will work on any distro that ships a pkgconfig file named `lua.pc`, which most distros should. However, if a user installs a Lua 5.2 package alongside 5.3 on a distro that ships, say, `lua-5.2.pc`, then conky will only see Lua 5.2. I don't know if this is intentional. The following might work better:

```
pkg_search_module(LUA REQUIRED lua>=5.1 lua53 lua5.3 lua-5.3 lua52 lua5.2 lua-5.2 lua51 lua5.1 lua-5.1)
```

Finally, after sleeping on it, I realized that my patch would be much simpler if I just overloaded the existing `safe_misc_trampoline` function template. No messy `#if` statements needed.

```
template<void (*misc)(lua_State *, int), int nresults>
int safe_misc_trampoline(lua_State *l)
{
	misc(l, 1);
	return nresults;
}

template<int (*misc)(lua_State *, int), int nresults>
int safe_misc_trampoline(lua_State *l)
{
	misc(l, 1);
	return nresults;
}
```

Like I often say, this "works for me." Input is welcome before I close this pull request, particularly because I don't know the finer details of overloading function templates with different function pointer parameters, and whether this code will work for everyone else.

Nooby4Ever:
The idea of overloading should work in theory, but i have never used it before so i unfortunately can not comment on the code itself.

antonio-malcolm:
I think tolua++ still needs the older 5.1 Lua, so end users need the legacy library for the dependency. Are there any plans to move to another means of binding to Lua, to output to Cairo? Neither tolua and tolua++ have been updated in YEARS (though tolua, miraculously, compiles against the latest verssion of Lua)..

PR #138, opened 2015-08-23
Be explicit about documentation encoding. Fixes #91.

### LINKED ISSUE #91: db2x_manxml fails on non-ascii chars

paulegan:
Fix for #91 which doesn't rely on removing content from the docs.


brndnmtthws:
Excellent, thanks!

Issue #131, opened 2015-08-16
Executable names with spaces are clipped if top_name_verbose = false


marcpayne:
I'm filing this bug against my own commit (06722e4a). In the (hopefully rare) case where an executable name contains spaces, the output of `${top name <num>}` is clipped if `top_name_verbose = false`. Everything after the first space is ignored, and the reason is clear from the commit.

From what I understand so far, the problem of getting the cmdline/executable of a process in Linux is not straightforward. There are several sources of this data: `/proc/<pid>/{cmdline,stat,status,exe}` and probably others. Conky uses `cmdline` and `stat`. Personally, I think this is plenty robust since conky isn't meant to replace top or ps. I'm just throwing this out there in case anyone wants to discuss pros and cons of other approaches.

To fix the current issue and make issue #121 easier to fix, it might be useful to add a `basename` entry in the `process` struct in `top.h` that will contain just the executable name. I hope to get some input on this idea before I try to make it happen.

Nooby4Ever:
Given the current string,  you can use some more advanced parsing (eg split on blank-space and return all groups, with again an added blank-space in between, until a group starts with "/" or "-"). But this will only lower the erroneous cases, it is in essence equally error prone just like the current solution.

I don't see any real downsides on the <code>basename</code> solution and that solution would be preferred (since making #121 easier) over my remark above, but my knowledge on the matter is rather limited so my advice must be take with caution.

marcpayne:
Yeah, I've been thinking about several different ways I could parse it. The most robust way seems to involve combining the information from `cmdline`, `stat`, and `exe`. While doing this will help with some of the edge cases, it makes the code too complex.

I think I'll combine the old behavior with the new. Executable file names from `stat` (old) will go into `basename`, and full command lines from `cmdline` (new) will go into `name`. This is how top from procps-ng handles it. Nice and simple, and it will fix both `${top name ...}` and `${if_running ...}` without breaking the new stuff.

Thanks for the input! It's helpful. My knowledge is limited too. I'm trying hard not to spaghettify the code as I figure things out.

Nooby4Ever:
Something else came to mind, while not directly related to this issue it might be worth considering since you're working on it anyway. -No pressure-

[...]

Issue #121, opened 2015-07-24
if_running unexpected behaviour + requires extra white-space


Nooby4Ever:
On Linux.
1) You're forced to give the full verbose name of the process you want to check, i think this is not (always) desired; or at least we should be given the option not to.

2) If there is no white-space behind the statement eg <code>${if_running firefox}</code> it will evaluate to false while <code>${if_running firefox }</code> will evaluate to true.
***
Minimal example (run conky in the console, with firefox running):
```
conky.config = {
	out_to_x = false,
	out_to_console = true,
};


conky.text = [[
${if_running conky }Running True${else}Running False${endif}
${if_running conky -c /home/nooby4ever/Desktop/test }Running True2${else}Running False2${endif}
${if_running firefox }Running True3${else}Running False3${endif}
${if_running firefox}Running True4${else}Running False4${endif}
]];
```
Output:
```
Running False
Running True2
Running True3
Running False4
```
While this may be more accurate, eg: when running multiple conkies you can check whether or not a specific instance is running instead of just "an instance" is running but for some applications this accuracy is not required/desired.
At least it needs to be specified in the documentation and i would suggest giving the user the option (short or verbose name); unfortunately my C/C++ skills are not enough to code it by myself.

In essence it's rather similar to: https://github.com/brndnmtthws/conky/issues/113

marcpayne:
@Nooby4Ever I'll take a look at how `if_running` is handled as time permits. Checking the git log might give some insight into the rationale behind the new behavior. I try to give the benefit of the doubt before hacking away. However, based on your report here, there is definitely a bug that needs to be fixed (namely the requirement for a trailing space to make `if_running` evaluate to true).

marcpayne:
OK, I finally got some time to look at this problem. The trailing space issue is really interesting to me because some `/proc/<pid>/cmdline` entries (e.g. `chromium --type=gpu-broker`) are padded with a *lot* of extra null characters for some reason, which are all replaced with spaces by conky's code in `linux.cc`. This part shouldn't be too hard to fix, hopefully.

BTW, nice job spotting this issue. Your report made my search much simpler. :)

To fix the issue about having to match against full command lines, I need to research some of the C string functions. I'm not that familiar with C yet.

This comment is partly a "note to self," and partly a note to C gurus who might want to tackle the issues themselves.

Nooby4Ever:
[...]

PR #140, opened 2015-08-28
Support for basenames (executable names)

marcpayne:
I have spent a lot of time recently learning how Conky gathers and stores process names. It became apparent that there was a need for two different internal representations of process names: the full command line with arguments (taken from `/proc/<pid>/cmdline`, with a fallback to `/proc/<pid>/stat` in case `cmdline` is empty), and the basename, aka executable name (taken from `/proc/<pid>/stat`). This approach is very similar to how `top` (from procps-ng) works, and it allowed me to fix three issues a little more elegantly than I could have otherwise, namely #113, #121, and #131.

I implemented the basename as a new entry in the `process` struct defined in `top.h`, and used it to address the issues mentioned above. I also took the opportunity to clean up some code formatting issues. Please see my commit messages for all the details.

The only change that might be controversial is my decision to remove code that dealt specifically with the string "kdeinit: " in process names. In my commit message for this change, I listed five reasons for doing so. This was my last commit, so if it turns out that this decision was shortsighted, it will be easy to revert.

### DIFF
--- a/doc/variables.xml
+++ b/doc/variables.xml
@@ -1849,9 +1849,21 @@
             </command>
             <option>(process)</option>
         </term>
-        <listitem>if PROCESS is running, display everything
-        $if_running and the matching $endif. This uses the
-        ``pidof'' command, so the -x switch is also supported. 
+        <listitem>If PROCESS is running, display everything
+        between $if_running and the corresponding $else or $endif.
+        Note that PROCESS may be either a full command line with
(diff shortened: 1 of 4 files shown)


arclance:
Has anyone looked at this yet?
I still see a lot of complaints about ${if_running} being broken and this would fix it.

brndnmtthws:
Lovely!

Issue #118, opened 2015-07-21
templateN Segmentation fault


Nooby4Ever:
Templates seem broken, this is the output of the minimal example in the console
```
  bar 2
 best
0.25 best2
```
(actually it are not question marks but those little squares with 4 numbers inside, 0,0,0,1 and 0,0,0,2 in this case. But i don't know how to display them in github)

And template1 throws the following error: <code>conky: invalid argument count for top</code>
If <code>out_to_x</code> is set to <code>true</code> conky produces a <code>Segmentation fault (core dumped)</code>

Minimal example:
```
conky.config = {
    own_window = true,
    own_window_type = 'normal',
    out_to_x = false,   -- Toggle 
    out_to_console = true,

    template0  = '\1 \2',
    template1 = '${top cpu \1}',
    template2 = '${top cpu 1}',
};

conky.text = [[
${template0 foo 1} bar 2
${template1 1} best
${template2 1} best2
]];
```
These commands did not help:
```
use_xft = true,
override_utf8_locale = true,
```

***
Debug output:
```
conky -DD -c /home/nooby4ever/Desktop/a
DEBUG(0) [/home/nooby4ever/Desktop/conky-lua-nv/src/conky-1.10.0/src/conky.cc:2671]: reading contents from config file '/home/nooby4ever/Desktop/test'
DEBUG(1) [/home/nooby4ever/Desktop/conky-lua-nv/src/conky-1.10.0/src/template.cc:149]: template0 foo 1: substituted arg 'foo' to 'foo'
DEBUG(1) [/home/nooby4ever/Desktop/conky-lua-nv/src/conky-1.10.0/src/template.cc:149]: template0 foo 1: substituted arg '1' to '1'
DEBUG(0) [/home/nooby4ever/Desktop/conky-lua-nv/src/conky-1.10.0/src/template.cc:155]: substituted template0 foo 1, output is ' '
DEBUG(1) [/home/nooby4ever/Desktop/conky-lua-nv/src/conky-1.10.0/src/template.cc:149]: template1 1: substituted arg '1' to '1'
DEBUG(0) [/home/nooby4ever/Desktop/conky-lua-nv/src/conky-1.10.0/src/template.cc:155]: substituted template1 1, output is '${top cpu }'
DEBUG(1) [/home/nooby4ever/Desktop/conky-lua-nv/src/conky-1.10.0/src/template.cc:149]: template2 1: substituted arg '1' to '1'
[...]

PR #141, opened 2015-08-29
Avoid segfault if template arg number is zero

marcpayne:
If the user enters `\0` as a template argument, print a warning and exit gracefully. Fixes an issue mentioned in #118.


brndnmtthws:
Makes sense, thanks!

PR #148, opened 2015-09-21
Fix convert.lua : Splitting fails

Archloy:
Fixed bug when script try to split settings and "text". If there is any characters (like space) after "TEXT" token, the script fails to split and return an error on settings local variable usage. Now accept any characters between "TEXT" and '\n'.

### DIFF
--- a/extras/convert.lua
+++ b/extras/convert.lua
@@ -133,7 +133,7 @@ end;
 local config = input:read('*a');
 input:close();
 
-local settings, text = config:match('^(.-)TEXT\n(.*)$');
+local settings, text = config:match('^(.-)TEXT(.*)\n(.*)$');
 
 local converted = 'conky.config = {\n' .. settings:gsub('.-\n', convert) .. '};\n\nconky.text = ' ..
                 quote(text) .. ';\n';


brndnmtthws:
Can you provide some examples? It's a little difficult to follow the regex.

brndnmtthws:
Also, according to the Lua manual, `.` matches all characters: http://www.lua.org/manual/5.2/manual.html#6.4.1

What will happen with the newline?

Archloy:
The base lua script to convert old conkyrc just split the config file as : 
{first part}\nTEXT\n{second part}. 
My "patch" split as : 
{first part}\nTEXT \n{second part}. 

The space character or any comments will be ignored by my patch. Otherwise, the splitting action will fail.

Tucos:
I believe there are 2 problems with the patch:
- The `*` is greedy, so the newline will match the last newline in the
file, not the first one after TEXT. This match should be `.-` instead (like
the first group).
- Your patch matches the whitespace/comment after TEXT in a group, but that
was not taken into account when assigning the return values, so the
variable `text` will actually contain whatever that group matched and the
actual text is discarded. The group is just not necessary here.

-- 
Tucos

Archloy:
After checking my own convert.lua, the patch does not match the last newline in file, otherwise, it can't work correctly for me.

Tucos:
It appears to work, because the 2 issues sort of compensate each-other.
You'll find the second group matches everything starting at the end of TEXT
up-to the last newline in the file. The third (unused) group will always be
empty.
Also, the whitespace/comment after TEXT will be included in the output for
the actual text (where it used to be ignored by conky 1.9.x).
I don't care whether the garbage is included in the result or not, but
having an unused group that never matches anything (well-formed files end
each line with an EOL) is just ugly.

>

brndnmtthws:

Issue #103, opened 2015-07-02
1.10.0 Randomly fails with memory allocation error after some time


corpix:
I am using Archlinux.

Log:
```
...
Jul 02 05:58:47 machine conky[14691]: conky: desktop window (268) is root window
Jul 02 05:58:47 machine conky[14691]: conky: window type - desktop
Jul 02 05:58:47 machine conky[14691]: conky: drawing to created window (0x2400001)
Jul 02 05:58:47 machine conky[14691]: conky: drawing to double buffer
Jul 02 06:02:46 machine conky[14691]: conky: std::bad_alloc
```

Waiting for next fail to collect more data.

Lompik:
I was having the same `std::bad_alloc` error but using conky to print to console. 

I removed all reference to `MPD` (music player daemon) in the config file and no more random errors.

cpmdolyp:
I was having the same std::bad_alloc

the other information is normal
conky: desktop window (297) is root window
conky: window type - override
conky: drawing to created window (0x1600001)
conky: drawing to double buffer
conky: forked to background, pid is 24319

aereaux:
I'm also having what appears to be the same problem.  Launching conky -DD gives me:

    DEBUG(0) [/build/conky/src/conky-1.10.0/src/conky.cc:2671]: reading contents from config file '/home/user/.conkyrc'
    conky: desktop window (1c00027) is subwindow of root window (d4)
    DEBUG(0) [/build/conky/src/conky-1.10.0/src/x11.cc:587]: Found ARGB Visual
    conky: window type - desktop
    conky: drawing to created window (0xa00002)
    conky: drawing to double buffer
    DEBUG(1) [/build/conky/src/conky-1.10.0/src/core.cc:1913]: no templates to replace
    DEBUG(1) [/build/conky/src/conky-1.10.0/src/core.cc:576]: Adding $cpu for CPU 0
    DEBUG(1) [/build/conky/src/conky-1.10.0/src/core.cc:594]: Adding $cpugraph for CPU 0
    DEBUG(1) [/build/conky/src/conky-1.10.0/src/core.cc:1913]: no templates to replace
    DEBUG(1) [/build/conky/src/conky-1.10.0/src/core.cc:1913]: no templates to replace
    DEBUG(1) [/build/conky/src/conky-1.10.0/src/core.cc:1913]: no templates to replace
[...]

PR #160, opened 2015-11-02
Expand several buffers identified by cppcheck

bldewolf:
I ran cppcheck on a whim while trying to debug #103 and it found these possible
memory corruptions. They all seem to be buffers that were made one byte too
small (leaving out the byte for the C-string null terminator).


brndnmtthws:
Hmm yes. Thanks!

PR #161, opened 2015-11-03
CMake fixes

mlichvar:
The following modifications of the cmake files seem to be needed to compile conky and install correctly on Fedora.


brndnmtthws:
Looks like the build is failing. Can you take another look?

mlichvar:
Oh, it looks like we need to include both CheckIncludeFile and CheckIncludeFiles. Should be fixed now.

cwebster2:
CheckIncludeFile doesn't exist on my system (and apparently others with newer versions of cmake).  See my PR for Issue #162 that gets rid of the need for CheckIncludeFile in favor of CheckIncludeFiles.

brndnmtthws:
Can you rebase again on master?

mlichvar:
The remaining patch is now rebased. Thanks.

PR #165, opened 2015-11-16
Fixes typo in ConkyPlatformChecks

cwebster2:
Fixes #162 by changing CheckInlcudeFile to
CheckIncludeFiles so that CMake can correctly include
the module.

PR #166, opened 2015-11-17
Reverts a change from check_include_file to files

cwebster2:
One of the check_include_file checks in ConkyPlatformChecks should
not have been changed to check_include_files because the latter
takes 2 arguments and this call gives 3 (and doens't work right
without the third).

### DIFF
--- a/cmake/ConkyPlatformChecks.cmake
+++ b/cmake/ConkyPlatformChecks.cmake
@@ -141,7 +141,7 @@ if(BUILD_MYSQL)
 endif(BUILD_MYSQL)
 
 if(BUILD_WLAN)
-	check_include_files(iwlib.h IWLIB_H -D_GNU_SOURCE)
+	check_include_file(iwlib.h IWLIB_H -D_GNU_SOURCE)
 	if(NOT IWLIB_H)
 		message(FATAL_ERROR "Unable to find iwlib.h")
 	endif(NOT IWLIB_H)


cwebster2:
This change in my PR yesterday should not have been made (all the others are good however).  This PR reverts one line back to its previous state and I confirmed locally that I can build with the BUILD_WLAN option.

>-- Looking for iwlib.h
-- Looking for iwlib.h - found

brndnmtthws:
Thanks!

arclance:
It looks like you forgot to add "include(CheckIncludeFile)" back (you changed it to "include(CheckIncludeFiles)" in Pull Request #165) with this pull request so this is not completely reverted and configuration still fails.

cwebster2:
I didn't forget.  That module, afaik no longer exists on CMake 3.4, so
adding it back in will break builds with that cmake version.  Perhaps what
is really needed is a cmake version check that includes the file "without
the s" on cmake < 3.4 and "with the s" on >= 3.4 or whatever version first
introduces check_include_files.

arclance:
Yes it does need something like that, not all people will be rushing to update to such a recent CMake and it does not work as is on older versions.
On Debian I would have to install CMake from Experimental to get 3.4 and I won't be installing any compilation stuff from there due to the how easy it is to break your setup installing those kinds of packages from Experimental.

cwebster2:
I doubt it is 3.4 that first makes the change.  Whatever version runs on
Travis CI that builds the PR has a version that works with
"check_include_files".  3.4 might be the first release that
"Check_include_file" stops working, but it isn't the first that other
starts working.

arclance:
Maybe I was not clear enough "CheckIncludeFiles" is not causing problems for me.
It is the lack of a "include(CheckIncludeFile)" that causes the one check that still uses "CheckIncludeFile" to fail for me making the entire configuration step fail.
This is the error I get without the include.
```
CMake Error at cmake/ConkyPlatformChecks.cmake:144 (check_include_file):
   Unknown CMake command "check_include_file".
 Call Stack (most recent call first):
   CMakeLists.txt:37 (include)
```
Adding the include to the file makes the configuration work for me and the build works.
I did have problems linking to libtolua++ but that might be due to changes in Debian since I don't think there was a Lua5.2 package the last time I built conky on this system.

cwebster2:
I guess I misunderstood.  It would be helpful to figure out which is the
first version of CMake to not need that file (3.2?) and include it only
when cmake is older than that.  That way it'll work for everyone.  If no
one else gets to it first, I'll look into it and submit a PR.

arclance:
I can't find anything that shows it is removed but this commit updates the documentation about them both in July 2015.
https://cmake.org/gitweb?p=cmake.git;a=commitdiff;h=81d58b0d

This seems to indicate that you can use a "CMAKE_REQUIRED_FLAGS" call immediately before the "CheckIncludeFiles" call with the third argument to "CheckIncludeFile" as the "CMAKE_REQUIRED_FLAGS" argument to get the same result as the current "CheckIncludeFile" line.
That seems like it could remove the need for using "CheckIncludeFIle" at all since the need for the third argument is why it is still being used right?
```
-# CHECK_INCLUDE_FILES(INCLUDE VARIABLE)
-#
-# ::
-#
-#   INCLUDE  - list of files to include
-#   VARIABLE - variable to return result
-#              Will be created as an internal cache variable.
-#
+#     CHECK_INCLUDE_FILES("<includes>" <variable>)
 #
+#   Check if the given ``<includes>`` list may be included together
+#   in a ``C`` source file and store the result in an internal cache
+#   entry named ``<variable>``.  Specify the ``<includes>`` argument
+#   as a :ref:`;-list <CMake Language Lists>` of header file names.
 #
 # The following variables may be set before calling this macro to modify
 # the way the check is run:
 #
-# ::
+# ``CMAKE_REQUIRED_FLAGS``
+#   string of compile command line flags
+# ``CMAKE_REQUIRED_DEFINITIONS``
+#   list of macros to define (-DFOO=bar)
+# ``CMAKE_REQUIRED_INCLUDES``
+#   list of include directories
+# ``CMAKE_REQUIRED_QUIET``
+#   execute quietly without messages
 #
-#   CMAKE_REQUIRED_FLAGS = string of compile command line flags
-#   CMAKE_REQUIRED_DEFINITIONS = list of macros to define (-DFOO=bar)
-#   CMAKE_REQUIRED_INCLUDES = list of include directories
-#   CMAKE_REQUIRED_QUIET = execute quietly without messages
+# See modules :module:`CheckIncludeFile` and :module:`CheckIncludeFileCXX`
+# to check for a single header file in ``C`` or ``CXX`` languages.
```
https://cmake.org/gitweb?p=cmake.git;a=blobdiff;f=Modules/CheckIncludeFiles.cmake;h=843cd35ef738c22b91d65571901ac232b97e06c2;hp=2494862457943318cfc7c2b6f1da4964eec2b889;hb=81d58b0de62d3e1fdaeaf7029b0a415c32cf6b2a;hpb=a6916a6c6e397bda8b381f65dbfc59d3c52f8525

cwebster2:
CMAKE_REQUIRED_DEFINITIONS looks perfect. I'll update my fork, test and
push in the morning (EST).

cwebster2:
@arclance PR #177 has the fix for this.  I've tested locally that I can build with -DBUILD_WLAN=1 and I verified the checks are run and succeed.  I'd appreciate if you could give it a shot too as the travis CI config doesn't test that config afaik.

arclance:
I will test it when I get home tonight.
I can't test anything at work.

arclance:
PR #177 made it configure, build, and run without problems for me.

PR #178, opened 2015-12-04
Fix for compiling against toluapp

antonio-malcolm:
This is needed for compiling against toluapp on some systems, Void and Arch included.


brndnmtthws:
Thanks!

PR #179, opened 2015-12-05
Add top_name_verbose to nano and vim syntax files

marcpayne:
I meant to do this a long time ago... Better late than never.


brndnmtthws:
Thanks!

Issue #155, opened 2015-10-15
Initial spike in network speed graph


mxmlnkn:
When starting conky after quite some time, the network graphs upspeedgraph and downspeedgraph show a very very large initial spike, which seems to correspond to the value given back by totalup and totaldown. Even when using logarithmic scales this spike of 40MB/s makes every other traffic the next 2 minutes (until the spike leaves the graph to the left) impossible to see. I think this initial traffic should be ignored.

stardiviner:
+!

txtsd:
:+1: 
Please implement this.

karlstiller:
I'm getting something similar with memgraph. I've noticed that the graph expands to the right when the spike appears (The left hand side of the graph stays put, but the right side extends further to the right). The spike appears infrequently but usually within one minute.

karlstiller:
Found my problem regarding the spike in memgraph. It was a malformed line in my .conkyrc file. The original was:
 ${color grey}RAM: $memgraph 20,130 000000 ffffff

Adding the missing curly brackets:
 ${color grey}RAM: ${memgraph 20,130 000000 ffffff}

I didn't change this line between the update from the previous version to 1.10.

mxmlnkn:
@karistiller: Unfortunately this doesn't seem to be the reason, for me.

I will take a look now at the source code and try my best, but I still think the maintainer, whoever that is would need 10 minutes for what I'll need several hours.

mxmlnkn:
see merge request #182

PR #182, opened 2015-12-10
Issue #155: Fix for initial spike on program 

mxmlnkn:
Also added documentation as much as needed to implement that fix.
Also two lines with white spaces at the end were automatically removed by SciTE.


brndnmtthws:
Awesome, looks great!

PR #184, opened 2015-12-11
Require Lua 5.1 when building the Lua bindings

marcpayne:
The cairo/imlib2/rsvg bindings require tolua++ [1], which is currently only compatible with Lua 5.1 or earlier. We need to force Conky to compile against Lua 5.1 if the user chooses to build the bindings. See [2] and [3] for some discussion on this topic.

This commit adds a separate pkg_search_module command specifically for Lua 5.1, along with a note in the documentation. The original pkg_search_module command was updated to include (and prefer) Lua 5.3+. Also, I threw in two other minor random fixes to the docs while I was at it.

One more thing to note: in case there is a distro out there that ships Lua 5.1 with a pkg-config file named lua.pc, this is accounted for with `lua<=5.1.5` (`lua=5.1` is too strict and `lua<5.2` is not possible with cmake, afaik). Since Lua 5.1.5 was released in February of 2012, this should be safe.

[1] https://github.com/LuaDist/toluapp
[2] https://github.com/brndnmtthws/conky/issues/116
[3] https://github.com/brndnmtthws/conky/issues/174


marcpayne:
Before this PR is merged, there is a small change I would like to make to the Lua 5.1 search so that it is a little more robust. Namely, replace `lua<=5.1.5` with `lua` to match any version, then check whether `LUA_VERSION` is less than `5.2.0`. Searching for `lua5.1` and `lua-5.1` should cover most (if not all) cases, but it doesn't hurt to be thorough.

I'll squash and force push my changes once I have time to test them. I had read that changing history once you've pushed to a public repo is a Bad Thing (perfectly understandable), but it seems to be an acceptable practice for PR's as long as no one has started working on something based on the original PR commit(s).

marcpayne:
Okay, fixed and tested.

brndnmtthws:
Thank you.

PR #185, opened 2015-12-11
Fix BUILD_MATH flag not working, partial #128

mxmlnkn:
changed MATH to BUILD_MATH + 2 smaller changes which didn't work out of the box after that change

Fixes:
 - cpugauge not working
 - logarithmic scales not working in all graphs
-  graph scale not showing, even if 'show_graph_scale yes' specified
 - ...

Also added documentation as much as needed to find these problems

This fixes partially Issue #128 , but there is a more fundamental problem: In the range of 0 to 1.0 a as used for the cpugraph usage, the 'logarithmic' scale (which isn't strictly true) calculates log10(f+1), meaning f=0 will result in 0 and f=1.0 will result in log10(2)=0.3. f=0.5 will result in 0.176 

All this means, that the logarithmic version is only very very slightly different than the linear version. It's visible, but you have to look for it, in contrast to the downspeedgraph, where the logarithmic scale has it's use.

The solution would be to scale ab the cpugraph values plotted e.g. by factor of 100 to range from 0 to 100. The resulting logscale: 100->2.0, 50->1.7, 0->0 would be clearly visible.
But then the scale shown if 'show_scale yes' would show 100 instead of 1.0 which is basically incorrect, because the percent sign after the 100 would be missing. But this is just ideology, I'm not sure if I should implement this
???

### DIFF
--- a/src/conky.cc
+++ b/src/conky.cc
@@ -281,7 +281,7 @@ static void print_version(void)
 #endif /* BUILD_MYSQL */
 #ifdef BUILD_WEATHER_METAR
                 << _("  * Weather (METAR)\n")
-#endif /* BUILD_WEATHER_METAR */                
+#endif /* BUILD_WEATHER_METAR */
 #ifdef BUILD_WEATHER_XOAP
                 << _("  * Weather (XOAP)\n")
 #endif /* BUILD_WEATHER_XOAP */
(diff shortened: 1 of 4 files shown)


brndnmtthws:
Looks great, thanks!

PR #186, opened 2015-12-15
Finalized nvidia module

fonic:
I finalized the nvidia module including code cleanup and documentation (variables.xml).

Changes to previous revision:
- Implemented `nvs->pmode` to allow control over how values are printed (int, float, temperature...)
- Moved decoding of GPU/MEM frequencies to `print_nvidia_value()` using QUERY_SPECIAL -> cleaner quirks handling
- Revised `print_nvidia_value()` -> cleaner string and memory allocation handling
- Renamed `set_nvidia_type()` to `set_nvidia_query()` -> required change in core.cc
- General code cleanup (comments, spacing)

Please note that:
- The module partially breaks compatibility to the previous (old) nvidia module since shortcuts (e.g. 'g' for 'gpufreq') are no longer supported. However, I was able to maintain compatibility for full argument strings (temp, threshold, ambient, gpufreq, memfreq)
- I decided against multiple GPUs/fans/... support. I don't feel it's worth the effort since users most likely have a single card or only want to monitor their primary card
- I decided against static metrics (e.g. min/max values which are only queried once to save CPU time) since this is a feature I'd like to propose in general for conky
- There are features (PRINT_FLOAT, QUERY_STRING) that are currently unused, but might come in handy in the future. I designed the whole module with easy further enhancement in mind.

### DIFF
--- a/doc/variables.xml
+++ b/doc/variables.xml
@@ -2715,45 +2715,119 @@
             <command>
                 <option>nvidia</option>
             </command>
-            <option>threshold</option>
-            <option>temp</option>
-            <option>ambient</option>
-            <option>gpufreq</option>
-            <option>memfreq</option>
-            <option>imagequality</option>
+            <option>metric</option>
         </term>
-        <listitem>Nvidia graficcard support for the XNVCtrl
-        library. Each option can be shortened to the least
-        significant part. Temperatures are printed as float, all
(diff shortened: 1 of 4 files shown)


fonic:
I'll close this pull request since I discovered two bugs that I'll fix first.

PR #192, opened 2015-12-30
Refactor the exec callback system

marcpayne:
### Overview

Prior to this commit, exec_cb callbacks were registered and the results were immediately fetched on every update interval... except the first. Besides not working on the first update, this meant that all exec_cb callbacks were re-registered on every update, which was not efficient.

The solution was to decouple the registration from the result retrieval and only register on the initial startup (or config reload). Now, the callbacks are registered when they are supposed to be, and the first-run issue is gone.

The code is hopefully more understandable now. I consolidated several functions in `exec.cc` into one, which makes the logic much easier to follow. I also added plenty of documentation, both to the code itself and to the manpage.

### Bonuses

  * execgraph works again!
  * optional execgraph parameters are now documented
  * execbar and execgauge now accept height and width parameters

### Known issues

  * technically, execgraph accepts commands with spaces, but such commands will need to be surrounded with double-quotes and you might as well forget adding optional arguments like colors, as the results may not be expected. Just put your spacey command into a script and run the script instead so that everything will just work.
  * the execgraph command comes *before* its arguments, whereas the execbar and execgauge commands come after their arguments.
  * texec{i} is registered on startup like everything else, but sometimes it will not produce output on the first interval. This is an intermittent issue.

### Example

Exec Demo

```
conky.text = [[
EXEC DEMO ${hr}
${exec date}

Default execbar w/ default_bar_height = 10
${execbar shuf -i 0-100 -n 1}

Variable size execbars
${execbar 20,150 shuf -i 0-90 -n 1}
${execbar 10,100 shuf -i 0-92 -n 1}
${execbar 3,50 shuf -i 0-94 -n 1}

Customized execgraphs
${execgraph ~/seconds.sh 25,150 FF0000 FFFF00 60 -t}${execgraph ~/random.sh 25,150 -l}

Variable size execgauges
${execgauge 50,80 shuf -i 0-100 -n 1}${execgauge 100,220 shuf -i 0-99 -n 1}
]];
```

### DIFF
--- a/src/text_object.h
+++ b/src/text_object.h
@@ -33,6 +33,7 @@
 #include "config.h"		/* for the defines */
 #include "specials.h"		/* enum special_types */
 #include "update-cb.hh"
+#include "exec.h"
 
 /* text object callbacks */
 struct obj_cb {
@@ -87,12 +88,24 @@ public:
         : Base(period, true, Base::Tuple(fn))
     {}
 };
-typedef conky::callback_handle<legacy_cb> legacy_cb_handle;
 
+typedef conky::callback_handle<legacy_cb> legacy_cb_handle;
+typedef conky::callback_handle<exec_cb> exec_cb_handle;
+
+/**
+ * This is where Conky collects information on the conky.text objects in your config
+ *
+ * During startup and reload, objects are parsed and callbacks are set. Note that
+ * there are currently two types of callback: obj_cb (old style) and
+ * conky::callback (new style). On each update interval, generate_text_internal()
+ * in conky.cc traverses the list of text_objects and calls the old callbacks.
+ * The new style callbacks are run separately by conky::run_all_callbacks().
+ */
 struct text_object {
 	struct text_object *next, *prev;	/* doubly linked list of text objects */
 	struct text_object *sub;		/* for objects parsing text into objects */
 	struct text_object *ifblock_next;	/* jump target for ifblock objects */
+
 	union {
 		void *opaque;		/* new style generic per object data */
 		char *s;		/* some string */
@@ -102,11 +115,16 @@ struct text_object {
 
 	void *special_data;
 	long line;
-	struct obj_cb callbacks;
-	bool parse;	//if this true then data.s should still be parsed
-	bool thread;	//if this true then data.s should be set by a seperate thread
+	bool parse;	/* if true then data.s should still be parsed */
+	bool thread;	/* if true then data.s should be set by a seperate thread */
 
-        legacy_cb_handle *cb_handle;
+	struct obj_cb callbacks;
+
+	/* Each _cb_handle is a std::shared_ptr with very tight restrictions on
+	 * construction. For now, it is necessary to store them here as regular
+	 * pointers so we can instantiate them later. */
+	exec_cb_handle *exec_handle;
+	legacy_cb_handle *cb_handle;
 };
 
 /* text object list helpers */
--- a/src/update-cb.hh
+++ b/src/update-cb.hh
@@ -56,13 +56,13 @@ namespace conky {
 
 			semaphore sem_start;
 			std::thread *thread;
-			const size_t hash;
-			uint32_t period;
-			uint32_t remaining;
+			const size_t hash;		/* used to determined callback uniqueness */
+			uint32_t period;		/* how often to run a callback */
+			uint32_t remaining;		/* update intervals remaining until we can run a callback */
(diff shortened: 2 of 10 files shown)


marcpayne:
I already see changes that I should make, e.g. my enum would make more logical sense in exec.h.

Before I do that, I have a style question. Which do you prefer: `if (foo)` or `if(foo)`? Likewise with for/while. I like clearing up inconsistencies as I go.

KamranMackey:
@marcpayne I prefer `if (foo)` personally.

stardiviner:
I like it too.

marcpayne:
Yeah, I prefer that too. I added spaces in exec.cc where I did most of the work and left the rest as is. Also, I moved the enum over to exec.h today, so I think I'm done for now, barring any requests for changes.

KamranMackey:
@brndnmtthws this look good to you, Brenden?

src/text_object.h
```diff
@@ -76,25 +77,37 @@ void gen_print_nothing(struct text_object *, char *, int);
...
+ * During startup and reload, objects are parsed and callbacks are set. Note that
+ * there are currently two types of callback: obj_cb (old style) and
+ * conky::callback (new style). On each update interval, generate_text_internal()
+ * in conky.cc traverses the list of text_objects and calls the old callbacks.
+ * The new style callbacks are run separately by conky::run_all_callbacks().
+ */
 struct text_object {
 	struct text_object *next, *prev;	/* doubly linked list of text objects */
-	struct text_object *sub;		/* for objects parsing text into objects */
+	struct text_object *sub;			/* for objects parsing text into objects */
```
brndnmtthws: Looks like some whitespace misalignment here.

src/text_object.h
```diff
@@ -31,8 +31,9 @@
 
 #include <stdint.h>		/* uint8_t */
 #include "config.h"		/* for the defines */
-#include "specials.h"		/* enum special_types */
+#include "specials.h"	/* enum special_types */
```
brndnmtthws: Whitespace.

src/text_object.h
```diff
@@ -76,25 +77,37 @@ void gen_print_nothing(struct text_object *, char *, int);
 void gen_print_obj_data_s(struct text_object *, char *, int);
 
 class legacy_cb: public conky::callback<void *, int (*)()> {
-    typedef conky::callback<void *, int (*)()> Base;
+	typedef conky::callback<void *, int (*)()> Base;
```
brndnmtthws: Whitespace.

src/update-cb.hh
```diff
@@ -56,14 +56,15 @@ namespace conky {
 
 			semaphore sem_start;
 			std::thread *thread;
-			const size_t hash;
-			uint32_t period;
-			uint32_t remaining;
+			const size_t hash;		// used to determined callback uniqueness
```
brndnmtthws: Whitespace, and `/* */` comment blocks are preferred.

brndnmtthws:
Looks good, made a few stylistic suggestions.

src/text_object.h
marcpayne: Yeah, I noticed that the tabs don't align in the diffs. The reason I changed these lines is that they were misaligned in my editors (Kate and vim), which both use a tab width of 4 characters. Should I set the tab width to 8 instead? In any case, I'll fix the spacing so they line up here. Thanks!

src/update-cb.hh
marcpayne: Ah, good to know about the `/* */` comment blocks! I've seen both, so I wasn't sure.

src/text_object.h
marcpayne: Do you prefer hard tabs or soft tabs? Kate took the liberty of changing these tabs automatically, but I figured I would keep the changes, since most of the files seem to use hard tabs. Sorry for all the issues. I'm making note of style preferences for next time.

brndnmtthws: Honestly, I don't care. Most of the existing codes uses hard tabs (i.e., `\t`), so we should stick with that or change all the tabs everywhere. If you want to do the latter, that should be in a separate PR.

marcpayne:
Okay, I cleaned up the whitespace issues and switched any new or modified comments to the `/* */` style. There are still some `//` comments around that I didn't touch, which I will leave for another PR someday.

I'll be happy to clear up any other issues if I missed something (not unusual). Thank you for your patience!

brndnmtthws:
Thank you.

PR #195, opened 2016-01-05
Add cat and catp variables.

evmcl:
Added the following variables:

  cat file

```
Reads a file and displays the contents in conky. This is useful
if you have an independent process generating output that you
want to include in conky.
```

  catp file

```
Reads a file and displays the contents in conky. This is useful
if you have an independent process generating output that you
want to include in conky. This differs from $cat in that it
parses the contents of the file, so you can insert things like
${color red}hi!${color} in your file and have it correctly parsed
by Conky.
```


chepioq:
I test your commit, that work for a simple cat, but not with a cat with command.
For example in my conky if I have 
${catp /home/dominique/.conkyrc/conky-saint/saint.txt}
I have a result (list of saint), but if I want just the first with 
${catp "/home/dominique/.conkyrc/conky-saint/saint.txt" | awk '{print $1}'}
I have nothing

I have same result with ${cat...} command.

evmcl:
> but if I want just the first with
> ${catp /home/dominique/.conkyrc/conky-saint/saint.txt | awk '{print $1}'}
> I have nothing

cat and catp just insert the contents of the specified file into the output of conky. You just specify a file and it reads it, it isn't trying to run a shell command, that's not its purpose. For that, you probably want to tinker with something like:

    ${execp awk '{print $1}' < /home/dominique/.conkyrc/conky-saint/saint.txt}

chepioq:
Ok  I understand.
Then I will continue to use this command:
${exec cat /home/dominique/.conkyrc/conky-saint/saint.txt | awk '{print $1}'}

It works well.

plikhari:
At first I couldn't figure the **why** for this addition of text output with a name **cat** - people will get more confused.
I have to agree with @chepioq

plikhari:
@chepioq 
Perhaps using **tail** or **head** would get you fine grained control on output.

evmcl:
The name "cat" comes from the linux comment. My motivation for adding this command is because I have an independent script that monitors some systems state, and among other things writes a file to be displayed by conky when things change.

I could have just used `${execp cat file}` but spawning the child processes every three seconds or so such a trivial task seemed wasteful. It made me itchy, so I scratched.

chepioq:
plikhari wrote : 
"Perhaps using tail or head would get you fine grained control on output - specially in cases where you want to print a particular line(s)"

In my case the file has one line with name of today's saints.

I just want that first name is in color2, and the rest in color1.
I have the good result with :
${color1}Bonne fête aux ${color2}${exec cat ~/.conkyrc/conky-saint/saint.txt | awk '{print $1}'} ${color1}${exec cat ~/.conkyrc/conky-saint/saint.txt | cut -f2- -d ' ' | xargs -n5}

The "xargs -n5" command is for limited, in conky, the number or word to 6 by lines.

evmcl:
Well, if whatever program is generating the saint.txt file could be tweaked to include the "${color1}" tag within the text file itself, then ${catp} could pull it in for you no problems.

I should have mentioned above that the monitor script produces the text file with some conky specific ${colorN}, ${goto} and other formatting in it.

chepioq:
"Well, if whatever program is generating the saint.txt file could be tweaked to include the "${color1}" tag within the text file itself, then ${catp} could pull it in for you no problems."

I understand that but I think it's more complicated to insert ${color1} and ${color2}  in my file.

The ${exec cat...} command does the job simply.

plikhari:
Well - there are **many ways to bell the cat** :-))
@evmcl  I see what you mean ...
@chepioq
Using your method - perhaps you could use execpi with a time interval of a few hours - so it is not parsed every update_interval - this would also reduce the number of shell calls and extra conky processes
The other method as @evmcl  suggests - use a script before conky to add the desired tags and let the new **catp** handle the rest. Agreed what ever makes you happy - that is why the remark re **bell the cat**
OR - use lua  (my way) - lua is the scripting language for conky ;-)

chepioq:
Ok 
For test I can add ${colorN} tag in my file with sed, and that work well with ${catp....}
Just one remark, that don't work with ~/.conkyrc/conky-saint/saint.txt , that need the full path : /home/dominique/.conkyrc/conky-saint/saint.txt

plikhari:
Use ```template# =  '/home/dominique/.conkyrc/conky-saint/saint.txt'```  for ```#``` use some number 0 to 9 in the conky.config section and then use it in the conky.text section with ```execpi 9000 cat ${template1}``` where ```# is 1```

@evmcl  We cannot use a time interval with ```catp ``` - so it will parse the file each iteration of ```update_interval```  - please confirm :-)

evmcl:
> We cannot use a time interval with catp - so it will parse the file each iteration of `update_interval` - please confirm

Correct.

> OR - use lua (my way)

Off topic, but I just used Lua scripting to get Conky reading some data from dbus. Impressed with it.

chepioq:
"Off topic, but I just used Lua scripting to get Conky reading some data from dbus. Impressed with it."

Have you a lua's script example to show ?

evmcl:
> Have you a lua's script example to show ?

Well, definitely going off topic now, so I've made a [gist](https://gist.github.com/evmcl/f09d0cb39c2a0b90b4b5) of some annotated example code. Direct further comments on DBUS or Lua stuff there.

chepioq:
Thank I look for the code.

plikhari:
@chepioq we are Off Topic here - I mentioned lua as a valid alternative since things were being discussed.
The whole idea of lua integration is to extend conky functionality - rather than do micro additions to conky source. 
I have mentioned earlier where to look at the code - I think you probably found it - look at my profile.

evmcl:
Do you think this will be merged in to the master branch at any point? Is there anything I can do to help that along?

Thanks.

plikhari:
Greetings @evmcl 

Here is the thing - there is already this functionality in the form of **tail** and **head** - see the documentation on **Configuration Variables** in the wiki.

These functions are in the file **tailhead.cc**. For some reason the MAX_HEADTAIL_LINES is set to 30 (line 43) - you can change it to 300 if you like while text buffer will handle the limits. Conky does not provide scroll bars anyway.

In **print_tailhead** function - line 177 has `ht->buffer = strdup(p);` below this you can add `evaluate(ht->buffer, p, p_max_size);`

What do you know - with single line of code you get exactly the same functionality as you want **catp** to do. This is great as - if the file has no such color codes - it will output data as normal. One can use color statement before the head or tail line as well.

Usage - `${color0}${head or tail  file_name  number_of_ lines  next_update}` - in this case set next_update to **1** - this will update the file each update_interval - as the default is each alternate update_interval.

Why reinvent the wheel - when it is already there !

cheers

plikhari:
Actually this opens a whole set of unprecedented opportunities within conky.

One can actually have external text file templates of sorts which can be used with this mechanism - rather than a whole bunch of text inside `if` conditions - each could just use this to display formatted data. I have tried calling lua functions from within an external text file and it works just fine ;-)

Something like an **include file** thing.

evmcl:
> with single line of code you get exactly the same functionality as you want **catp** to do.

So... does that mean you're planning a **tailp** command instead? Would that handle if the file it is tailing is replaced (all new contents)?

> Actually this opens a whole set of unprecedented opportunities within conky.

Does that mean this pull request will be merged after all? Don't feel like we're reaching a conclusion one way or the other. Getting confused.

Cheers.

plikhari:
I am just trying to assist you in seeing that the functionality you seek is already there in the program.

The point being is that you want to:
A. read a text file
B. you also want the ability to add conky variables in that text file such as **color** etc

The variables **head** and **tail** show you the first and last (up to) 30 lines of the file. These variables track changes in your file similar to the command line tools.
 
I showed you where to look and what changes to make
A. if you want to read more than 30 lines
B. if you want to use conky variables while using **head** and **tail**

All merge ops are done by Brenden and it is his call.

brndnmtthws:

evmcl:
Done. https://github.com/brndnmtthws/conky/pull/507

PR #196, opened 2016-01-07
respect LIB_INSTALL_DIR

marguerite:
many other cmake projects use this flag too. and some distros have this preset.


chepioq:
I agree with that, I use Fedora and libs are not installed in /usr/local/lib but in /usr/lib or usr/lib64 .
I rebuild the rpm with your commit and that work fine, and now I'm not obliged to make a link in /usr/local/

marguerite:
@chepioq actually it's the patch I use to maintain conky for openSUSE :-)

chepioq:
Please, do you push this commit in master branch, it's very important for maintainer of many distro (the maintainer for fedora need that commits are in the master's branch to add this in the rpm).

brndnmtthws:
Thanks.

PR #198, opened 2016-01-10
Use the canonical source for the number of CPUs.

marius:
Counting cpu\d lines from /proc/stat can give the wrong number if CPUs
are present but offline.


marius:
I forgot to mention that this fixes a segfault on startup on my system.
I set cpu1 offline via `echo 0 >/sys/devices/system/cpu/cpu1/online`
because it is broken.

brndnmtthws:
Thanks.

PR #202, opened 2016-01-21
config: add graph_scale_keep_max option

bkueng:
Hi

This config option prevents graphs from constantly rescaling to the current maximum value, which is distracting.

graph_scale_keep_max option: if true, graphs will always keep the largest scaling instead of rescaling to the maximum visible value.
I've been testing it for a while now and it works fine. However when conky autoreloads a changed config file, the first few values can be very large.

And I'm not entirely sure if removing `s->scale = 1;` has any other negative side effects.


bkueng:
Is there no interest in this?

Nooby4Ever:
I guess it wouldn't hurt to add it, just another option. But it would have been nicer (and make more sense/be more consistent) if you can set the option for each graph individually (like the log scale -l) (slightly related discussion, #246).

mxmlnkn:
> However when conky autoreloads a changed config file, the first few values can be very large.
Has been solved in issue #155
> This config option prevents graphs from constantly rescaling to the current maximum value
Just a suggestion, but you could also use a mnual set maximum, e.g. of your internet connection, like in issue #242, although I guess that would not be portable.

bkueng:
@Nooby4Ever I wasn't quite sure whether to go for a global or per-graph option. My reasoning was: some graphs like cpu & memory have a fixed limit anyway so this option does not affect these. Then for the others you probably want them all to have the same behavior.

@mxmlnkn Thanks. The scale is not really an option for me, as I switch between WiFi's with different speeds.

MattSturgeon:
I can't see a reason why _both_ a global setting and per-graph override couldn't work ;-)

brndnmtthws:

PR #209, opened 2016-02-08
keep next_update_time modulo update_interval zero

aandergr:
This patch modifies logic for setting next_update_time such that the
update is done at full update_interval seconds. That is, for example
with update_interval 60.0, update is done at 00:00:00, 00:01:00,
00:02:00, etc.

This might be useful for using conky as a clock widget which only needs
to be updated once per minute, but is obliged to update exactly at
minute change.

It is implemented in a way that is (hopefully) immune to clockchanges
(as for example by NTP or DST).

### DIFF
--- a/src/common.cc
+++ b/src/common.cc
@@ -123,7 +123,7 @@ double get_time(void)
 {
 	struct timespec tv;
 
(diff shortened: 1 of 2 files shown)


brndnmtthws:
Lovely, thanks!

PR #214, opened 2016-02-16
Add systemd journal support.

tohojo:
This adds support for reading log entries from the Systemd journal. A
new command, called simply 'journal', is added which takes a number of
lines as a parameter and optionally which journal (user or system) to
read from.

Obviously, libsystemd is required to build journal support. Tested on
Arch Linux (Systemd v228).

Signed-off-by: Toke Høiland-Jørgensen <toke@toke.dk>


Nooby4Ever:
You forgot to update the vim and nano syntax, new keywords should be added here:
```
/extras/nano/conky.nanorc
/extras/vim/syntax/conkyrc.vim
```

tohojo:
Right; updated :)

tohojo:
Awesome, thanks! :)

Issue #207, opened 2016-02-06
Feature Request: nvidiabar and nvidiagraph


MattSturgeon:
I have been trying out the updated nvidia module from 22fe3b1ea4f450061dc8dcff1bf4b8c4b560cd7f and I have found `${nvidia gpuutil}` a vastly superior alternative to `${exec nvidia-settings -t -q '[gpu:0]'/GPUUtilization | tr ',' '
'|grep graphics|sed 's/[^0-9]//g'}` (for example).

However I have one glaring issue. So far as I can tell, there is no way to make a graph or bar out of these internal variables. My config has a lot of `${execgraph "nvidia-settings #...etc"}` in it that I would like to replace with something like `${nvidiabar memutil}` or `${nvidiagraph temp}` for example.

I appreciate this may be more complicated than it appears, but I would appreciate any feedback on the suggestion. :)

Nooby4Ever:
You can always use Lua, you can pretty much design the bar however you want. Depends on your knowledge of programming/Lua. There are examples on DeviantArt on how to do it.

MattSturgeon:
I've never used Lua but I do feel like this is something that should really be built into the nvidia module... If I was to try and hack anything together I'd rather it be built into conky rather than a Lua extension.

Right now I have a working solution using `${execgraph}` and the above shell commands so it's not the end of the world.

MattSturgeon:
From a quick look over the code (I don't know C++ and hardly know C) it looks like we need to define an object like this in core.cc??

    	END OBJ_ARG(nvidiabar, 0, "nvidiabar needs an argument")
		if (set_nvidia_type(obj, arg)) {
			CRIT_ERR(obj, free_at_crash, "nvidiabar: invalid argument"
				 " specified: '%s'\n", arg);
		}
		obj->callbacks.barval = &nvidia_bar_value;
		obj->callbacks.free = &free_nvidia;

and we need to split `print_nvidia_value` into functions that print to buffer and a function to return a float for `barval`.

I'm almost certainly missing something really important but I feel like if I was to work on this that's where I'd start

MattSturgeon:
Fixed in d8c659e7de8f7173ff66161e5bf191f5aa621362

I still need to add docs

kgreenek:
Updating the docs would be very helpful. I scoured the internet for way to long until I came across this PR.

Thanks for this feature btw, it's awesome!

PR #215, opened 2016-02-17
nvidiabar, etc

MattSturgeon:
This is not yet ready, however I thought I'd try and get some feedback since I'm new to conky (and C, and C++).

I have managed to setup an object `nvidiabar` in `core.cc` which successfully runs `get_nvidia_barval()` on each update, however it does not yet print a bar to the screen.

I haven't yet found how other bars print to screen so any hints would be greatly appreciated.

(See also #207)

### DIFF
--- a/src/nvidia.cc
+++ b/src/nvidia.cc
@@ -56,6 +56,15 @@
  * TEMP  GPU ${nvidia gputemp}°C (${nvidia gputempthreshold}°C max.), SYS ${nvidia ambienttemp}°C
  * FAN   ${nvidia fanspeed}% RPM (${nvidia fanlevel}%)
  * 
+ * --==| NVIDIA Bars |==--
+ * LOAD ${nvidiabar gpuutil}
+ * VRAM ${nvidiabar memutil}
+ * RAM ${nvidiabar membwutil}
+ * VIDEO ${nvidiabar videoutil}
+ * PCIe ${nvidiabar pcieutil}
+ * Fan ${nvidiabar fanlevel}
+ * TEMP ${nvidiabar gputemp}
+ * 
  */
 
 
@@ -730,10 +739,42 @@ void print_nvidia_value(struct text_object *obj, char *p, int p_max_size)
 }
 
 double get_nvidia_barval(struct text_object *obj) {
-	double val;
-	val = 50; //Dummy test value
-
-	return val;
+	struct nvidia_s *nvs = static_cast<nvidia_s *>(obj->data.opaque);
+	int temp1, temp2;
+	double value;
+	
+	// Assume failure
+	value = 0;
+	
+	// Convert query_result to a percentage
+	if (nvs != NULL) {
+		switch (nvs->attribute) {
+			case ATTR_UTILS_STRING: // one of the percentage utils (gpuutil, membwutil, videoutil and pcieutil)
+				value = get_nvidia_string_value(nvs->target, ATTR_UTILS_STRING, nvs->token, nvs->search);
+				break;
+			case ATTR_MEM_UTIL: // memutil
+				temp1 = get_nvidia_value(nvs->target, ATTR_MEM_USED);
+				temp2 = get_nvidia_value(nvs->target, ATTR_MEM_TOTAL);
+				value = ((float)temp1 * 100 / (float)temp2) + 0.5;
+				break;
+			case ATTR_FAN_LEVEL: // fanlevel
+			case ATTR_FAN_SPEED: // TODO warn user to use fanlevel if they use fanspeed
+				value = get_nvidia_value(nvs->target, ATTR_FAN_LEVEL);
+				break;
+			case ATTR_GPU_TEMP: // gputemp (calculate out of gputempthreshold)
+				temp1 = get_nvidia_value(nvs->target, ATTR_GPU_TEMP);
+				temp2 = get_nvidia_value(nvs->target, ATTR_GPU_TEMP_THRESHOLD);
+				value = ((float)temp1 * 100 / (float)temp2) + 0.5;
+				break;
+			// TODO: calculate gpufreq, memfreq, etc
+			// can use (val-min)÷(max-min)×100. Perhaps a helper function or macro
+			
+			// TODO: throw errors if unsupported args are used
+		}
+	}
+	
+	// Return the percentage
+	return value;
 }
 
 
--- a/src/nvidia.cc
+++ b/src/nvidia.cc
@@ -331,6 +331,31 @@ namespace {
 	nvidia_display_setting nvidia_display;
 }
 
+// Extract arguments for nvidiabar, etc, and run set_nvidia_type
+int scan_nvidia_args (struct text_object *obj, const char *args, unsigned int special_t) {
+	const char *arg = args;
+
+	switch (special_t) {
+		case BAR:
+			arg = scan_bar(obj, arg, 100);
+			break;
+		case GRAPH:
+			arg = scan_graph(obj, arg, 100);
+			break;
+		case GAUGE:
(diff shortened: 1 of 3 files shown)


MattSturgeon:
Ok, so since my `get_nvidia_barval` test function runs, that means `new_bar(obj, p, p_max_size, (*obj->callbacks.barval)(obj));` ([conky.cc:866][newbar]) in the `generate_text` loop is running correctly. 

As I understand it, if everything works correctly, `new_bar` should go on to run `new_bar_in_x11` which will add a `special_data` with `type = BAR`.

If I'm following the code correctly this should then get picked up by `generate_text` when it checks for "specials".

When I read the code I don't see any missing steps, so therefore my test function `get_nvidia_barval` must be returning a value that causes issues?


[newbar]: https://github.com/brndnmtthws/conky/blob/master/src/conky.cc#L866

MattSturgeon:
Ok, finally found what I was missing: you have to run `scan_bar` to parse bar-specific options (height and width) which adds necessary stuff to the object.

Next step is to actually integrate with the nvidia module, then maybe learn to program after!

src/conky.cc
```diff
@@ -293,6 +293,7 @@ static void print_version(void)
 #endif /* BUILD_IBM */
 #ifdef BUILD_NVIDIA
                 << _("  * nvidia\n")
+                << _("  * nvidiabar\n")
```
MattSturgeon: Don't think this is needed

src/nvidia.cc
```diff
@@ -704,6 +728,13 @@ void print_nvidia_value(struct text_object *obj, char *p, int p_max_size)
 	
 }
 
+double get_nvidia_barval(struct text_object *obj) {
+	double val;
+	val = 50; //Dummy test value
+	printf("Running get_nvidia_barval: %lf\n", val);// Check
```
brndnmtthws: Please remove this.

src/nvidia.cc
```diff
@@ -331,6 +331,30 @@ namespace {
 	nvidia_display_setting nvidia_display;
 }
 
+// Extract arguments for nvidiabar, etc, and run set_nvidia_type
+int scan_nvidia_args (struct text_object *obj, const char *args, unsigned int special_t) {
+	const char *arg = args;
+
+	switch (special_t) {
+		case BAR :
```
brndnmtthws: Remove whitespace before `:`.

src/nvidia.cc
```diff
@@ -331,6 +331,30 @@ namespace {
...
+			arg = scan_graph(obj, arg, 100);
+			break;
+		case GAUGE :
+			arg = scan_gauge(obj, arg, 100);
+			break;
+		default:
+			return 1;
+	}
+
+	if (!arg)     // Return error if no argument
```
brndnmtthws: Use `{}` around statement.

src/conky.cc
brndnmtthws: Agreed.

MattSturgeon:
I squashed the changes together, rebased off master and incorporated your feedback.

--------------

I'm still undecided how best to approach restructuring `print_nvidia_value()` to integrate `get_nvidia_barval()`.

I could seperate the query `switch` statement into a shared funtion that either returns an `int` (letting `print_nvidia_value()` handle string values separately) or a struct with both an `int` and a `char*`.

Alternatively I could implement a separate `switch` (perhaps based on `nvs->attribute`) for `get_nvidia_barval()` that only deals with suitable nvidia queries.

Could even go the OO way and create a class that has a `print_to_buffer()` method as well as a `get_percentage()` method.

MattSturgeon:
I squashed the changes together, rebased off master and incorporated your feedback @brndnmtthws .

I've also got a basic implementation of `get_nvidia_barval()` now.

MattSturgeon:
Is there anything else I need to change? Feature wise it would be "nice to have" support for more than just the `util`s but I don't think that should block this.

mxmlnkn:
I tested it and it seems to work. I especially like the GPU utilization bar, I would definitely use that (I am using that now :) as I have compiled it from source anyway). I tested that one with a small CUDA example.

There are some things I missed in this pull request while trying this for a few minutes:
 - a documentation how to use it and what options are available (ok there are in-code comments, but those are only read by developers). I'm not sure though where the correct place for the current documentation is. The wiki here lacks heavily and the [sourceforge site](http://conky.sourceforge.net/variables.html) is outdated, ...
 - all other bars like `membar` or `cpubar` have options for `width` and `height` see [here](http://conky.sourceforge.net/variables.html). Those don't seem to work with `nvidiabar`
    ```
    conky: nvidiabar: invalid argument specified: 'gpuutil 10,100'
    ```
   Weirdly `nvidiagraph` supports those options (only tested width, height, gradient colors)
 - what does it look like. Here are 

Here is how I tried it out:
```Shell
${color2}${voffset 2}${hr 1}$color
${color2}${alignc}GPU$color
${color2}${voffset -5}${hr 1}$color
${color2}GPU Temp${goto 70}${color0}${nvidia temp}C${goto 125}${color2}Fan Speed${goto 195}${color0}${execi 10 nvidia-settings -q [fan:0]/GPUCurrentFanSpeed -t} %
${color2}GPU Clock${goto 70}${color0}${nvidia gpufreq} MHz${goto 125}${color2}Mem Clock${goto 195}${color0}${nvidia memfreq} MHz
${color2}Mem Used${goto 70}${color0}${execi 10 nvidia-settings -q [gpu:0]/UsedDedicatedGPUMemory -t} / ${exec nvidia-settings -q [gpu:0]/TotalDedicatedGPUMemory -t} MiB
Load ${goto 55}${nvidiabar gpuutil}
VRAM ${goto 55}${nvidiabar memutil}
RAM  ${goto 55}${nvidiabar membwutil}
Video${goto 55}${nvidiabar videoutil}
PCIe ${goto 55}${nvidiabar pcieutil}
Fan  ${goto 55}${nvidiabar fanlevel}
Temp ${goto 55}${nvidiabar gputemp}
Load: ${nvidiagraph gpuutil 50,120 F0F0F0 0ABFFF}
${goto 15}Load${goto 55}VRAM${goto 95}RAM${goto 130}Videos${goto 175}PCIe${goto 218}Fan${goto 253}Temp
${nvidiagauge gpuutil}${nvidiagauge memutil}${nvidiagauge membwutil}${nvidiagauge videoutil}${nvidiagauge pcieutil}${nvidiagauge fanlevel}${nvidiagauge gputemp}
```
And this is what it looks like:
nvidia-bar

Nooby4Ever:
New keywords should be added here:
```
/extras/nano/conky.nanorc
/extras/vim/syntax/conkyrc.vim
```

Documentation should be added in the right files located in[ `/doc`](https://github.com/brndnmtthws/conky/tree/master/doc), as far as I know these are the most up-to-date docs on conky (accessible using: `man conky`, if they are compiled).

Btw nice work!
(although sadly I don't have the time to play with it at the moment)

MattSturgeon:
@mxmlnkn Just double checkedc and setting the dimensions does work... Currently I'm using the exec* syntax, where *bar/*gauge use `nvidiabar (height),(width) command` and *graph uses `nvidiagraph command (height),(width) (gradient color 1) (gradient color 2) (scale) (-t) (-l)`

Perhaps this should be standardised across the API - but that is separate (and would break backwards compatibility...) I've opened #246 regarding this.

@Nooby4Ever Thanks for the pointers - always appreciated! I'll do my best to look into adding documentation soon(ish)

mxmlnkn:
@MattSturgeon Ah yes, you are completely right. As you can see in the error message I tried `nvidiabar command (height),(width)` instead of `nvidiabar (height),(width) command`, not it works. Thanks for opening the issue.

brndnmtthws:
Thanks for this!

PR #219, opened 2016-02-21
solve double counting at start issue

mxmlnkn:
In issue #218 I noticed a completely different bug, where after a restart the total traffic displayed changed from 60GB to 69GB. That was because the actual total traffic was actually 35GB. Before this commit `ns->recv` was set to `r` on init and also some lines down below `r` was additionally added to `ns->recv` resulting in a doubling of the total traffic read from `/proc/net/dev`


brndnmtthws:
Thanks!

PR #226, opened 2016-03-16
Make evaluate useful

inzanity:
Make evaluate actually double-evaluate itself, allowing for constructs
such as: `${eval $${addr ${gw_iface}}}' to work.

### DIFF
--- a/src/common.cc
+++ b/src/common.cc
@@ -553,7 +553,9 @@ void print_cached(struct text_object *obj, char *p, int p_max_size)
 
 void print_evaluate(struct text_object *obj, char *p, int p_max_size)
 {
-	evaluate(obj->data.s, p, p_max_size);
+	std::vector<char> buf(max_user_text.get(*state));
+	evaluate(obj->data.s, &buf[0], buf.size());
+	evaluate(&buf[0], p, p_max_size);
 }
 
 int if_empty_iftest(struct text_object *obj)


Nooby4Ever:
You might want to update the documentation, maybe with an example, docs for `eval` are located in[ `/doc/variables.xml`](https://github.com/brndnmtthws/conky/tree/master/doc/variables.xml
)

inzanity:
This patch actually makes `eval` work as documented, but I guess adding an example would not hurt.

Nooby4Ever:
Oh I see, sorry I had interpreted your pull request wrong. Though you enabled the use of ${...} within ${...} within ${...} instead of fixing the normal/documented behaviour. My bad.

michalrus:
Big :+1: to this!

michalrus:
@brndnmtthws, what do you think of this? Thank you!

michalrus:
Ohmigosh, thaaaaaaank you! That’s awesome! :heart: :revolving_hearts: :heart_eyes: :heart_eyes_cat:

src/common.cc
```diff
@@ -553,7 +553,9 @@ void print_cached(struct text_object *obj, char *p, int p_max_size)
 
 void print_evaluate(struct text_object *obj, char *p, int p_max_size)
 {
-	evaluate(obj->data.s, p, p_max_size);
+	std::vector<char> buf(max_user_text.get(*state));
```
michalrus: @brndnmtthws, @inzanity, having second thoughts here. Should it be `text_buffer_size` instead of `max_user_text`? :astonished:

brndnmtthws: Yeah I think you're right. Want to submit a PR?

michalrus: Sure, @brndnmtthws, see #260. :smile_cat:

PR #227, opened 2016-03-23
Improve to_bytes to work with input values with a decimal point

alessandro-g89:
Adjust the parsing of to_bytes to work correctly with non-integer input values, e.g. 1.24 MiB/s. Such kind of values may be returned by diskio_read and diskio_write for example.


brndnmtthws:
Thanks!

PR #230, opened 2016-03-31
Update immediately when the root pixmap changes

luebking:
Second patch prefers the monotic timer, which cannot run backwards (and is cheaper to fetch ;-)


aandergr:
This breaks #209.
get_time() is used to schedule next update, which since #209 is aligned to real time. This alignment breaks when using monotonic clock.
The exact commit which breaks this functionality is 0edb520. Why does it change clock source to CLOCK_MONOTONIC?

PR #232, opened 2016-04-04
Parse quoted execgraph cmd before parsing options

marcpayne:
When parsing the argument string for graph objects, check if there is a double-quoted command in the beginning of the string and extract it before parsing any options. This prevents items in a command from being improperly recognized as graph options. It should also eliminate the need to place commands with spaces into a separate script.

Also, fall back on default_graph_{height,width} instead of using hard-coded values so that the default values set in a user's config will work.


brndnmtthws:
Looks good, thanks!

PR #233, opened 2016-04-05
BUG: Fix race conditions in signal handling

layus:
SIGHUP, SIGUSR1, SIGINT and SIGTERM are no longer lost to race conditions.

In particular, a SIGHUP can no more cancel a SIGTERM.
This could happen before because only the last received signal was stored in a global variable.

This also resets the global variables tracking signals before handling these signals.
The previous code cleared them at the end of the loop, thus possibly clearing a signal received in the meantime.

### DIFF
--- a/src/conky.cc
+++ b/src/conky.cc
@@ -192,7 +192,7 @@ int top_io;
 #endif
 int top_running;
 static conky::simple_config_setting<bool> extra_newline("extra_newline", false, false);
-static volatile int g_signal_pending;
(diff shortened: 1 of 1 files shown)


layus:
@brndnmtthws, this bug is quite critical as conky may miss SIGTERM signals and ignore shutdown of the machine. This delays the entire shutdown of the machine until conky gets SIGKILLed after some system-specific timeout.

brndnmtthws:
Makes sense. Thanks!

PR #237, opened 2016-04-19
Make conky resolve symlinks before using them.

ticpu:
This will fix an issue where someone wants to use a symlink like /dev/disk/by-id/wwn-* in its conkyrc.  Usually, those links are pointed to ../../sd? and will break when used with readlink since conky cwd can be anywhere.

I made sure to replace readlink by realpath where diskio/diskiograph were in use.

I made sure to test conky with symlink, real device and label:.

### DIFF
--- a/src/core.cc
+++ b/src/core.cc
@@ -122,17 +122,15 @@
  */
 const char *dev_name(const char *path)
 {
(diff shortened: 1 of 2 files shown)


ticpu:
New PR in its own branch and GPG signed.

PR #254, opened 2016-05-10
Fix the `symbol stdscr not found` compilation error

lgeorget:
The compilation fails when the symbols of terminfo are not present in
the libncurses.so. This can happen when the terminfo library libtinfo.so
is compiled separately. Using pkg_config is the correct approach to
handle all cases.

Addresses #167 
See also https://bugs.gentoo.org/show_bug.cgi?id=562154


brndnmtthws:
Thanks!

PR #257, opened 2016-05-13
Increase info.cpu_count for single-core systems

icanhasserver:
On single-core systems, sscanf() does't match, as the sysfs file contains the value 0 instead of i.e. 0-7 for multi-core systems.

### DIFF
--- a/src/linux.cc
+++ b/src/linux.cc
@@ -820,9 +820,10 @@ void get_cpu_count(void)
 		}
 
 		if (sscanf(buf, "%*d-%d", &highest_cpu_index) == 1) {
-			info.cpu_count = highest_cpu_index + 1;
+			info.cpu_count = highest_cpu_index;
 		}
 	}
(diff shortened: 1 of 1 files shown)


brndnmtthws:
Thanks :)

Didn't know anyone was still using machines with 1 CPU, good to know!

thereallightbeam:
I think this issue (#257/#258) should be revisited because users, including myself, are still experiencing segmentation faults due to this issue. Downgrading to conky-1.10.1-2 fixes the issue, and with conky-1.10.2-1 removing the CPU output settings ($cpu, $cpubar) from .conkyrc file allows Conky to work. Neither solution is ideal. 

In the following discussion, you will notice that some users have reported the issue when using Virtualbox, however the issue also occurs on my single-core AMD machine with integrated graphics.

https://forum.manjaro.org/t/troubleshooting-conky-for-jwm-edition/3713/36

Thanks!

Issue #269, opened 2016-05-26
nvidia: segfault when trying to use $nvidia without a nvidia card installed


MattSturgeon:
I've been working on a few more changes to `$nvidia` and updating the docs (see my various `nvidia-` branches) and noticed a segfault when I tried using my nvidia conky.conf on my non-nvidia laptop.

I suspect this will be fixed by having `$nvidia` check to see if the target GPU exists before continuing (and spitting out a `CRIT_ERR` if not), however I've not really investigated this yet.

MattSturgeon:
Done a bit of testing and I can confirm the segfault is caused by [calling `XNVCTRLQueryTargetAttribute`](/brndnmtthws/conky/blob/master/src/nvidia.cc#L585).

Oddly `XNVCTRLQueryTargetStringAttribute` causes no errors (other than a lot of numbers being printed as "N/A")

Need to read the [docs](/NVIDIA/nvidia-settings/blob/master/doc/NV-CONTROL-API.txt) and work out the best way to check for a GPU.

PR #271, opened 2016-05-27
nvidia: Better sanity checks when running queries (Issue #269)

MattSturgeon:
Introduced a function that checks whether the target we are querying exists (and how many of the target there are).

**Warning:** Since I'm away from my main PC I was unable to test for regressions with a nvidia card installed. I have tested that it fixes the segfault in #269 but hopefully the code is simple enough not to cause issues :)

### DIFF
--- a/src/nvidia.cc
+++ b/src/nvidia.cc
@@ -574,15 +574,34 @@ int set_nvidia_type(struct text_object *obj, const char *arg)
 	return 0;
 }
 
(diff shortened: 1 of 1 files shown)


MattSturgeon:
Amended the commit removing the `CRIT_ERR` as I decided it was probably overkill. (old commit is b32af1309c4222e0288f9e478be9b4c50dc5f587)

PR #278, opened 2016-06-04
Nvidia updates, docs and fixes

MattSturgeon:
Having finally tested my previous PRs (#270 & #271) I have tidied up the commits and combined them into this branch.

*  36aeb1b Adds the other graph compatible arguments to `get_nvidia_barval` and also (finally) re-writes the docs for the nvidia object.

My only concern with this is the spacing of the arguments columns in the docs - I think this can only be sorted with stylesheets, but it's a minor issue.

*  dd7ffd3 Adds more thorough error checking and printing to the nvidia object
*  b294f24 fixes Issue #269

### DIFF
--- a/src/nvidia.cc
+++ b/src/nvidia.cc
@@ -43,7 +43,6 @@
  *   so that all quirks are located there
  * - Implement nvs->print_type to allow control over how the value is printed
  *   (int, float, temperature...)
- * - Rename set_nvidia_type() to set_nvidia_query (requires changes in core.cc)
  * 
  * Showcase (conky.conf):
  * --==| NVIDIA | ==--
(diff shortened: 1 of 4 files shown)


src/nvidia.cc
```diff
@@ -354,28 +384,11 @@ int scan_nvidia_args (struct text_object *obj, const char *args, unsigned int sp
 		case GAUGE:
 			arg = scan_gauge(obj, arg, 100);
 			break;
-		default:
-			return 1;
 	}
-
+
```
brndnmtthws: Whitespace.

src/nvidia.cc
```diff
@@ -569,20 +586,41 @@ int set_nvidia_type(struct text_object *obj, const char *arg)
...
 			return 1;
 	}
 	return 0;
 }
 
+// Check we are able to query tid and return the amount of targets present
+static inline int get_nvidia_target_count(Display *dpy, TARGET_ID tid){
+	int num_tgts;
+	if (!XNVCTRLQueryTargetCount(dpy, translate_nvidia_target[tid], &num_tgts)) {
+		//NORM_ERR("%s: Failed to query number of nvidia targets (dpy: %d, tid: %d)", __func__, dpy, tid);
```
brndnmtthws: Why's this commented out?

brndnmtthws:
Couple minor comments, but overall it looks good.

src/nvidia.cc
MattSturgeon: > Why's this commented out?

Originally the `CRIT_ERR`s in `get_nvidia_value` and `get_nvidia_string` were `NORM_ERR`s, so having two errors would spam the console a lot.

My mindset was to leave this commented, so that if anyone was struggling to debug something they had the option of uncommenting an extra error to see whether XNVCTRLQueryTargetCount had failed or simply set `num_tgts` to something < 1.

--------

I'll amend the commit to save the number of targets and print it in the `CRIT_ERR`s instead of having a secret error. Something like:

```C++
int target_count = get_nvidia_target_count(dpy, tgt);

if (target_count <1) CRIT_ERR("blah, blah, blah, (blah, target_count: %d)", target_count);
```

MattSturgeon:
> Couple minor comments, but overall it looks good.

Tidied up my whitespace, rabased off of master and rewrote b294f24 as 545ee56 `"Better sanity checks when running queries"`.

brndnmtthws:
Lovely, thanks!

PR #287, opened 2016-06-19
Use coordinate offsets to draw shades/outlines.

Ovnuniarchos:
Previously, a ${goto} would force the x-coordinate of the shade/outline drawing.


brndnmtthws:
Can you describe how it was tested?

Ovnuniarchos:
I compared the master version:
uncorrected
with the corrected version:
corrected
This is the config file that generates both images (renamed to .txt).
[conky.txt](https://github.com/brndnmtthws/conky/files/362369/conky.txt)

No formal testing was done.

PR #292, opened 2016-07-03
Fix $scroll'ing of UTF-8 strings (for v1.9.1).

AlexeyBond:
Fix issue #129 for version 1.9.1

Strings in UTF-8 mode are shifted by characters (not by bytes).

Differences from aidecoe's patch are:

* Strings are shifted by characters (in original pach they are still shifted by byte and bytes that are trailing bytes of UTF-8 character are skipped when found in begining of string) so the same amount of characters is displayed at any moment.

* Behavior is changed only in UTF-8 mode so scroll will work correct with other encodings.

### DIFF
--- a/src/scroll.c
+++ b/src/scroll.c
@@ -40,9 +40,12 @@ struct scroll_data {
 	long resetcolor;
 };
 
-int utf8_charlen(char c) {
+static int scroll_character_length(char c) {
 	unsigned char uc = (unsigned char) c;
 	int len = 0;
+
+	if (!utf8_mode)
+		return 1;
 
 	if (c == -1)
 		return 1;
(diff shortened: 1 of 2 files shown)


brndnmtthws:
Can you describe the change in a little detail? And how it fixes the problem?

src/scroll.c
```diff
@@ -100,28 +139,66 @@ void print_scroll(struct text_object *obj, char *p, int p_max_size, struct infor
 				break;
 		}
 	}
+
+	bufLength = strlen(buf);
```
brndnmtthws: Use `strnlen`.

brndnmtthws:
It would be nice if you could port this to `master`, although I'm not sure if it's necessarily required.

PR #299, opened 2016-07-16
goto action for ncurses mode

roy-mustang:
Hi,
I implemented a version of the goto object for the out_to_ncurses mode. In this case, the position to go to is expressed as columns rather than pixels.

### DIFF
--- a/src/conky.cc
+++ b/src/conky.cc
@@ -61,6 +61,9 @@
 #include "imlib2.h"
 #endif /* BUILD_IMLIB2 */
 #endif /* BUILD_X11 */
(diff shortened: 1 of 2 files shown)


brndnmtthws:
Cool! Can you attach a screenshot and describe how you tested it?

roy-mustang:
Here's a two-column layout. Since some values can change in width, the second column can not be perfectly aligned unless I use a goto.
screenshot

This is the configuration I used to make the screenshot:  [conkyrc.txt](https://github.com/brndnmtthws/conky/files/371543/conkyrc.txt)

PR #325, opened 2016-09-26
update url to get NOAA METAR data

krakenfury:
NOAA moved their METAR data to this URL.  I changed the URL in the code, compiled the source with the weather_metar flag enabled, added some weather to the default config and it worked as it did before.


gamecss (commented):
That's good

PR #327, opened 2016-09-28
FreeBSD changes

madpilot78:
I have included in this request a few changes I needed to update hte FreeBSD conky port to the latest version.

I added each change as a separate commit with explanation in the comment.

Hope these are useful and can be imported together.


brndnmtthws:
Thanks for the BSD love.

PR #330, opened 2016-10-04
Try more than one sysctl to get CPU temperature.

madpilot78:
Suggested by:	walter@pelissero.de
Reference:	https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=210235

### DIFF
--- a/src/freebsd.cc
+++ b/src/freebsd.cc
@@ -426,12 +426,13 @@ double get_acpi_temperature(int fd)
 	int temp;
 	(void)fd;
 
-	if (GETSYSCTL("hw.acpi.thermal.tz0.temperature", temp) == 0)
+	if (GETSYSCTL("hw.acpi.thermal.tz0.temperature", temp) == 0) {
 		return KELVTOC(temp);
-	if (GETSYSCTL("dev.cpu.0.temperature", temp) == 0)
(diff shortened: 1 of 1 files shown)


src/freebsd.cc
```diff
@@ -426,13 +426,15 @@ double get_acpi_temperature(int fd)
...
 	(void)fd;
 
-	if (GETSYSCTL("hw.acpi.thermal.tz0.temperature", temp)) {
-		fprintf(stderr,
-			"Cannot read sysctl \"hw.acpi.thermal.tz0.temperature\"\n");
-		return 0.0;
-	}
-
-	return KELVTOC(temp);
+	if (GETSYSCTL("hw.acpi.thermal.tz0.temperature", temp) == 0)
```
brndnmtthws: Can you include braces `{}`?

src/freebsd.cc
```diff
@@ -426,13 +426,15 @@ double get_acpi_temperature(int fd)
...
+		return KELVTOC(temp);
+	if (GETSYSCTL("dev.cpu.0.temperature", temp) == 0)
```
brndnmtthws: Also use `else if` instead, to be more idiomatic.

madpilot78: Done too.

PR #344, opened 2016-11-13
Cached versions of get_nvidia_value and get_nvidia_string_value.

h3nnn4n:
- Removed the TODO topic about using static variables as a cache for some values;
 - Fixed a typo in the sample nvidia config;
 - `static int cache_nvidia_value(TARGET_ID tid, ATTR_ID aid, Display *dpy, int *value);` queries the value for the first time, if it was already queried before just returns it;
 - `static int cache_nvidia_string_value(TARGET_ID tid, ATTR_ID aid, char *token, SEARCH_ID search, int *value, int update);` If `update` is set to 1, then the function stores `value` in the static variable associated with `token` and `search`. If `update` is set to 0, then the function checks if the `token` and `search` associated variable has already been queried, if not it does not and `get_nvidia_string` does its usual stuff.

### DIFF
--- a/src/nvidia.cc
+++ b/src/nvidia.cc
@@ -707,6 +707,8 @@ static int cache_nvidia_string_value(TARGET_ID tid, ATTR_ID aid, char *token, SE
 	static int memclockmax = -1;
 	static int memTransferRatemin = -1;
 	static int memTransferRatemax = -1;
+	static int perfmin = -1;
+	static int perfmax = -1;
 
 	if (update) {
 		if (strcmp(token, (char*) "nvclockmin") == 0 && nvclockmin < 0){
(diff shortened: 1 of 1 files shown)


src/nvidia.cc
```diff
@@ -618,6 +616,38 @@ void check_nvidia_target_count(Display *dpy, TARGET_ID tid, ATTR_ID aid)
...
+			if(!dpy || !XNVCTRLQueryTargetAttribute(dpy, translate_nvidia_target[tid], 0, 0, translate_nvidia_attribute[aid], value)){
+				NORM_ERR("%s: Something went wrong running nvidia query (tid: %d, aid: %d)", __func__, tid, aid);
+				return -1;
+			}
+			gputempthreshold = *value;
+		} else {
+			*value = gputempthreshold;
+		}
+	} else {
+		//CRIT_ERR(NULL, NULL, "TRETA:");
```
brndnmtthws: Might as well remove this.

h3nnn4n: Done.

brndnmtthws:
Thanks!

PR #356, opened 2017-01-14
Fix an out-of-range error in new_graph

labath:
The code was multiplying the index with the size of the element, and
then adding it to the typed pointer (resulting in a double
multiplication and an OOB access).

Replace the buggy code with a slightly safer c++ alternative.


brndnmtthws:
Makes sense. Thanks!

PR #357, opened 2017-01-16
fs.cc, linux.cc Use /proc/mounts for list of currently mounted fs

zakame:
Per manpage of mount(1) in newer util-linux:

>The programs mount and umount traditionally maintained a list of currently mounted filesystems in the file /etc/mtab.  This real mtab file is still supported, but on current Linux systems it is better to make it a symlink to /proc/mounts instead, because a regular mtab file maintained in userspace cannot reliably work with namespaces, containers and other advanced Linux features.

In most new Linux (e.g. Debian) /etc/mtab is already symlinked to /proc/mounts.

See https://github.com/karelzak/util-linux/commit/e778642a9eb96975fcf3baa61dfa10add628af1a for the gory details.

### DIFF
--- a/src/fs.cc
+++ b/src/fs.cc
@@ -151,13 +151,13 @@ void get_fs_type(const char *path, char *result)
 #else				/* HAVE_STRUCT_STATFS_F_FSTYPENAME */
 
 	struct mntent *me;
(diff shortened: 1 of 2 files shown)


brndnmtthws:
Thanks!

PR #359, opened 2017-01-23
eve: fix various issues

bzy-xyz:
Fixes for various issues in the current EVE module.

* Since the EVE module was contributed, CCP has changed how the API works and
what the correct API URL is. So the module had to be updated to respect the
changes made on the EVE side.

* Additionally, undefined behavior was occurring for characters without active
skill training queue, as is common for folks with multiple characters on a
single account, and was also occurring when Conky was reloaded following a
config file change. Now the `$eve` variable should just contain an empty
string if the character isn't training a skill (to make it detectable by
`$is_empty`), and the module should do The Right Thing (tm) when the
config file is changed while Conky is running.

* The time remaining text has been adjusted to be slightly less verbose.

* The module should do the right thing for up to 64 distinct characters.
Presently, though, every additional character requires a blocking CURL
request, meaning that querying many characters is currently not ideal. A
rewrite of the module to use asynchronous requests is probably in order, but
will be left for a future PR.

* On Linux `gettimeofday()` doesn't return valid timezone data; we use
`localtime()` and the `tm_gmtoff` instead. Training times are no longer
incorrect as a result.

* Sometimes the EVE API will return spurious non-XML tags as part of the
response. This previously caused the XML parser to return a null document.
We tell libxml to `XML_PARSE_RECOVER` to try to recover from these errors
where they occur on real data.

### DIFF
--- a/doc/variables.xml
+++ b/doc/variables.xml
@@ -1103,11 +1103,13 @@
             <command>
                 <option>eve</option>
             </command>
-            <option>api_userid api_key character_id</option>
+            <option>api_keyID api_vCode character_id</option>
         </term>
-        <listitem>Fetches your currently training skill from the
-        Eve Online API servers (http://www.eve-online.com/) and
-        displays the skill along with the remaining training time. 
+        <listitem>Fetches a character's currently training skill from
+        the Eve Online API servers (http://www.eveonline.com/) and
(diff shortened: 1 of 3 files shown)


brndnmtthws:
Thanks!

PR #362, opened 2017-01-26
enable persistent view of irc messages

deep-42-thought:
* remove directory from `libirc`-include
* give possibility to specify number-of-last-lines to show for irc instead of showing all lines since last time `${irc server(:port) #channel (number-of-lines)}`
* is backwards compatible: omitting number-of-lines resorts to old behaviour

### DIFF
--- a/src/irc.cc
+++ b/src/irc.cc
@@ -62,23 +62,27 @@ void addmessage(struct ctx *ctxptr, char *nick, const char *text) {
 	newmsg->text = (char*) malloc(strlen(nick) + strlen(text) + 4);	//4 = ": \n"
 	sprintf(newmsg->text, "%s: %s\n", nick, text);
 	newmsg->next = NULL;
+	int msgcnt = 1;
 	if(!lastmsg) {
 		ctxptr->messages = newmsg;
 	} else {
+		msgcnt++;
 		while(lastmsg->next) {
 			lastmsg = lastmsg->next;
+			msgcnt++;
+			if(msgcnt<0) {
+				NORM_ERR("irc: too many messages, discarding the last one.");
+				free(newmsg->text);
+				free(newmsg);
+				return;
+			}
 		}
 		lastmsg->next = newmsg;
 	}
 	if(ctxptr->max_msg_lines>0) {
-		int msgcnt = 0;
 		newmsg = ctxptr->messages;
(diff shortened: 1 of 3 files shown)


src/irc.cc
```diff
@@ -69,6 +70,22 @@ void addmessage(struct ctx *ctxptr, char *nick, const char *text) {
 		}
 		lastmsg->next = newmsg;
 	}
+	if(ctxptr->max_msg_lines>0) {
+		int msgcnt = 0;
+		newmsg = ctxptr->messages;
+		while(newmsg) {
```
brndnmtthws: There needs to be some sort of bounds checking here.

src/irc.cc
```diff
@@ -69,6 +70,22 @@ void addmessage(struct ctx *ctxptr, char *nick, const char *text) {
...
+			newmsg = newmsg->next;
+			msgcnt++;
+		}
+		msgcnt -= ctxptr->max_msg_lines;
+		while(msgcnt>0) {
```
brndnmtthws: Maybe here too?

src/irc.cc
deep-42-thought: I'm not sure, what variable should be checked. The loop is pretty similar to the one in line 68..70. Or do you mean, I should check for an overflow of `msgcnt`? This, however, only increases by one each time, the function is called, and gets truncated to `ctxptr->max_msg_lines` at the end of the function call iff `ctxptr->max_msg_lines>0`. Should I prepare the function for a changing `ctxptr->max_msg_lines`, so it may switch from 0 to some positive value while the current message count is larger than max(int)?

brndnmtthws: I'm wondering what happens if there's a case where there's no null pointer in `newmsg`. Rather than having to read through all the code, I prefer to see some kind of bounds check. This is how security vulnerabilities are created (although I'm not saying you created one, I'm just being conservative).

deep-42-thought: ok, I can assume some upper bound for max_msg_lines and msg_count - this should be reasonable. However, the loop in line 68 to 70 will last indefinitely if there is no null pointer, anyway.

deep-42-thought: Is the approach in bcaeffd something you had in mind?

PR #371, opened 2017-02-22
Fix drawing text shades

billie80:
I was notified in a Gentoo bug [1] that there are issues with drawing text shades. This was also mentioned in issue #331. According to the bug/issue this pull request should fix the problem.

[1] https://bugs.gentoo.org/show_bug.cgi?id=610444


Deluxo:
when this is merged?

PR #382, opened 2017-03-28
NVIDIA: Cleared TODO for the multi-GPU support

Erinor2:
## Added an optional parameter to the commands nvidia and nvidiabar for Multi-GPU support

- If the parameter is omited default value is 0 for GPU0
              The parameter is the GPU index starting at 0 as 0,1,2,3,...

              Note: Except for the command nvidiagraph where the parameter is not optional

- Updated the conky.conf showcase to see how to use the enhancement

- Enhanced the cache system for multi-GPU support.

- Added a new variable for the command nvidia:
             modelname return a string with the model name of the GPU

- Optimized execution:
            In every execution the query XNVCTRLQueryTargetCount was executed. 
            Now this query is only executed once the first time

- Updated the documentation to explain how to use the multi-gpu as well as the new command "modelname"

## Proof test of the cache system for cache-value and cache-string:

[nvidia_cache_value_test.txt](https://github.com/brndnmtthws/conky/files/877302/nvidia_cache_value_test.txt)

[nvidia_cache_string_test.txt](https://github.com/brndnmtthws/conky/files/877304/nvidia_cache_string_test.txt)

## Suggested update for wiki/configuration-variables:
[Configuration-Variables_new_text.txt](https://github.com/brndnmtthws/conky/files/877307/Configuration-Variables_new_text.txt)

## Conky running on four Nvidia 1080:
Advantages of this pull request

- Avoiding the execi commands and the optimization of the query has reduced dramatically the CPU usage. 
- Because of the resource usage reduction it is possible now to have it with an interval of 1s without impact

image

### DIFF
--- a/src/nvidia.cc
+++ b/src/nvidia.cc
@@ -35,8 +35,6 @@
  * Fonic <fonic.maxxim@live.com>
  * 
  * TODO:
- * - Add third argument to module to allow querying multiple GPUs/fans etc.,
- *   e.g. ${nvidia gputemp 2}, ${nvidia fanlevel 1}
  * - Move decoding of GPU/MEM freqs to print_nvidia_value() using QUERY_SPECIAL
  *   so that all quirks are located there
  * - Implement nvs->print_type to allow control over how the value is printed
@@ -44,23 +42,24 @@
  * 
  * Showcase (conky.conf):
  * --==| NVIDIA | ==--
- * GPU   ${nvidia gpufreq}MHz (${nvidia gpufreqmin}-${nvidia gpufreqmax}MHz)
- * MEM   ${nvidia memfreq}MHz (${nvidia memfreqmin}-${nvidia memfreqmax}MHz)
- * MTR   ${nvidia mtrfreq}MHz (${nvidia mtrfreqmin}-${nvidia mtrfreqmax}MHz)
- * PERF  Level ${nvidia perflevel} (${nvidia perflevelmin}-${nvidia perflevelmax}), Mode: ${nvidia perfmode}
- * VRAM  ${nvidia memutil}% (${nvidia memused}MB/${nvidia memtotal}MB)
- * LOAD  GPU ${nvidia gpuutil}%, RAM ${nvidia membwutil}%, VIDEO ${nvidia videoutil}%, PCIe ${nvidia pcieutil}%
- * TEMP  GPU ${nvidia gputemp}°C (${nvidia gputempthreshold}°C max.), SYS ${nvidia ambienttemp}°C
- * FAN   ${nvidia fanspeed} RPM (${nvidia fanlevel}%)
+ * GPU    ${nvidia gpufreq [gpu_id]}MHz (${nvidia gpufreqmin [gpu_id]}-${nvidia gpufreqmax [gpu_id]}MHz)
+ * MEM    ${nvidia memfreq [gpu_id]}MHz (${nvidia memfreqmin [gpu_id]}-${nvidia memfreqmax [gpu_id]}MHz)
+ * MTR    ${nvidia mtrfreq [gpu_id]}MHz (${nvidia mtrfreqmin [gpu_id]}-${nvidia mtrfreqmax [gpu_id]}MHz)
+ * PERF   Level ${nvidia perflevel [gpu_id]} (${nvidia perflevelmin [gpu_id]}-${nvidia perflevelmax [gpu_id]}), Mode: ${nvidia perfmode [gpu_id]}
+ * VRAM   ${nvidia memutil [gpu_id]}% (${nvidia memused [gpu_id]}MB/${nvidia memtotal [gpu_id]}MB)
+ * LOAD   GPU ${nvidia gpuutil [gpu_id]}%, RAM ${nvidia membwutil [gpu_id]}%, VIDEO ${nvidia videoutil [gpu_id]}%, PCIe ${nvidia pcieutil [gpu_id]}%
+ * TEMP   GPU ${nvidia gputemp [gpu_id]}°C (${nvidia gputempthreshold [gpu_id]}°C max.), SYS ${nvidia ambienttemp [gpu_id]}°C
+ * FAN    ${nvidia fanspeed [gpu_id]} RPM (${nvidia fanlevel [gpu_id]}%)
+ * OPENGL ${nvidia imagequality [gpu_id]} 
  * 
  * --==| NVIDIA Bars |==--
- * LOAD ${nvidiabar gpuutil}
- * VRAM ${nvidiabar memutil}
- * RAM ${nvidiabar membwutil}
- * VIDEO ${nvidiabar videoutil}
- * PCIe ${nvidiabar pcieutil}
- * Fan ${nvidiabar fanlevel}
- * TEMP ${nvidiabar gputemp}
+ * LOAD ${nvidiabar gpuutil [gpu_id]}
+ * VRAM ${nvidiabar memutil [gpu_id]}
+ * RAM ${nvidiabar membwutil [gpu_id]}
+ * VIDEO ${nvidiabar videoutil [gpu_id]}
+ * PCIe ${nvidiabar pcieutil [gpu_id]}
+ * Fan ${nvidiabar fanlevel [gpu_id]}
+ * TEMP ${nvidiabar gputemp [gpu_id]}
  * 
  */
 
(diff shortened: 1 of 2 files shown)


src/nvidia.cc
```diff
@@ -35,33 +35,52 @@
...
  * 
+ * --==| NVIDIA Graph |==-- (gpu_id is not optional in this case)
+ * LOAD  ${nvidiagraph gpuutil [height][,width] [gradient color 1] [gradient color 2] [scale] [-t] [-l] gpu_id}
+ * VRAM  ${nvidiagraph memutil [height][,width] [gradient color 1] [gradient color 2] [scale] [-t] [-l] gpu_id}
+ * RAM   ${nvidiagraph membwutil [height][,width] [gradient color 1] [gradient color 2] [scale] [-t] [-l] gpu_id}
+ * VIDEO ${nvidiagraph videoutil [height][,width] [gradient color 1] [gradient color 2] [scale] [-t] [-l] gpu_id}
+ * PCIe  ${nvidiagraph pcieutil [height][,width] [gradient color 1] [gradient color 2] [scale] [-t] [-l] gpu_id}
+ * Fan   ${nvidiagraph fanlevel [height][,width] [gradient color 1] [gradient color 2] [scale] [-t] [-l] gpu_id}
+ * TEMP  ${nvidiagraph gputemp [height][,width] [gradient color 1] [gradient color 2] [scale] [-t] [-l] gpu_id}
  */
```
MattSturgeon: The showcase was originally intended as an example `conky.conf` that a user could past into a text file and run to see how it works. It looks to me like you've re-written it in the style of a man page usage example.

Personally I'm not sure that `nvidia.cc` is the right place for either, but if we keep it I like the idea of a user being able to simply copy-paste to get a working example..

src/nvidia.cc
```diff
@@ -309,10 +334,35 @@ struct nvidia_s {
...
 	SEARCH_ID search;
+//  added new field for GPU id
+    int gpu_id;
+};
+
+//Cache by value
+struct nvidia_c_value {
+    int memtotal = -1;
+	int gputempthreshold = -1;
+};
```
MattSturgeon: Minor formatting spot: it looks like you've mixed tabs and spaces here. Probably best to do a quick find/replace (I think the file was tab-indented before)

src/nvidia.cc
```diff
@@ -827,27 +888,37 @@ void print_nvidia_value(struct text_object *obj, char *p, int p_max_size)
 	struct nvidia_s *nvs = static_cast<nvidia_s *>(obj->data.opaque);
 	int value, temp1, temp2;
 	char* str;
+	
+    Display *dpy = nvdisplay ? nvdisplay : display;
+    
+    //num_GPU calculated only once based on the physical target
+	static int num_GPU = get_nvidia_target_count(dpy, TARGET_GPU) - 1;
```
MattSturgeon: Shouldn't this also be done in `get_nvidia_barval`? Also why the -1? Is that related to line 905? If so would it be better to move that logic there?

Also, if the same error checking and gpu id reducing needs to be done in both print_nvidia_value and get_nvidia_barval, should it be moved into a separate inline function or macro?

fonic:
Hi there,
I'm the one who rewrote the nvidia module back in 2015 and thought I'll join the discussion. I'm really glad people still care about this module!

1. The idea of multi-GPU support is of course a good one. In fact, I already implemented this in 2015, but since I had no system with multiple cards at hand to test this and none of the other developers did, I never submitted the code.

2. Yes, the showcase was intended to be a copy&paste block to get started using the module without the need of studying the documentation in detail, and I think it should stay that way. I agree that the current location (i.e. in nvidia.cc) is not the right place for this - back in 2015 I could not find a suitable place, that's why I put it there so it is at least somewhere...

3. I'd recommend renaming `GPU_ID` in documentation to `gpuid` (i.e. lowercase and without underscore) to be in sync with the rest of the documentation. Also, since `GPU_ID` is optional, I suggest using brackets, which seems to be used to indicate optional arguments throughout the documentation. And finally, since there's more than one argument to the module now, I'd recommend the following structural change to make things easier to understand:

        nvidia metric (gpuid)
              Nvidia graphics card information queried using the XNVCtrl library.

              Argument 'metric' (mandatory):
              ...

              Argument 'gpuid' (optional):
              For multi-GPU systems, this argument may be used to specify which GPU to query (0, 1, ...). Defaults to 0.

4. Parsing of `GPU_ID` using deprecated/unsafe `atoi` is not a good idea, there are way better alternatives available, e.g. [listed here](https://stackoverflow.com/a/7021750). Personally, I prefer `strtonum`, but this is not available on all systems. Also, your code silently defaults to 0 when the value is out of range 0 <= guid <= max_gpu, but I think it's better for users to print an error and only default to 0 if the argument was not specified.

5. Judging from your screenshot, there seems to be a problem regarding the GPU clock. I don't believe your cards actually clock down to `9 MHz`... This may or may not be related to your changes, I'm not sure.

6. I really don't want to offend you, but I have to say that your pull request is very unclean. There are lot's of formatting issues (wrong indentation, mixed tabs/spaces, whitespace-only lines, different comment style) and there is even stuff included that doesn't belong to the pull request at all (e.g. commit `Update 2 of documentation`). This should be cleaned up.

Other than that, your changes look promising! It would be great if some testers with multi-GPU setups could join the effort to finish this. Sadly, I still can't test myself, I never had much use for more than one GPU in my computers...

BTW, if you don't mind my asking: what are you doing with those nice four GTX 1080s, computations or gaming?

plikhari:
Seriously - 4 GPUs huahahahahaha - probably very heavy animation rendering

fonic:
@plikhari:
> Seriously - 4 GPUs huahahahahaha - probably very heavy animation rendering
Very helpful. You do realize that this platform (GitHub) is used for development purposes and not a place to post random stuff, right?

src/nvidia.cc
Erinor2: Hello Matt,

It doesn't harm to have it there and it is pretty self explanatory.

Let see what the comrades prefer.

src/nvidia.cc
Erinor2: I am struggling with a lot of topic and one is what framework to use. Those changes were done using gedit out of the box.

Maybe you can give me some hints about what framework to use for developing? 

Actually I am using Ubuntu 16.04LT as it is certified by NVIDIA CUDA Framework. It was from the certified distributions, the one with higher linux kernel version.

I want the framework to solve my little issues like having multiple NVidia. Might be I get some Radeon Vega too in an heterogeneous environment if this the case I will add a module for them. 

About my main project I am not sure if I will branch GRUB2 or KERNEL, maybe both it is a boolean OR.

As I said those are the requirements I have in mind, if you have any idea about what framework to use it will be really welcome.

src/nvidia.cc
Erinor2: When I did it I  I was seeing my cards as an array zero based. Then it was more logical for me to have it that way.

But I agree that can make more sense when you see num_GPU you expect to have at least one. Therefore I ahve to move the -1 operation when reading the number of GPU as you have seen.

I change it.

Erinor2:
Hello Fonic,
1. It is nice you did this module because I started using the execi commands and it was a disaster if you look at the screen-shot the top one is the rose of my processor 22 cores 44 threads (INTEL XEON5 2699V4) and with execi I was really showing the load of useless work. But the module as it was only solved 25% of my needs then I cleaned the TODO. Now I have an idle situation even with a refresh rate of 1s. Thanks for the job done before.

2. I don't think it is wrong to have it there and in case of changing the code we can be sure we don not forget any case.

3. I use square brakets on the showcase to self explain they are optional. I like the concept for the documentation to make more clear. I'll do it.

4. Thanks for the tip about atoi I'll do the change. It is very long I don't program using C I have to update my skills to the new trends.

5.Here is the screen-shot of my NVIDIA X Server where you can see when idle the freq is in between 8-10MHz. It is a mechanism to reduce power consumption.

image

6. No offense comrade. I am still struggling to chose a framework for programming I was needing the changes then I just did it straight forward with gedit out of the box. 
With a simple editor you don't see if it is a space or a tab...
In the meanwhile I didn't synchronized upstream and then it was messy. First time I use github it was a learning.

BTW: You are right :) this machine is a simple battle field to set some gears together and test some improvements for a bigger one. 
Regards...

fonic:
Hi Erinor2,
thank you for taking my criticism as constructive, I think your modifications will make a fine addition to the module.

Regarding 5.:
I know that the core clock is lowered to save power, but not this low. The screenshot shows that the card you selected is in performance level 0, which has a core clock of 215 MHz to 683 MHz, so the core clock should be in that range. That seems to be a bug in the nvidia driver.

As for the 'framework':
I guess you mean _IDE_ instead of _framework_, i.e. a development environment. Personally, I use KDevelop, but there are also some nice IDEs available for Gnome, e.g. [Builder](https://wiki.gnome.org/Apps/Builder).

PR #383, opened 2017-03-29
Transparency: Double-buffer not taken into consideration when selecting X11 when using ccmake

Erinor2:
## Symptoms:
1) When combining window_argb_visual = true with double_buffer = true result in the following error:

conky: desktop window (1000003) is subwindow of root window (d4)
conky: window type - normal
conky: drawing to created window (0x1600002)
X Error of failed request: BadMatch (invalid parameter attributes)
Major opcode of failed request: 62 (X_CopyArea)
Serial number of failed request: 160
Current serial number in output stream: 163

2) When avoiding this combination Conky is not anymore transparent


Erinor2:
Please can you take a look why check failed? The change for me solve the transparency issue and the flickering when disabling double-buffer.

dbriba:
I use five conky with this combinaison (window_argb_visual = true and with double_buffer = true) and I have no problem my conky are transparent and I have no error.

Erinor2:
travis.yml was missing libimlib2-dev :)

Issue #377, opened 2017-03-10
Month name problem with UTF-8 and LANG=de_DE (while month March only)


bymoz089:
Following config for date string:
${time %a %d.%b %Y}

 gives a strange output with environment var LANG set to:
LANG="de_DE.UTF-8"

Screenshot:
![lang-de_month-problem-march](https://cloud.githubusercontent.com/assets/15787117/23793870/b83e4c14-058d-11e7-9443-c2ab97e0b437.jpg)

it should state:
Fr 10.Mär.2017
(end of the year string from 2017 is cut because of window dimensions, but this is not the problem)

AlexeyBond:
Looks like march is the only one month name of which in german language contains a character represented by more than one byte in UTF-8. So probably the problem is that conky with your configuration is not able to display any character behind first 127.

If you keep see strange characters when just put a plain text "Fr 10.Mär.2017" or "Frühling In Paris"(I have no idea what it means) in config then try the following:

- enable UTF-8 mode:
    `override_utf8_locale = true`
- enable xft fonts and set custom font:
    ```
    use_xft = true,
    font = 'DejaVu Sans Mono:size=7' -- or any other font
    ```

plikhari:
@bymoz089 

If you issue is resolved - can you perhaps close this issue - thanks

Shura0:
I have similar issue with Russian.

Config contains the lines
> override_utf8_locale = true
> use_xft = true,
> font = 'Bitstream vera sans mono:size=10

But it does not help.
OS is FreeBSD.

bymoz089:
Thank you for helping. `override_utf8_locale = true` fixes the Umlauts display issue, they are displayed now.

But for every Umlaut conky thinks the displayed string is longer than it actually is and cuts off the string, leaving empty space at the end. I will file a new bug, explaining this

PR #384, opened 2017-04-07
Fix scrolling of UTF-8 text.

### LINKED ISSUE #223: Conky scroll multibyte character splitting
I run i3 WM in arch. I feed/pipe conky into my i3bar, which is great. Runs great.
However, when listing the filename of songs, i3bar sometimes gives me:
json lexical error invalid bytes in utf8 string conky
I have found that by removing the scrolling part, I no longer have this issue. However, many of the names are extremely long, so this is inconvenient.
I saw mentioned there was an issue at one point with conky splitting characters mid-byte. Personally, I don't care if the whole name displays perfectly all the time, as long as it doesn't crash my i3 bar.
I'll attach my config.
I honestly assumed this issue would already be on here, but as I don't see it, I'll create one.

[i3conkyrc.txt](https://github.com/brndnmtthws/conky/files/162555/i3conkyrc.txt)

AlexeyBond:
Fix issue #129 (UPD: should fix #223 and #45 too).

Port of PR #292 to master with fixes for new scroll modes.

+ right scroll now works with color changes

### DIFF
--- a/src/conky.h
+++ b/src/conky.h
@@ -298,6 +298,12 @@ void generate_text_internal(char *, int, struct text_object);
 int percent_print(char *, int, unsigned);
 void human_readable(long long, char *, int);
 
+#ifdef BUILD_X11
+
+/* UTF-8 */
+extern conky::simple_config_setting<bool> utf8_mode;
+#endif
+
 /* maximum size of config TEXT buffer, i.e. below TEXT line. */
 extern conky::range_config_setting<unsigned int> max_user_text;
 
--- a/src/scroll.cc
+++ b/src/scroll.cc
@@ -35,6 +35,51 @@
 #include "x11.h"
 #include <vector>
 
+/**
+ * Length of a character in bytes.
+ * @param c first byte of the character
+ */
+inline int scroll_character_length(char c) {
+#ifdef BUILD_X11
+	if (utf8_mode.get(*state)) {
+		unsigned char uc = (unsigned char) c;
+		int len = 0;
+
+		if (c == -1)
+			return 1;
+
+		if ((uc & 0x80) == 0)
+			return 1;
+
+		for (len = 0; len < 7; ++len) {
+			if ((uc & (0x80 >> len)) == 0) {
+				break;
+			}
+		}
+
+		return len;
+	}
+#endif
+
+	return 1;
+}
+
+/**
+ * Check if a byte should be skipped when counting characters to scroll text to right.
+ */
+inline bool scroll_check_skip_byte(char c) {
+#ifdef BUILD_X11
+    if (utf8_mode.get(*state)) {
+        // Check if byte matches UTF-8 continuation byte pattern (0b10xxxxxx)
+        if ((c & 0xC0) == 0x80) {
+            return true;
+        }
+    }
+#endif
+
+    return SPECIAL_CHAR == c;
+}
+
 #define SCROLL_LEFT 1
 #define SCROLL_RIGHT 2
 #define SCROLL_WAIT 3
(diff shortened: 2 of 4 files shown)


Erinor2:
Hello Alexey,

Will #377 also be fixed?

Regards,
Philippe

src/conky.h
```diff
@@ -298,6 +298,12 @@ void generate_text_internal(char *, int, struct text_object);
 int percent_print(char *, int, unsigned);
 void human_readable(long long, char *, int);
 
+#ifdef BUILD_X11
```
AlexeyBond: Variable `utf8_mode` is defined inside of `#ifdef BUILD_X11 ... #endif` in `conky.cc`, so if it will be used when `BUILD_X11` is not defined then the program won't link.

src/scroll.cc
```diff
@@ -35,6 +35,51 @@
 #include "x11.h"
 #include <vector>
 
+/**
+ * Length of a character in bytes.
+ * @param c first byte of the character
+ */
+inline int scroll_character_length(char c) {
+#ifdef BUILD_X11
```
AlexeyBond: (see answer to comment above)

Scroll still will work in console mode, but without UTF-8 support.

src/scroll.cc
```diff
@@ -35,6 +35,51 @@
...
+#endif
+
+	return 1;
+}
+
+/**
+ * Check if a byte should be skipped when counting characters to scroll text to right.
+ */
+inline bool scroll_check_skip_byte(char c) {
+#ifdef BUILD_X11
```
AlexeyBond: (see responses to comments above)

src/scroll.cc
```diff
@@ -50,6 +95,45 @@ struct scroll_data {
...
+/**
+ * Get count of characters to right from (sd->start) position.
+ */
+static unsigned int scroll_count_characters_to_right(struct scroll_data* sd, const std::vector<char>& buf) {
+    unsigned int n = 0;
+    int offset = sd->start;
+
+    while ('\0' != buf[offset]) {
+        offset += scroll_character_length(buf[offset]);
+        ++n;
```
AlexeyBond: Fixed.

src/scroll.cc
```diff
@@ -50,6 +95,45 @@ struct scroll_data {
...
+    }
+
+    return n;
+}
+
+static void scroll_scroll_left(struct scroll_data* sd, const std::vector<char>& buf, unsigned int amount) {
+    for (int j = 0; (j < amount) && (buf[sd->start] != '\0'); ++j) {
```
AlexeyBond: Fixed.

src/scroll.cc
```diff
@@ -50,6 +95,45 @@ struct scroll_data {
...
+        sd->start += scroll_character_length(buf[sd->start]);
+    }
+
+    if(buf[sd->start] == 0 || (unsigned) sd->start > strlen(&(buf[0]))) {
```
AlexeyBond: Fixed.

src/conky.h
ght: My bad, didn't notice that.

src/scroll.cc
```diff
@@ -111,11 +108,11 @@ static unsigned int scroll_count_characters_to_right(struct scroll_data* sd, con
 }
 
 static void scroll_scroll_left(struct scroll_data* sd, const std::vector<char>& buf, unsigned int amount) {
-    for (int j = 0; (j < amount) && (buf[sd->start] != '\0'); ++j) {
+    for (int i = 0; (i < amount) && (buf[sd->start] != '\0') && (sd->start < buf.size()); ++i) {
         sd->start += scroll_character_length(buf[sd->start]);
     }
 
-    if(buf[sd->start] == 0 || (unsigned) sd->start > strlen(&(buf[0]))) {
+    if (buf[sd->start] == 0 || sd->start > strlen(buf.data())) {
```
ght: Not sure, but as far as I understand it would be safer to use `strnlen` instead of `strlen`.

AlexeyBond: In [conky.cc#891](https://github.com/brndnmtthws/conky/blob/master/src/conky.cc#L891)(function `generate_text_internal` that generates content of `buf`) there is a `'\0'` written to the end of buffer, so if there is any unsafe call to `strlen` then that's [conky.cc#885](https://github.com/brndnmtthws/conky/blob/master/src/conky.cc#L885).

ght: Right, didn't see that.

Issue #396, opened 2017-05-29
Build fails with gcc7 due to missing include


oberon2007:
**gcc7** is less tolerant with non-standard and deprecated code than **gcc6** was 😉 .
As a consequence conky can no longer be compiled.
There are various warnings about dynamic exception of the kind
```
/src/semaphore.hh:40:36: warning: dynamic exception specifications are deprecated in C++11 [-Wdeprecated]
  semaphore(unsigned int value = 0) throw(std::logic_error)
```
and a lot of errors like
```
src/luamm.hh:39:15: error: ‘function’ in namespace ‘std’ does not name a template type
  typedef std::function<int(state *)> cpp_function;
```
or
```
src/luamm.hh:250:26: error: ‘cpp_function’ does not name a type; did you mean ‘lua_CFunction’?
   void pushclosure(const cpp_function &fn, int n);
```
full buildlog is [here](https://pastebin.com/u5saAbdY)

oberon2007:
In fact all the fatal errors result from **luamm.hh** file. The C/C++ - melange there doesn't seem to work out any longer ... 😜

LoneFox78:
The error is caused by a missing #include:

	diff -bur conky-1.10.6-orig/src/luamm.hh conky-1.10.6/src/luamm.hh
	--- conky-1.10.6-orig/src/luamm.hh      2016-12-04 17:13:57.000000000 +0200
	+++ conky-1.10.6/src/luamm.hh   2017-05-31 20:05:59.000000000 +0300
	@@ -28,6 +28,7 @@
	 #include <exception>
	 #include <stdexcept>
	 #include <string>
	+#include <functional>
	 
	 #include <lua.hpp>

oberon2007:
Awesome 😆 Thank you!

easysid:
This issue can now be closed as the patch is included in the repo

PR #402, opened 2017-07-17
Fix gcc-7 compilation error.

albert2004:
Fix for issue #396.
Also gentoo bug:
https://bugs.gentoo.org/show_bug.cgi?id=624104

Needed due to GCC include changes as stated in documentation:
Several C++ Standard Library headers have been changed to no longer include the <functional> header. As such, C++ programs that used components defined in <functional> without explicitly including that header will no longer compile.

### DIFF
--- a/src/luamm.hh
+++ b/src/luamm.hh
@@ -28,6 +28,7 @@
 #include <exception>
 #include <stdexcept>
 #include <string>
+#include <functional>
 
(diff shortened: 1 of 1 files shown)


andir:
Same fix would be appreciated for the NixOS move to gcc-7 (https://github.com/NixOS/nixpkgs/issues/31747)

PR #406, opened 2017-07-24
Add "detect_battery" configuration setting

### LINKED ISSUE #190: update_interval_on_battery does not work
I get this error when starting Conky:
conky: can't open /sys/class/power_supply/BAT0/uevent: Document or folder doesn't exist
conky: can't open /proc/acpi/battery/BAT0/state: Document or folder doesn't exist

This is because my old script had "update_interval_on_battery 5.0 BAT1", but in Conky 1.10 this is no longer valid syntax. Now it is "update_interval_on_battery = 5.0," and I cannot find a way to specify the correct device.

Although, it would probably be suitable for conky to detect that BAT0 does not exist, and use BAT1 by default, rather than requiring user configuration.

djpohly:
Allows configuration to specify which battery is used for detection of current on-battery/not-on-battery state.

Fixes #190.

### DIFF
--- a/src/conky.cc
+++ b/src/conky.cc
@@ -207,6 +207,8 @@ conky::range_config_setting<double> update_interval("update_interval", 0.0,
 										std::numeric_limits<double>::infinity(), 3.0, true);
 conky::range_config_setting<double> update_interval_on_battery("update_interval_on_battery", 0.0,
 										std::numeric_limits<double>::infinity(), NOBATTERY, true);
+conky::simple_config_setting<std::string> detect_battery("detect_battery",
+										std::string("BAT0"), false);
 static bool on_battery = false;
 
 double active_update_interval()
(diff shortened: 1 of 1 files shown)


djpohly:
Hm, looks like it doesn't fully fix #190 since the detected state doesn't change when I plug/unplug the adapter. :(

djpohly:
(Removed debug line that was accidentally left in.)

txpaper:
I suggest to make `detect_battery` array, so it is possible to specify more than one battery to check.

On my Thinkpad the `update_interval_on_battery` triggers only when BAT0 (internal battery) starts to discharge, while does not get triggered by BAT1 discharging.

I successfully fixed it by hardcoding an "OR" between `get_battery_short_status` of BAT0 and BAT1 but making an array of `detect_battery` it's the right solution

I would like to help by making a pull request, but I'm not a developer and I can not even understand how the parsing of the config works.

PR #457, opened 2018-01-31
Revert back drawing text shades fix (#371)

work40:
This patch affect diskio read (write) graphs too. And possible on rest graphs... This is absolutely not good.


brndnmtthws:
What exactly is the problem? I don't think this is necessarily the right fix.

brndnmtthws:
Can you try #458? I think that might be what you want.

PR #462, opened 2018-02-06
Support multiple include paths for libbircclient

billie80:
Gentoo uses /usr/include/libircclient instead of the default /usr/include because some "important" packages depend on the includes being in this directory.

The attached patch tries to support both scenarios supporting  /usr/include/libircclient and /usr/include.

I am not very familiar with cmake however my tests seem to work in both cases.


billie80:
Thanks a lot!

Syderitic:
Thank you for solving this upstream.

PR #465, opened 2018-03-05
Add support for libical 3.x

AdamMajer:
In libical3, deprecated icaltime_from_timet is removed
Use icaltime_from_timet_with_zone instead. This function
exists in libical 2.x so no need to change requirements


brndnmtthws:
Seems reasonable, thanks!

PR #479, opened 2018-04-15
Fixed crash caused by temporary virtual net interfaces

Barben360:
Net interfaces are managed the following way in Conky:
- A static array of size `MAX_NET_INTERFACES` is created (by default, 64),
- Net interfaces are parsed from system file (e.g. `/proc/net/dev` for general linux),
- When a new net interface is found, it is added to the array,
- If the array is full, Conky fails.

However, if an interface does not exist anymore in the system list, it is never removed from the list. With processes like Docker which creates a lot of virtual net interfaces, Conky is likely to crash quite fast.

I modified `net_stat.<h,cc>` to prevent Conky from crashing. The new behaviour is:
- For the `MAX_NET_INTERFACES` first interfaces, the behaviour is still the same,
- When a new interface is met, it is stored in a unique static `struct net_stat` which address is returned.
Therefore, when a new interface is found, the same static `struct net_stat` is cleared and rebuilt from the new interface's parameters. This way, there are never more than `MAX_NET_INTERFACES + 1` interfaces saved and no crash.

N.B: I used Eclipse auto-indent function so the files indentations have changed. I can fix that if necessary.

### DIFF
--- a/src/net_stat.cc
+++ b/src/net_stat.cc
@@ -51,24 +51,20 @@
 /* network interface stuff */
 
 enum if_up_strictness_ {
-	IFUP_UP,
-	IFUP_LINK,
-	IFUP_ADDR
+	IFUP_UP, IFUP_LINK, IFUP_ADDR
 };
 
 template<>
-conky::lua_traits<if_up_strictness_>::Map conky::lua_traits<if_up_strictness_>::map = {
-	{ "up",      IFUP_UP },
-	{ "link",    IFUP_LINK },
-	{ "address", IFUP_ADDR }
-};
+conky::lua_traits<if_up_strictness_>::Map conky::lua_traits<if_up_strictness_>::map =
+		{ { "up", IFUP_UP }, { "link", IFUP_LINK }, { "address", IFUP_ADDR } };
 
-static conky::simple_config_setting<if_up_strictness_> if_up_strictness("if_up_strictness",
-																		IFUP_UP, true);
+static conky::simple_config_setting<if_up_strictness_> if_up_strictness(
+		"if_up_strictness", IFUP_UP, true);
 /**
  * global array of structs containing network statistics for each interface
  **/
 struct net_stat netstats[MAX_NET_INTERFACES];
+struct net_stat foo_netstats;
 
 /**
  * Returns pointer to specified interface in netstats array.
@@ -78,8 +74,8 @@ struct net_stat netstats[MAX_NET_INTERFACES];
  *
  * @param[in] dev  device / interface name. Silently ignores char * == NULL
  **/
-struct net_stat *get_net_stat(const char *dev, void *free_at_crash1, void *free_at_crash2)
-{
+struct net_stat *get_net_stat(const char *dev, void *free_at_crash1,
+		void *free_at_crash2) {
 	unsigned int i;
 
 	if (!dev) {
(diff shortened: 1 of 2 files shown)


src/net_stat.cc
```diff
@@ -51,24 +51,20 @@
 /* network interface stuff */
 
 enum if_up_strictness_ {
-	IFUP_UP,
-	IFUP_LINK,
-	IFUP_ADDR
+	IFUP_UP, IFUP_LINK, IFUP_ADDR
```
brndnmtthws: Let's keep each enum value on a separate line here.

src/net_stat.cc
```diff
@@ -51,24 +51,20 @@
...
 };
 
 template<>
-conky::lua_traits<if_up_strictness_>::Map conky::lua_traits<if_up_strictness_>::map = {
-	{ "up",      IFUP_UP },
-	{ "link",    IFUP_LINK },
-	{ "address", IFUP_ADDR }
-};
+conky::lua_traits<if_up_strictness_>::Map conky::lua_traits<if_up_strictness_>::map =
+		{ { "up", IFUP_UP }, { "link", IFUP_LINK }, { "address", IFUP_ADDR } };
```
brndnmtthws: Same here, separate lines.

src/net_stat.cc
```diff
@@ -93,47 +89,57 @@ struct net_stat *get_net_stat(const char *dev, void *free_at_crash1, void *free_
...
+					shownetmask = true;
+				if (nextarg[j] == 's')
+					showscope = true;
 			}
-		}
-		else netstat = get_net_stat(nextarg, obj, free_at_crash);
-		i+=strlen(nextarg);	//skip this arg
-		while( ! (isspace(arg[i]) || arg[i] == 0)) i++; //and skip the spaces in front of it
+		} else
+			netstat = get_net_stat(nextarg, obj, free_at_crash);
```
brndnmtthws: Please wrap with braces.

src/net_stat.cc
```diff
@@ -93,47 +89,57 @@ struct net_stat *get_net_stat(const char *dev, void *free_at_crash1, void *free_
...
+		i += strlen(nextarg);	//skip this arg
+		while (!(isspace(arg[i]) || arg[i] == 0))
+			i++; //and skip the spaces in front of it
 	}
-	if(netstat == NULL) netstat = get_net_stat(DEFAULTNETDEV, obj, free_at_crash);
+	if (netstat == NULL)
```
brndnmtthws: Please use braces.

src/net_stat.cc
```diff
@@ -93,47 +89,57 @@ struct net_stat *get_net_stat(const char *dev, void *free_at_crash1, void *free_
...
+					showscope = true;
 			}
-		}
-		else netstat = get_net_stat(nextarg, obj, free_at_crash);
-		i+=strlen(nextarg);	//skip this arg
-		while( ! (isspace(arg[i]) || arg[i] == 0)) i++; //and skip the spaces in front of it
+		} else
+			netstat = get_net_stat(nextarg, obj, free_at_crash);
+		i += strlen(nextarg);	//skip this arg
+		while (!(isspace(arg[i]) || arg[i] == 0))
```
brndnmtthws: Please use braces.

src/net_stat.cc
```diff
@@ -361,7 +355,7 @@ void print_wireless_mode(struct text_object *obj, char *p, int p_max_size)
 	struct net_stat *ns = (struct net_stat *)obj->data.opaque;
 
 	if (!ns)
-		return;
+	return;
```
brndnmtthws: Looks like a bunch of formatting changes. Can you either leave it as is, or convert these to:

```c
if (!ns) {
    return;
}
```

Barben360: Yes as I stated I used Eclipse auto-indent on the whole file by habit. I'll just take the original files, cautiously re-add my modifications and do a new commit.

Barben360:
I made the requested changes. Only parts of the file with code modifications were modified compared to the original.

PR #480, opened 2018-04-15
Basic macOS support

npyl:
Hi everyone,
This is my work for getting conky to run on macOS and load the entire default config.

## Preview

conky running on my mac

At the moment everything that is needed for the default conky config to be shown, is supported.  (There are other things working too: e.g. `if_mounted`, `if_existing` etc)

## CHANGES
- In `ConkyPlatformChecks.cmake`, `Conky.cmake` and `src/CMakeLists.txt` introduce support for Darwin
- Introduce the files `darwin.cc`, `darwin.h` and `darwin_sip.h`
- In `ConkyPlatformChecks.cmake` force cmake to search for `statfs64` in `sys/mount.h` instead of `sys/statfs.h` as it doesn't exist in the latter.
- In `.gitignore` add `.DS_Store` which are spotlight-generated files that offer directory-related attributes; unnecessary for conky development on macOS.
- In `luamm.cc` I encountered the error: `No member named '__cxa_exception_type' in ‘std::exception_ptr’` and couldn’t find a solution so I disabled the code for macOS
- In `semaphore.hh` I had to introduce a variation of `semaphore` class because macOS does not support named-unix semaphores (although available) . Also, I didn't want to bloat the code with too much preprocessor-code
- In `core.cc` enable the conky-variables `running_processes`, `threads `, `running_threads` for macOS
- Introduce `previous_total_cpu_time` as member variable in  
- In `top.h` I introduced the member variable `previous_total_cpu_time` in an attempt to optimise to the finest `get_top_info()`and specifically for getting %CPU usage foreach process.  I have implemented a slightly different approach than the one used in linux.cc:
My approach is based on the concept that ***total-CPU-time-delta*** should be *process-specific* instead of *iteration-specific* as implemented in `linux.cc`.
For the sake of keeping this PR compacked I will add a link to the issue in conky-for-macOS repo: [link](https://github.com/Conky-for-macOS/conky-for-macOS/issues/15)
- Introduce a new conky-variable: `sip_status`.  This variable is Only available for macOS and offers information about the security measure ***'System Integrity Protection'*** or ***SIP***
- In `core.cc` introduce `sip_status`
- In `extras/vim/syntax/conkyrc.vim`, `extras/nano/conky.nanorc` and `doc/variables.xml` add `sip_status` variable and provide documentation for it in the latter
- Introduce `HAVE_CLOCK_GETTIME` cmakedefine to provide backwards compatibility with versions of macOS that don't provide `clock_gettime()`

## TESTED WITH
- **Apple's Instruments** - For finding memory leaks
- **Apple’s** `top` - For memory stats
- **Activity Monitor** - For CPU usage

## MY QUESTIONS
1. I noticed in almost all source-file's License headers a line saying the identation used.  Though, for development I was using XCode and the identation is different in `darwin.cc`, `darwin.h` and `darwin_sip.h`.  Is this a necessary requirement for merging? 
2. I have included code from apple’s top utility which is APSL-licensed in order to implement memory-stats.  The problem is, that the algorithm apple uses, is the only CORRECT, and is fairly **undocumented**.  **AFAIK**, there is no other way to calculate memory statistics than using apple-code.  Does this cause a problem, keeping in mind that I am using small portions of their software and copying the only available way of finding memory stats?

I apologise for giving you a pull request 130-commits-long but lots of work had to be done to bring conky to a usable state first. :)


brndnmtthws:
Wow, this is really cool. It's going to take some time to review this, but I'll get back to you on it.

npyl:
Thats awesome thanks. If you need to build this i have wriiten a simple guide here: [link](https://github.com/Conky-for-macOS/conky-for-macOS/wiki/How-to-Build)

Apart from installing the required libraries it is pretty much normal conky compilation!

brndnmtthws:
@npyl can you address the build error?

At this point I think I'll just merge this, and go in and clean stuff up later. I'd rather get this merged and get a release out so we can find bugs.

I want to add some tools like clang-format into the build system so we can standardize on formatting and include some static analysis.

npyl:
@brndnmtthws Yep, i will check it out

npyl:
Hey btw, I should also tell you that I am really stressed with school right now, so after i finish it (approx 2 months) i will be able to dedicate alot time.  Until then I am not sure how much i will be able to dedicate thanks for the understanding

brndnmtthws:
@npyl not to worry. Conky has been around for more than 12 years, and I think it'll still be around in another 12 :)

npyl:
Hi!! I found some time and solved the problem.  Thank you for your patience! :)

brndnmtthws:
@npyl as a follow-up to this, could you take a stab at adding a macOS Travis CI build? That would be super valuable!

npyl:
@brndnmtthws Yeah! I will do it as soon as I can! thats cool!

brndnmtthws:
@npyl also take a look at #486. I'm trying to get the build process into better shape.

PR #507, opened 2018-05-15
Add cat and catp variables.

evmcl:
Added the following variables:

  cat file

```
Reads a file and displays the contents in conky. This is useful
if you have an independent process generating output that you
want to include in conky.
```

  catp file

```
Reads a file and displays the contents in conky. This is useful
if you have an independent process generating output that you
want to include in conky. This differs from $cat in that it
parses the contents of the file, so you can insert things like
${color red}hi!${color} in your file and have it correctly parsed
by Conky.
```


brndnmtthws:
For some reason the sonar config, which is here:

https://github.com/brndnmtthws/conky/blob/master/.travis.yml#L82-L85

is missing from your build:

https://travis-ci.org/brndnmtthws/conky/jobs/379157167/config

I'm not exactly sure why it's being dropped. Let me poke around.

brndnmtthws:
Seems like it's not even possible yet: https://stackoverflow.com/questions/45612758/how-do-i-get-sonarcloud-to-run-on-pull-requests-from-forks-with-travis-maven

I'll have to figure out something else.

PR #530, opened 2018-07-02
Documentation fixes for macOS

npyl:
This commit fixes documentation of a macOS specific variable (`sip_status`) as promised in #480.

Also, fix some identation problems in `darwin.cc`

**EDIT:** the second commit is quite interesting, though it would seem be wiser that @brndnmtthws  waits before merging for the respective [issue](https://github.com/Conky-for-macOS/conky-for-macOS/issues/18) to be fixed.


brndnmtthws:
I'll merge #532 instead of this (it's the same).

PR #534, opened 2018-07-08
Added destroy methods to some functions to fix memory leaks.

Monsterovich:
Accidentally I noticed that my conky .lua file produces a lot of memory leaks. I used Google (and this too: http://u-scripts.blogspot.com/2011/01/memory-leaks-in-conkyluacairo.html) and fixed everything I could. It doesn't work.

By increasing the refresh rate conky ate over 700 mb in a few seconds.
qtox_image_2018-07-08 23-04-37 478

Then I just did this for testing:
```
for i = 1, 1000 do
    local te = cairo_text_extents_t:create()
    local fe = cairo_font_extents_t:create()
end
```
It seems that memory created by cairo_*_extents_t:create() is never freed.

I added functions to free unused memory.
```
for i = 1, 1000 do
    local te = cairo_text_extents_t:create()
    local fe = cairo_font_extents_t:create()
    cairo_text_extents_t:destroy(te)
    cairo_font_extents_t:destroy(fe)
end
```

Now, it works fine even on very high refresh rate.

### DIFF
--- a/lua/cairo.pkg
+++ b/lua/cairo.pkg
@@ -61,6 +61,7 @@ typedef struct _cairo_matrix {
 	double x0;
 	double y0;
 	static tolua_outside cairo_matrix_t* create_cairo_matrix_t @ create();
+	static tolua_outside void destroy_cairo_matrix_t @ destroy(cairo_matrix_t *pointer);
 } cairo_matrix_t;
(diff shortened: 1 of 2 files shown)


brndnmtthws:
Merged in #535. Thanks!

Issue #303, opened 2016-08-02
No cast snafu


ghost:
Hello,

Why do you use the ctype header functions with arguments of **signed** int ?

https://www.securecoding.cert.org/confluence/display/c/STR37-C.+Arguments+to+character-handling+functions+must+be+representable+as+an+unsigned+char

su8:
Thank you for reporting, it's fixed in the following commit - https://github.com/brndnmtthws/conky/commit/7bca0ca8819f03883affa5fba9e8506d8c2c5f29

PR #544, opened 2018-08-01
core.cc,x11.cc,x11.h: Add Num,Caps,Scroll lock text objects, and fix #439

### LINKED ISSUE #439: Add XKB Indicators
Some laptops and keyboards don't come with LED Caps Lock or Num Lock indicators.

It would be nice if conky could display the state of XKB indicators such as Caps Lock, Scroll Lock, Num lock etc.

su8:
Three new text objects are added: `num_led, scroll_led, caps_led` to solve https://github.com/brndnmtthws/conky/issues/439

edit:
It also solves issue https://github.com/brndnmtthws/conky/issues/303

### DIFF
--- a/src/algebra.cc
+++ b/src/algebra.cc
@@ -147,7 +147,7 @@ enum arg_type get_arg_type(const char *arg) {
     p++;
   }
   while (p <= e) {
-    if (isdigit(*p) == 0) {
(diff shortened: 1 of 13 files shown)


su8:
Rebased 3 times and my commits are still not squashed :{

Issue #306, opened 2016-08-11
tcp_ping returns raw data (nanoseconds)


BondarenkoArtur:
When i use `${tcp_ping google.com}` i get something like 18120 or 18928 when my ping is 18 ms
Probably is supposed to be shorten

su8:
Thanks for reporting this one, it's fixed in the latest commit.

PR #545, opened 2018-08-01
Fix issues #439, #303

su8:
Three new text objects are added: `num_led, scroll_led, caps_led` to solve https://github.com/brndnmtthws/conky/issues/439

It also solves issue https://github.com/brndnmtthws/conky/issues/303

edit:

It also solves issue #306

### DIFF
--- a/src/algebra.cc
+++ b/src/algebra.cc
@@ -147,7 +147,7 @@ enum arg_type get_arg_type(const char *arg) {
     p++;
   }
   while (p <= e) {
-    if (isdigit(*p) == 0) {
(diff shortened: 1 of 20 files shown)


su8:
The test run also "build-no-x11", so they fail because x11.cc is not compiled while the functions in core.cc rely on x11.cc to be compiled.

su8:
@brndnmtthws Are you interested in getting the currently used keyboard layout ?

brndnmtthws:
@su8 sure that sounds good. Let's get this merged first, then open a separate PR for that.

PR #563, opened 2018-08-05
Fixes compile issue with builtin config.

Vallimar:
Quick fix that resolves a compile error.


Vallimar:
This is for issue #547

PR #569, opened 2018-08-06
journal.cc: Add more overflow checks

su8:


brndnmtthws:
I merged #570 instead of this PR (it's the same).

PR #575, opened 2018-08-07
Fix problems found when fixing docker

ngaro:
These are all (mostly small) fixes and enhancements. I found them missing while building a container.

Only the 4 oldest have something to do with docker and only the 2nd oldest is a docker-only fix.


brndnmtthws:
It might be easier to do separate PRs next time  :)

I'm still trying to figure out how to use the code quality checks effectively. The good thing about having them is that we can have more autonomous collaborators, the downside is that it creates more friction.

Issue #26, opened 2013-10-19
Can't set more than one font with transparency


MichaelZer0:
It would be great if we could use transparency for each font independently, something like this:

${font Bitstream Vera Sans Mono:size=40:xftalpha=0.3}${time %M}
${font Droid Sans Mono:size=10:xftalpha=0.5}${time %S}

mswanson-me:
Is this issue still outstanding in the current version of Conky? If so, please update this issue. Otherwise, the issue will be closed in 2 weeks.

PR #577, opened 2018-08-07
More improvements for macOS and conky generally

npyl:
Hi!
This is some work on both general stuff and macOS specific stuff. 😝

# General Changes
- Activated the option `BUILD_WLAN` for all OSs (set to `false` by default)
- Fixed `$wireless_essid` (also affected Linux, probably other OS, too!) crashing conky when no `arg` was passed.  Correct behaviour would be to allocate `dev` with default interface (e.g. `eth0`). Adding `parse_net_stat_arg` fixes the problem.
- Removed interface name from `$downspeed` and `$upspeed` from  `conky.conf` and `conky_no_x11.conf`; we want conky to always use the `DEFAULTNETDEV` which is set during configuration with cmake and my vary from OS to OS.  This is part of 147499f320850210aeb69954fed2e690aeace252.

# macOS Changes
- Fixed `$if_up` by removing the `SOCK_CLOEXEC` flag from `socket()` in `net_stats.cc` when building for macOS; for some reason the function fails with error 43: `Protocol not supported`.  This may also affect linux but I haven't checked yet.
- Improve support for macOS (Patches for [#29](https://github.com/Conky-for-macOS/conky-for-macOS/issues/29) and [#26](https://github.com/Conky-for-macOS/conky-for-macOS/issues/26) are applied automatically now as of 147499f320850210aeb69954fed2e690aeace252 during cmake configuration)
- Improve `get_freq()` using the [Intel Power Gadget API](https://software.intel.com/en-us/blogs/2012/12/13/using-the-intel-power-gadget-api-on-mac-os-x) if `BUILD_IPGFREQ` is enabled.
- Add build option `BUILD_IPGFREQ` (set to `false` by default) so that the user can choose whether to use IPG or stick with old behaviour (Print **constant factory** frequency of CPU).

# IPG (Intel Power Gadget concerns)

**Intel Power Gadget** can be installed using **Homebrew-cask** and it installs a .app, a Framework and a kernel-extension.  This will seem a bit troublesome in my opinion for many users thus the `BUILD_IPGFREQ` option.
Now, one has the option to either use a better `$freq` or avoid the hussle of installing a library/kernel extension (with sacrifices).

***NOTE:*** IPG doesn't seem to support multiple cores although stating the opposite in the documentation thus, all readings are for core0.

# More Concerns

I am not sure if using Objective-C++ in `darwin.mm` will be a problem for sonar but if there is a problem I will work it out.

# Epilogue

This is a prequel; actual support for WLAN on macOS is coming soon!


brndnmtthws:
Let me make you a collaborator, and you can push your branch directly to this repo.

npyl:
There must be some problem with sonar, probably because of I changed `darwin.cc` to `darwin.mm` for Objective-C++

brndnmtthws:
@npyl Sonar doesn't work with external pull requests. The PR must originate from a branch on this repo, so it won't pass with this PR.

brndnmtthws:
@npyl Have a look at #579, the sonar checks should run on that PR.

npyl:
I see!! Thats great!  But still, why hasn't it passed the check yet?  Sonar has been going from the moment I opened the PR

brndnmtthws:
@npyl it can take 5-10 minutes.

PR #579, opened 2018-08-07
Conky for mac os master

brndnmtthws:


src/core.cc
```diff
@@ -121,6 +121,12 @@
 #include "darwin.h"
 #endif
 
+#if defined (__APPLE__) && defined (__MACH__)
+#define DARWIN_DEALLOCATE_CPU_SAMPLE obj->callbacks.free = &deallocate_cpu_sample;
```
brndnmtthws: This is a nit pick, but can you change this to a function instead of a macro? We have too many crazy C macros.

npyl: Can you please give me an example because I am not sure how this would work?  If I understand correct, you want me to introduce a general function e.g. `deallocate_cpu` but this would mean that all OS-specific files (linux.cc, freebsd.cc, etc) should introduce this function, too!

brndnmtthws: Yes precisely, an OS-dependent function (like `deallocate_cpu`) which can be set as the `free` callback, like this:

https://github.com/brndnmtthws/conky/blob/4d5db325d99ac5081f0e770c055562093654a6bd/src/core.cc#L579-L582

brndnmtthws: This function can just be a no-op on other platforms btw.

This is mostly about establishing better practices/patterns. I try to make sure everything continually improves :)

npyl: I think I can make it so that we won't have to make this change either and I will get rid of the macros!

npyl: Ok! I am on it now!

npyl: @brndnmtthws I got it! 😉

src/darwin.mm
```diff
@@ -286,7 +286,7 @@ void allocate_cpu_sample(struct cpusample **sample) {
   sample_handle = *sample; /* use a public handle for deallocating */
 }
 
-void deallocate_cpu_sample(struct text_object *obj) {
+void free_cpu(struct text_object *) {
```
brndnmtthws: Brilliant, thanks!

src/linux.cc
```diff
@@ -936,6 +936,9 @@ int update_cpu_usage(void) {
   return 0;
 }
 
+void free_cpu(struct text_object *) {
```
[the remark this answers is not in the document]
npyl: @brndnmtthws Would writing something like `void free_cpu(struct text_object *) {}` bypass the error?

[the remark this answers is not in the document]
brndnmtthws: It just needs an empty comment, I went ahead and added it.

brndnmtthws:
Sorry, I made a mess of the PR. I think we are good to go now though.

PR #642, opened 2018-08-29
conky.cc: quit on SIGHUP

lasers:
Addresses https://github.com/brndnmtthws/conky/issues/436.


lasers:
Tested OK with with `out_to_console`.

PR #646, opened 2018-09-01
exec.cc: Cache the result to another variable and return it when we have a empty/invalid value

su8:
Fix for https://github.com/brndnmtthws/conky/issues/580

shoutout to @lasers


j4nu5:
IMHO this causes a bug : https://github.com/brndnmtthws/conky/issues/580#issuecomment-468965537

Issue #56, opened 2014-07-15
Any plan for wayland?


zhw2101024:
Is it possible to add support for wayland?

AquaL1te:
Conky is the last one that needs to add support for Wayland on my desktop. The other one was Redshift, but GNOME created an alternative themselves with Wayland support. Please add Wayland support :)

q2dg:
Resolving this issue is crucial for Conky's future

mmuman:
Hi, I started a Conky port to Haiku #433 ... while we don't use Wayland, we don't use X11 either. So we also need the X11 code to be factored out to add proper Haiku GUI support.

This year Haiku is participating into the [Google Code-In](https://developers.google.com/open-source/gci/how-it-works) contest that is just starting, and Ubuntu is participating as well, and since they are now moving to Wayland, they might also be interested. If any dev is willing to help mentor the students, we could propose GCI tasks to clean up the code and add proper Wayland and Haiku support.

It would be the occasion to introduce some GUI platform base class and have the different implementations override it.

CottonEaster:
I too would very much like Wayland support in Conky.

carlosgonzz:
Any info on wayland support?

npyl:
Please checkout my issue in conky-for-macOS project: https://github.com/Conky-for-macOS/conky-for-macOS/issues/35 .  A similar problem has arisen for the macOS port:  there is need to ditch XLib and switch to modern APIs such as Metal.

mmuman:
So yeah, we should start by cleaning up and adding some platform base classes…

npyl:
@mmuman An idea is porting conky to wayland and then creating a translation layer between Wayland and Metal.  This way both macOS and Linux get supported **BUT** other OSes get left out, eg. BSDs.  I think Wayland doesn't run on BSD.

mmuman:
Using a toolkit would be an option but again, the usage pattern in conky is quite specific with desktop placement, and I don't think it's properly implemented in most toolkits.

mmuman:
I mean, putting an app's window on the desktop is a corner case that is probably not supported in most toolkits because it is usually very system-specific, and not used enough to require people to properly implement it.
At least I don't recall seeing a single toolkit mentioning it.

mmuman:
Ok, I had a try at cleaning up the mess. I already managed to decouple the HTTP stuff.

mmuman:
[...]

PR #664, opened 2018-10-07
GUI backends phase 1

mmuman:
This is my first attempt at trying to sort out the X11-specific code from the rest, and allow implementing other display backends, to help with Haiku port, Wayland (#56), or Metal (OSX)…

It's mostly working but not really finished. The last commit kept the original code around with changed ifdefs for easier comparison until it's done, and I'll keep amending it for now.

I'm not used much to templating in C++ so maybe the classes look a bit naive, but I believe it's a reasonable implementation (maybe using templates would avoid RTTI and speed things up, I don't know, but I'm not confident enough to try it).
I'm not used to non CamelCase class names, but oh well.

I'm open to comments and suggestions.

Some details:
- the `BUILD_GUI` define is declared when either `BUILD_X11` or any later added GUI backend is enabled.
- Each backend has a priority (0 for text ones, 1 for ncurses, 2 for graphical ones), and they are sorted to try the more interesting ones first (hmm, maybe I should just use a sorted vector right away?).
- The display accessors added to `conky.cc` are a bit uncertain. First I wanted to just loop over each of the displays, but then part of the code actually assume graphical output, or not. So it's still a mixture of `for` loop on `current_display_outputs` and calls to `display_output()`. In `draw_stuff()` I call `[un]set_display_output()` to make sure the display we're drawing to is the graphical one. Although in theory it's the highest priority, so the first on the list, so always the "current" one. So this part can probably be simplified.
- I'm not entirely sure how to handle settings reloading, currently it breaks. I noticed the X11 window is actually created in the settings class, and not really closed between reloads. I'll have to think about this. On platforms where we would support more than one GUI displays (like, X11, SDL, Wayland…) having two enabled in the settings would cause trouble as even though only the first found would be `initialize()d`, the other one would still have the settings on and so the window created…
- In theory we could display to many outputs at once, but I believe some parts of the code just isn't designed for this. For multiple text outputs it's ok, not for others. For example X11+ncurses didn't work already. Is it desirable? I don't think it's worth it.
- the interface mimics the few X11 calls that are used in the drawing code, I believe they shouldn't be too hard to implement on other platforms (except maybe the stipple patterns on Haiku but well).
- I believe most of the X11 settings are perfectly valid for other GUIs, except maybe -w.
- I added messages about backends disabled at compile-time, but they are probably useless (and X11 on Haiku wouldn't make much sense anyway). Probably I should use the logging interface instead of cerr?
- the lua scripting interface is concerning: it exposes X11 stuff for no particular reason, I believe it should rather provide a GUI-agnostic `cairo_surface_create()` call. Else it means having to fix all scripts for other platforms. We'll likely have to provide fake calls on other platforms to mimic the X11 API for compatibility anyway.
- ~~I think most backend methods should just return void actually, the bool is useless.~~ DONE
- I noticed I used the American "initialize" instead of British, easy to change.
- the `fonts.clear()` call when not using X11 is a bit weird, and not easy to do properly from a backend that's not been initialised. Why initialize fonts if we won't use them…

### DIFF
--- a/src/display-console.cc
+++ b/src/display-console.cc
@@ -24,6 +24,7 @@
 
 #include "conky.h"
 #include "display-console.hh"
+#include "nc.h"
 
 #include <iostream>
 #include <sstream>
@@ -45,7 +46,8 @@ display_output_console::display_output_console(const std::string &name_)
 }
 
 bool display_output_console::detect() {
-  if (out_to_stdout.get(*state) || out_to_stderr.get(*state)) {
+  if ((out_to_stdout.get(*state) || out_to_stderr.get(*state)) &&
+      !out_to_ncurses.get(*state)) {
     std::cerr << "Display output '" << name << "' enabled in config."
               << std::endl;
     return true;
--- a/src/display-console.cc
+++ b/src/display-console.cc
@@ -1,0 +1,60 @@
+/*
+ *
+ * Conky, a system monitor, based on torsmo
+ *
+ * Please see COPYING for details
+ *
+ * Copyright (C) 2010 Pavel Labath et al.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+#include <config.h>
+
+#include "conky.h"
+#include "display-console.hh"
+
+#include <iostream>
+#include <sstream>
+#include <unordered_map>
+
+namespace conky {
+namespace {
+
+conky::display_output_console console_output("console");
+
+}  // namespace
+
+namespace priv {}  // namespace priv
+
+display_output_console::display_output_console(const std::string &name_)
+    : display_output_base(name_) {
+  // lowest priority, it's a fallback
+  priority = 0;
+}
+
+bool display_output_console::detect() {
+  if (out_to_stdout.get(*state) || out_to_stderr.get(*state)) {
+    std::cerr << "Display output '" << name << "' enabled in config."
+              << std::endl;
+    return true;
+  }
+  return false;
+}
+
+bool display_output_console::initialize() { return true; }
+
+bool display_output_console::shutdown() { return true; }
+
+}  // namespace conky
--- a/src/display-console.cc
+++ b/src/display-console.cc
@@ -34,6 +34,9 @@
 #include <sstream>
 #include <unordered_map>
 
+static conky::simple_config_setting<bool> extra_newline("extra_newline", false,
+                                                        false);
+
 namespace conky {
 namespace {
 
@@ -62,5 +65,17 @@ bool display_output_console::detect() {
 bool display_output_console::initialize() { return true; }
 
 bool display_output_console::shutdown() { return true; }
+
+void display_output_console::draw_string(const char *s, int w) {
+  if (out_to_stdout.get(*state)) {
+    printf("%s\n", s);
+    if (extra_newline.get(*state)) { fputc('\n', stdout); }
+    fflush(stdout); /* output immediately, don't buffer */
+  }
+  if (out_to_stderr.get(*state)) {
+    fprintf(stderr, "%s\n", s);
+    fflush(stderr); /* output immediately, don't buffer */
+  }
+}
 
 }  // namespace conky
--- a/src/display-console.cc
+++ b/src/display-console.cc
@@ -4,7 +4,11 @@
  *
  * Please see COPYING for details
  *
- * Copyright (C) 2010 Pavel Labath et al.
+ * Copyright (C) 2018 François Revol et al.
+ * Copyright (c) 2004, Hannu Saransaari and Lauri Hakkarainen
+ * Copyright (c) 2005-2019 Brenden Matthews, Philip Kovacs, et. al.
+ *	(see AUTHORS)
+ * All rights reserved.
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
--- a/src/display-console.cc
+++ b/src/display-console.cc
@@ -65,7 +65,7 @@ bool display_output_console::initialize() { return true; }
 
 bool display_output_console::shutdown() { return true; }
 
-void display_output_console::draw_string(const char *s, int w) {
+void display_output_console::draw_string(const char *s, int) {
   if (out_to_stdout.get(*state)) {
     printf("%s\n", s);
     if (extra_newline.get(*state)) { fputc('\n', stdout); }
--- a/src/display-console.cc
+++ b/src/display-console.cc
@@ -43,6 +43,7 @@ namespace {
 conky::display_output_console console_output("console");
 
 }  // namespace
+void init_console_output() {}
 
 namespace priv {}  // namespace priv
 
--- a/src/display-console.cc
+++ b/src/display-console.cc
@@ -55,8 +55,7 @@ display_output_console::display_output_console(const std::string &name_)
 bool display_output_console::detect() {
   if ((out_to_stdout.get(*state) || out_to_stderr.get(*state)) &&
       !out_to_ncurses.get(*state)) {
-    std::cerr << "Display output '" << name << "' enabled in config."
-              << std::endl;
+    DBGP2("Display output '%s' enabled in config.", name.c_str());
     return true;
   }
   return false;
--- a/src/display-console.cc
+++ b/src/display-console.cc
@@ -6,7 +6,7 @@
  *
  * Copyright (C) 2018 François Revol et al.
  * Copyright (c) 2004, Hannu Saransaari and Lauri Hakkarainen
- * Copyright (c) 2005-2019 Brenden Matthews, Philip Kovacs, et. al.
+ * Copyright (c) 2005-2021 Brenden Matthews, Philip Kovacs, et. al.
  *	(see AUTHORS)
  * All rights reserved.
  *
--- a/TODO.md
+++ b/TODO.md
@@ -1,0 +1,14 @@
+- [X] fix crash when no output is enabled
+- [X] display: fix logging (stderr)
+- [X] display: test build without X11, ncurses
+- [ ] conky.cc: case GOTO: cur_x always = current->arg regardless !? (draw_mode == BG = dead code!!) -> per display ++ ? (but exists already in master if CONFIG_X11 && CONFIG_NCURSES)
+- [X] check for regression (changes in master missed since rebase)
+- [ ] move BUILD_IMLIB2 stuff from conky.cc to display-x11 ??
+- [ ] display: add required ignored -W pragmas for clang & GCC in new files (display-x11.*)
+- [X] move font handling to X11 code
+- [ ] drop x11.h include from font.h
+- [ ] display: profile code
+- [ ] display: remove extra virtual on final classes
+- [ ] display: use int*_t types from cstdint instead of plain int.
+- [ ] display: use std::string in draw_string() and friends
+- [ ] display: use std::pair for x,y arguments
--- a/cmake/ConkyBuildOptions.cmake
+++ b/cmake/ConkyBuildOptions.cmake
@@ -207,6 +207,11 @@ else(BUILD_X11)
   set(BUILD_NVIDIA false)
 endif(BUILD_X11)
 
+# if we build with any GUI support
+if(BUILD_X11)
+  set(BUILD_GUI true)
+endif(BUILD_X11)
+
 if(OWN_WINDOW)
   option(BUILD_ARGB "Build ARGB (real transparency) support" true)
 else(OWN_WINDOW)
--- a/cmake/config.h.in
+++ b/cmake/config.h.in
@@ -113,6 +113,8 @@
 
 #cmakedefine BUILD_HTTP 1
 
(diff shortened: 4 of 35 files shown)


mmuman:
Hmm, the font stuff also needs work it seems.

brndnmtthws:
Wow, this is an epic PR. Thank you for the work. It sounds like this is still a work in progress, so I will let you continue, but please let me know when you think it's time for me to take a closer look.

src/display-console.cc
```diff
@@ -0,0 +1,68 @@
+/*
+ *
+ * Conky, a system monitor, based on torsmo
+ *
+ * Please see COPYING for details
+ *
+ * Copyright (C) 2010 Pavel Labath et al.
```
brndnmtthws: Please change this to `Copyright (C) 2018 François Revol et al.` here and elsewhere.

mmuman: Ok I'll include it in my next round.

mmuman:
Second round: Not much more, mostly cleanup of history and code.
Fixed copyright on new files.

lasers:
@mmuman Can you fix https://www.codefactor.io/repository/github/brndnmtthws/conky/pull/672?

npyl:
@mmuman For macOS if you include `<vector>` in `display-output.cc ` the build may pass (It was required when i tried to compile your first round of patches)
Btw, cheers for this amazing PR, it will really help conky. 🍺

mmuman:
> @mmuman For macOS if you include `<vector>` in `display-output.cc ` the build may pass (It was required when i tried to compile your first round of patches)

Ok.

> @mmuman Can you fix https://www.codefactor.io/repository/github/brndnmtthws/conky/pull/672?

Thanks, I'll have a look.

mmuman:
Another round: autosquashed all the fixup commits. Also moved the accessors around as I'll probably need them in font code.

mmuman:
I was quite busy with work so I still didn't touch the font stuff, I'll try this week.

mmuman:
Ok, I managed to fix the only thing left that was holding the cleanup up (crash on config reload).
I think I'll deal with the font code separately, the changes are already large enough.
Please comment.

mmuman:
and we need #include <algorithm> for sort()…

mmuman:
`build-wrapper-linux-x86-64: command not found`
Hmm, don't think it's related, is Travis setup just bork?

lasers:
@mmuman It broke too many times in that same spot. I don't know why. I'm not even sure about this solution, but change commit hash and try again. Travis sometimes works okay on different runs.

mmuman:
let's see…

mmuman:
Oh well, 😴

mmuman:
Ah yes, some `//if (out_to_x.get(*state)) {` to remove still.

mmuman:
So, anyone tried yet?
I didn't try to benchmark it, but I suppose the cost of going through the vtables is probably equivalent to all the removed `if (out_to_x…)`.

lasers:
I tried it now. Worked okay out of box. :+1: The rest is up to @brndnmtthws to review your code.

npyl:
I will test it on macOS tomorrow (I don't have internet)

npyl:
@mmuman  BTW, i think that BUILD_GUI should be the first option to be enabled and not BUILD_X11.
BUILD_X11, BUILD_METAL, ... should be available only if BUILD_GUI is.

I am not sure if this is possible.

mmuman:
> (the window is slightly smaller though! Don't really know if this is a problem)

Odd. It's probably some additions to cur_y in the older code that isn't done in the new code-path. I'll look into this.

npyl:
@mmuman I took a look into the code this morning and I have managed to separate in a better way `BUILD_GUI` and `BUILD_X11`, guess you didn't have time to do it yourself :) but since I did it I am thinking of pushing it in a separate PR to see what happens! :)

(NOTE: It will break the build on macOS because of the X11 stuff in the fonts.h and some should-be-public stuff defined only in `x11.h`... Linux should be fine... I guess... Can't really test it to be honest. )

mmuman:
> > (the window is slightly smaller though! Don't really know if this is a problem)
> 
> Odd. It's probably some additions to cur_y in the older code that isn't done in the new code-path. I'll look into this.

Actually, I can't reproduce it here… screenshot shows official git HEAD, my branch, and Debian Sid's package, and they are perfectly aligned.
capture d ecran de 2018-11-22 01-48-20

Did you compile both with the same options?

npyl:
@mmuman I compiled it with same options! Don't really know what is going wrong... but probably it is some X11 stuff... which we are not going to care about on the mac anyway so we could probably leave this as it is right now and bother later :)

mmuman:
Damn I should really get this mergeable someday, rebasing is always painful when things change under your feet…

mmuman:
sigh… indentation fixes …

mmuman:
must… fix… almost… every… commit :-(

mmuman:
Commit log for 4d098b053 should have included "screw up any future merge / rebase" 😁

mmuman:
Managed to rebase, builds with X11 but it's probably full of bugs since I don't know what was actually fixed since last time.

mmuman:
Fixed no-X11 build.

Caellian:
I skimmed through your code and regarding template usage, I don't see any benefit they would add. They would make your current code harder to read and in this case proper OO inheritance is a way better choice than using templates.

I don't see why subclass methods for different rendering frontends are also declared virtual. If anyone is going to override them later by making a specialization class, they can easily add virtual on their own. This way it's just unneeded and confusing syntax.

Regarding indentation, clang-tidy should've done that for you. If you're not using it, I suggest you do before this PR is merged so that every time someone does an additional change after the merge, random bits of your code don't get formatted thus forcing additional line changes that could've been avoided from the start.

mmuman:
> I don't see why subclass methods for different rendering frontends are also declared virtual. If anyone is going to override them later by making a specialized class, they can easily add virtual on their own. This way it's just unneeded and confusing syntax as it implies it's already being done somewhere else in code.

That's probably some habits of coding in BeOS & Haiku where we have a lot of inheritance… Although, some frontends reuse less specialized ones already. Besides, I wasn't sure which methods I'd need to override when I started, maybe we can clean that later.

> Regarding indentation, clang-tidy should've done that for you. If you're not using it, I suggest you do (although it seems you do, not sure what you meant) before this PR is merged so that every time someone does an additional change after the merge, random bits of your code don't get formatted thus forcing additional line changes that could've been avoided from the start.

Ah, didn't know about clang-tidy, I'll give it a try.

> Also, considering conky uses C++11, consider using types defined in [<cstdint>](https://en.cppreference.com/w/cpp/types/integer) to ensure proper size allocation for integral types. Take a look at [this](https://stackoverflow.com/questions/14256695/why-does-sizeofint-vary-across-different-operating-systems).

Yeah, again I'm used to old-school C++ from BeOS :-)

> This is something very needed, good work and it's amazing you're still working on it. I asked that my commit is held off until this is accepted so that you don't have to rebase again.

It's still on my list, although the list is still very large and growing those days.

Caellian:
> Ah, didn't know about clang-tidy, I'll give it a try.

There are plugins for most IDEs and editors that apply the formatting automatically on file save. These will greatly improve your productivity in all projects with provided `.clang-tidy` files.

> It's still on my list, although the list is still very large and growing those days.

This is always a problem with big PRs. I found it more productive to do several smaller ones just to avoid problems like rebasing and being burnt out. I'd like to help if I can.

Can you provide a TODO.md file in root of your repo so that people can see what else is required for the PR and help you out? It can be removed before merge.

mmuman:
I tried to rebase the branch and use clang-format on each commit for a cleaner history, some conflicts as expected but at least once it does something weird: it wants to align an endif comment with the comment above which has nothing to do with it… oh well?
Screenshot_20201003_233429

mmuman:
> Also, considering conky uses C++11, consider using types defined in `<cstdint>` to ensure proper size allocation for integral types. Take a look at this.

Oh, hmm which types are you talking about, most of the ints passed don't really matter, it's not like we're reading an on-disk format or anything. Unless you ever want to have the API used by add-ons I'm not sure it's very important. I mean, we use these (actually our own versions) a lot in Haiku but that's to have a stable ABI. Maybe size_t would be better for string lengths?

> > Ah, didn't know about clang-tidy, I'll give it a try.
> 
> There are plugins for most IDEs and editors that apply the formatting automatically on file save. These will greatly improve your productivity in all projects with provided `.clang-format` files.

I tend to use several editors so I didn't really try to configure either.

> I said clang-tidy, I meant clang-format. clang-tidy is for linting, format actually formats your code.

Yeah, works mostly as I mentionned.

> This is always a problem with big PRs. I found it more productive to do several smaller ones just to avoid problems like rebasing and being burnt out. I'd like to help if I can.

Yes and the more time we wait the more it'll diverge and I'll have to review.

> Can you provide a TODO.md file in the root of your repo so that people can see what else is required for the PR and help you out? It can be removed before merge.

I'm more used to org-mode but yeah :-)

> Also, use `std::string` for passing strings to functions like `draw_string_at`. It can be easily turned into `char*` with `c_str()` when needed, but is more C++ish and nicer, also C++11. Using `std::u16string` might be even better because of emoji support.
> 

Hmm, I don't know u16string but is it UTF-16 or some other weirdo encoding which still has to split codepoints anyway? Not sure it'd help. Of the (recent) OSes I know only Windows doesn't use UTF-8.

The original intent was to touch the original code as less as possible to compare and make sure we didn't drop stuff, which will be useful as although I rebased several times I didn't check much what changed since then so I'll have to review again. Like, it already unmaps the window on click which master branch doesn't, so I'll have to check this.

> It's also possible to use `std::pair` for nicer passing of points on the screen. I would also allow for nicer and cleaner initialization and possibly defining some utility methods. `draw_rect` call then becomes
> I know these sound nitpicky, but using the newest possible syntax makes updating conky a lot simpler in the future. They also allow compilers to do some additional optimizations in some use cases. For instance, in C++20, a lot of functions are constexpr allowing the compiler to do compile-time calculations, making the compiled code run a lot faster.

Again, it's already a large piece, I'm not sure I want to add other changes yet, specially using constructs I'm not familiar with. I think we should rather merge it first then let you clean it up. Maybe have a ticket to list these?

Also, I'd someday like have a native Haiku GUI with replicant support which needs gcc2.95, but I guess the actual code already won't build with gcc2.95 anyway and I'll have to write some proxy app to draw from the main process anyway.

mmuman:
Of course CI doesn't like arg names in the methods… pfff

Caellian:
u16string should be UTF-16.

I agree with everything you said. I'm a bit busy with work currently, but I'm looking forward to helping out with this PR some time soon. I also sort of abandoned ricing my OS currently as there's a lot of things that need to fall into place before I feel like it's worth the effort. I'm hoping Wayland handles click events a bit better than Xorg for my commit to become actually useful on minimal WMs :smile:. Glad to see you're still working on this.

mmuman:
Ok, what could I possibly do without Internet on a christmas day… :shrug: 

I looked at all the history since I started working on this, and imported manually all the changes I missed when moving the code out of conky.cc, and splitted out the X11 font stuff. Still some things to check but it's getting there.

mmuman:
Hmm the case where no valid output is found segfaults currently (I asked for only http out but it's built without it):
```terminal
$ ./src/conky -c ../data/conky.conf 
conky: drawing to single buffer
conky: Unknown setting 'out_to_http'
Unable to find a usable display output.
conky: initialize_display_outputs() failed.
Segmentation fault
```

npyl:
Hello, I can confirm that your branch builds and runs on macOS Mojave so I guess this is something!
I also notice that the default config isn't being loaded on the desktop, so I guess X11 is not enabled on macOS?
If that's the case then, even better, because we are going to be using Metal anyway!
I can't test on linux yet.

npyl (approved):
Seems OK I think we should merge!!! Good job btw my friend.

mmuman:
Well there are still some rough edges like the crash when no output is enabled, but at least it would bother not only me so someone can fix it before I get the time to :-D

mmuman:
Rebased and fixed the crash with no enabled outputs.

mmuman:
I forgot to rerun clang-format, done.

Should I keep the TODO list around?

mmuman:
I had a try at adding an Haiku GUI backend, but `conky.cc` still pulls in quite a lot from `x11.cc`, I guess we'll look at this can of worms in a second phase.

Could we try and get this merged?

npyl:
@mmuman  Great work there. In my opinion the TODO is helpful, so I guess keep it.
I don't have the authority to merge this and @brndnmtthws seems to have been away from the project for a while now.

mmuman:
Btw, failed CI seems to be due to bad configuration (clang not being a full path?)…

npyl:
Hey @mmuman That's amazing work.  Im very happy you are still into this issue.
I am a bit occupied with real life stuff at the moment and I'm not really able to test this out, but I will certainly find some time.

As for the CI, I am not sure if as a collaborator I can do anything... Maybe @brndnmtthws  can look into it?

brndnmtthws:
> Btw, failed CI seems to be due to bad configuration (clang not being a full path?)…

I'll try and get the build sorted out ASAP. It just needs to be updated with the latest clang/distro matrix.

brndnmtthws:
@mmuman can you try rebasing now?

mmuman:
There you go :-)

brndnmtthws:
@mmuman I've added you as a collaborator, you should be able to merge this yourself now.

mmuman:
Oh cool, let's see…

mmuman:
good idea to bump the version before such a big change btw.

brndnmtthws:
> Oh cool, let's see then… Should I squash or keep the history?

I'll leave that up to you. Squashing will keep the history a bit cleaner, but if you don't squash then you'll get credited for 23 commits in one merge.

mmuman:
Since we still have things to fix I though it'd be useful to have the full history, but I'll take the `git blame` regardless :-D

npyl:
Thanks so much for this PR @mmuman  and @brndnmtthws 
I have been waiting for it since 2018 haha.

mmuman:
Getting there. Anyone experienced with profiling who'd want to compare before/after?

PR #686, opened 2018-11-29
Change variable name to avoid clash when compiling with GCC on FreeBSD

madpilot78:
When compiling with gcc the file /usr/include/machine/md_var.h is included. This header defines a function "void cpu_setup(u_int)" which clashes with the "static short cpu_setup" variable in src/freebsd.cc.

The patch renames this variable.

Reported by: Piotr Kubaj <pkubaj at anongoth.pl>
Ref: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=233616


brndnmtthws:
Thanks!

PR #688, opened 2018-12-03
Include draw_blended in convert.lua

dmccombs:
A followup to #682 , the new `draw_blended` configuration option needs to be included in `extras/convert.lua` in order for it to be converted to boolean correctly.

Without this change, if `draw_blended` is including in an old-style configuration file, on start up it will fail to use the option with the message:
`conky: Invalid value of type 'string' for setting 'draw_blended'. Expected value of type 'boolean'.`

Adding this and rebuilding conky allows it to be converted correctly on start up.

### DIFF
--- a/doc/config_settings.xml
+++ b/doc/config_settings.xml
@@ -918,6 +918,18 @@
     <varlistentry>
         <term>
             <command>
+                <option>render_draw_blended</option>
+            </command>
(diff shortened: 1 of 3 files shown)


dmccombs:
Closing this due to conflicts when updating this branch, in favor of #689.

PR #795, opened 2019-02-25
Added option to disable redraw on desktop change

brndnmtthws:
Same as #789


brndnmtthws:
> ⚠️ Missing documentation... and is `forced_redraw` a good config name? Does it make sense?

Good point, it's a bit confusing. Although I'm actually just trying to test the CI right now :)

Can you comment on the original PR? #789

PR #809, opened 2019-03-08
Use flock with FreeBSD pkg install.

brndnmtthws:
It seems that `pkg` is quite sensitive, so I'll try being more delicate. My previous attempts at avoiding race conditions were insufficient.


brndnmtthws:
@labhub-bot retry

PR #812, opened 2019-03-09
Put libraries into the proper directories and install portmon library.

billie80:
**Descriptions**

**first changeset**
* Unlike the other libraries conky_core is installed to hardcoded /usr/lib instead of /usr/lib/conky
* This also ignores 64 bit systems, using LIB_INSTALL_DIR takes care of this

**second changeset**
* when building shared libraries with SET (BUILD_SHARED_LIBS ON CACHE BOOLEAN "") which is now the default for Gentoo Linux conky fails to start with:

> conky: error while loading shared libraries: libtcp-portmon.so: cannot open shared object file: No such file or directory
* adding tcp-portmon as install target fixes this issue

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -297,10 +297,16 @@ else()
   target_link_libraries(conky ${conky_libs})
 endif()
 
-# Install libtcp-portmon too?
 install(TARGETS conky
         RUNTIME DESTINATION bin
         LIBRARY DESTINATION ${LIB_INSTALL_DIR}/conky
         ARCHIVE DESTINATION ${LIB_INSTALL_DIR}/conky)
+
+if(BUILD_PORT_MONITORS)
+  install(TARGETS tcp-portmon
+          RUNTIME DESTINATION bin
+          LIBRARY DESTINATION ${LIB_INSTALL_DIR}/conky
+          ARCHIVE DESTINATION ${LIB_INSTALL_DIR}/conky)
+endif(BUILD_PORT_MONITORS)
 
 print_target_properties(conky)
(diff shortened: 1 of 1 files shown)


brndnmtthws:
Thank you kindly.

billie80:
There seems to be an issue with my patch.

First I was testing only with ${LIB_INSTALL_DIR} instead of ${LIB_INSTALL_DIR}/conky which worked well. However when putting the libraries in ${LIB_INSTALL_DIR}/conky they wont be found again when trying to run conky.

I am sorry I did not test this change properly.

Is there anything required in addition to make the library installation work when installed in subdirectories of /usr/{lib,lib64}?

brndnmtthws:
Hmm, presumably they need to be in a path that's in the default `ld` library paths.

It wasn't my intention for these libs to actually be installed as shared libraries (it was just to make testing easier).

Anyway, after doing a little searching, I think it should just be `${LIB_INSTALL_DIR}`, without the `/conky`. Configuring the install dir seems like a distro specific thing to me. Thoughts?

billie80:
Okay I make another pull request changing it back to ${LIB_INSTALL_DIR} only.

I am just wondering how all the other packages on my system do this which are below subdirectories.
Btw. what about /usr/lib64/conky/{libcairo.so,libcairo_imlib2_helper.so,libimlib2.so,librsvg.so}?

brndnmtthws:
Good question. I'm not sure what the best way to do that is. My preference would be to simply statically link everything (I don't think anyone is actually using this code as a library).

We should probably also add a `make install` test to the CI, so we can verify it's putting things in the right places.

PR #819, opened 2019-03-27
Add a separate option for building html documentation.

billie80:
**Descriptions**
* This pull request separates building the documentation from MAINTAINER_MODE with a new option called BUILD_DOCS
* IMO it deserves and extra option as it is not related to the other functionality of MAINTAINER_MODE which sets the build type and enables tests which nowadays also enables the building of the conky_core debug library.
* This way if one only wants documentation he doesn't get a lot of other stuff he probably doesn't need.
* Building with and without the new options seems to work fine.

**Notes**
* I have not yet added this option to the `static void print_version(void)` function. If you think it is necessary I can add it .
* Maybe the travis and gitlab-ci configuration also needs to be adapted to reflect this changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -88,11 +88,11 @@ set(MAN_FILES doc/conky.1)
 
 install(FILES ${DOC_FILES} DESTINATION ${DOC_PATH})
 
-if(MAINTAINER_MODE)
+if(BUILD_DOCS)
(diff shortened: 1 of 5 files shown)


brndnmtthws:
@labhub-bot retry

brndnmtthws:
This looks good to me, thanks for the PR. I'll merge once all the build stuff finishes.

PR #825, opened 2019-04-03
Add clang 8 to build.

brndnmtthws:


brndnmtthws:
The Fedora build isn't working, and there's no separate clang 7 package (yet) in rawhide. I'll fix this later.

brndnmtthws:
For some reason the clang binaries are no longer producing coverage data (`default.profraw` is missing) after running the tests. I'm not sure why this is happening. I need to dig a bit deeper. I might just stick with clang 7 for the test coverage instead of using 8.

PR #840, opened 2019-05-20
Fix sonarcloud warnings

Tomunu:
This fixes a few SonarCloud warnings, especially the if-statement side effects.


brndnmtthws:
Sweet, thanks! Although I did end up disabling sonarcloud because it wasn't working properly.

PR #842, opened 2019-05-24
Add Tests to test-common.cc

Tomunu:
Some of these Tests might be a little trivial or overkill, but hopefully it will make it easier to add more.


Tomunu:
I see that it's the use of "eviron" in some builds. I'll work on it. 
Also, one use of round_to_int didn't get renamed.

Tomunu:
I'll open it up once I get the build problems fixed.

PR #859, opened 2019-06-28
linux: Memory statistics should match TOP tools

jbenden:
This PR modifies how memory statistics are calculated on Linux, to match the values shown in the `top` tool.

### DIFF
--- a/src/common.cc
+++ b/src/common.cc
@@ -293,6 +293,14 @@ void update_stuff() {
 
   /* if you registered a callback with conky::register_cb, this will run it */
   conky::run_all_callbacks();
+
+#if !defined(__linux__)
+  /* XXX: move the following into the update_meminfo() functions? */
+  if (no_buffers.get(*state)) {
+    info.mem -= info.bufmem;
+    info.memeasyfree += info.bufmem;
+  }
+#endif
 }
 
 /* Ohkie to return negative values for temperatures */
--- a/src/linux.cc
+++ b/src/linux.cc
@@ -61,6 +61,7 @@
 #define _LINUX_IF_H
 #endif
 #include <linux/route.h>
+#include <linux/version.h>
 #include <math.h>
 #include <pthread.h>
 #include <atomic>
@@ -186,7 +187,7 @@ int update_meminfo(void) {
    * These variables keep the calculations local to the function and finish off
    * the function by assigning the results to the information struct */
   unsigned long long shmem = 0, sreclaimable = 0, curmem = 0, curbufmem = 0,
-                     cureasyfree = 0;
+                     cureasyfree = 0, memavail = 0;
 
   info.memmax = info.memdirty = info.swap = info.swapfree = info.swapmax =
       info.memwithbuffers = info.buffers = info.cached = info.memfree =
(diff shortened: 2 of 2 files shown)


jbenden:
conky-mem-linux

Here's my current info, while using this PR.

brndnmtthws:
First of all, thanks for the PR & contribution.

I'm going to leave this open for a bit to collect some feedback, because what you're proposing seems fairly significant. Since Conky has been around for more than 10 years, this behaviour might throw some people off. I'd like to maintain backward compatibility for those who already rely on specific behaviours.

An alternative approach may be to simply add new vars with the calculations you want. How does that sound?

jbenden:
I wonder if the reason for oddness stems from VM kernel subsystem changes over time?

In the image below, I've captured my system's stats using the current implementation. It shows ~11G in use, as opposed to the actual ~9G. [Do not take the one column from TOP as matching correctly, graphing that variable would mean the graph grows top-to-bottom.]

Screenshot from 2019-06-30 14-08-26

That said, yes, I'd be willing to convert the patch over to newly added variables. I'd even be willing to modify the patch to account for a certain version of the kernel or newer.

lasers:
>I wonder if the reason for oddness stems from VM kernel subsystem changes over time?

@brndnmtthws For what it's worth, we had to do similar thing with https://github.com/ultrabug/py3status/pull/1493 because the module `sysdata` was easily off with `mem` values, percents, etc too.

Tomunu:
It seems like this is really about whether or not to inlude buffers/cache when calculating memory usage. If that's right, then since there are already both `memfree` and `memeasyfree` to be able to get the free memory with or without including buffers/cache, it seems reasonable to be able to get values for both interpretations of "used" memory. 

Also, it looks like the py3status change that @laser linked to has pretty much the opposite intent of this, since that was about wanting a way to get the memory usage that INCLUDES buffers/cache.

jbenden:
That option doesn't account for memory usage graphing, though.

Tomunu:
So, there could be 5 new variables? One each for `mem`, `membar`, `memgauge`, `memgraph`, and `memperc`. It just seems like there are people who care strongly for both ways of calculating memory. 

Or am I missing something else about graphs? Or is it just that the variables get kind of messy? And maybe it's only a problem for Linux?

jbenden:
Yes, this is Linux only.

I feel that adding new variables is not the correct way to approach this.

I view the current memory values as incorrect, a bug. This PR fixes the bug.

Having said that, I also feel the values are only wrong with which ever Linux version included additional memory statistics, such as `SReclaimable`, etc.

jbenden:
Hello,

I have found the best way to approach this with recent Linux kernels; which, is close to what I originally calculated, but much more accurate than without this patch.

The kernel introduced a new field explicitly for running calculations. See this commit: https://github.com/torvalds/linux/commit/34e431b0ae398fc54ea69ff85ec700722c9da773

Thanks,
-Joe

jbenden:
Please read over the latest changes, as I'm positive this is an acceptable change now.

@lasers @brndnmtthws

PR #864, opened 2019-07-17
Fix `link_qual_max` getting updated with bad value

Tomunu:
This fixes Issue #650  'Bad arguments: " unk" and '59"'

`link_qual_max` would occasionally get updated with an invalid value. Now the update will ignore bad values and only use valid numbers. 

(If anyone is really concerned about the additional clean up and refactor, I don't mind if this gets changed to include only the first commit, which is the one that fixes the problem.)

### DIFF
--- a/src/linux.cc
+++ b/src/linux.cc
@@ -133,10 +133,10 @@ int update_uptime(void) {
   } else
 #endif
   {
-    static int rep = 0;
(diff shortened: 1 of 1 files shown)


brndnmtthws:
Very nice, thank you kindly.

PR #866, opened 2019-07-20
Fix heap use-after-free with setlocale

hexchain:
In glibc the return value of setlocale() is allocated in static storage,
which may be invalidated by subsequent calls.

conky fails to start when built with AddressSanitizer because of this.


brndnmtthws:
Thanks!

PR #871, opened 2019-08-07
Fix 860

CtrlSequence:
**Changes:** completely moved code block:
```c
if (no_buffers.get(*state)) {
  info.mem -= info.bufmem;
  info.memeasyfree += info.bufmem;
}
```
from `common.cc` to `linux.cc`
No other source file use the `bufmem` variable from the `info` struct: they all seem to have their own distinct ways of handling memory quantities, so this block only _specifically applies_ to the `linux.cc` version of the code.
The problem is that by having this in `common.cc`, the memory quantities are incorrect when using `${exep}` and `${execpi}` with linux.
**Testing/Validation:** Unfortunately I don't have the resources to build for any other system, but by looking through each code file I can say with certainty that `info.bufmem` only appears in `linux.cc`. Outside of that, I have the Arch package version to test against which is ver. `conky 1.11.4`.
I made a few simple configs and scripts to use for testing
conky.conf
```lua
conky.config = {
    no_buffers = true,
    out_to_x = false,
    out_to_console = true,
    update_interval = 1.0,
    total_run_times = 1,
    short_units = true,
}
conky.text = [[
${execp ./test.sh }
Total memory:   ${memmax}
memperc:        %${memperc}
mem:            ${mem}
memfree:        ${memfree}
memdirty:       ${memdirty}
memeasyfree:    ${memeasyfree}
]]
```
and test.sh
```sh
#!/bin/zsh
print 'Test'
```
Here's a screenshot of the output of both versions using the config/script:
conky vs ./conky
The first instance is the Arch package build and the second is the compiled version from my branch.
There's a noticeable difference between the two versions for `${memperc}`, `${mem}`, and `${memeasyfree}`. However, htop only reports me using 1.70G:
htop

### DIFF
--- a/src/common.cc
+++ b/src/common.cc
@@ -293,14 +293,6 @@ void update_stuff() {
 
   /* if you registered a callback with conky::register_cb, this will run it */
   conky::run_all_callbacks();
-
-#if !defined(__linux__)
-  /* XXX: move the following into the update_meminfo() functions? */
-  if (no_buffers.get(*state)) {
-    info.mem -= info.bufmem;
-    info.memeasyfree += info.bufmem;
-  }
-#endif
 }
 
 /* Ohkie to return negative values for temperatures */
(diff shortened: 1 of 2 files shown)


brndnmtthws:
@labhub-bot retry

Issue #877, opened 2019-08-14
Negative RAM Usage


leogx9r:
**Issue**

Recently updated to version v1.11.5. After this update, I've noticed that RAM usage will be negative when using the `$mem` variable as seen below which did not occur

![image](https://user-images.githubusercontent.com/40126089/63051039-e7511500-becb-11e9-8935-38397c60c8a3.png)

**Information**

OS: `Arch Linux`
Package: `conky-nvidia`
Version: `conky 1.11.5_pre`

Relevant configuration to produce the image above:

```lua
conky.config = {
    --========================================================================

    background = true,
    update_interval = 2,

    -- Info Settings

    cpu_avg_samples = 2,
    net_avg_samples = 2,

    double_buffer = true,
    no_buffers = true,

    format_human_readable = true,

    -- Window Settings

    own_window = true,
    own_window_type = 'desktop',
    own_window_argb_visual = true,
    own_window_argb_value = 0,
    own_window_hints = 'undecorated,below,sticky,skip_taskbar,skip_pager',

    gap_x = 1300,
    gap_y = 0,
    minimum_width = 600,
    maximum_width = 600,

    border_inner_margin = 40,
    border_outer_margin = 0,
    draw_graph_borders = true,

    alignment = 'top_right',

    -- Display Settings

    use_xft = true,

    font = 'Ubuntu Mono:regular:size=10',

    default_color = '#FFFFFF',
    own_window_colour = '#000000',

    -- Local Variables

	color0 = 'green',
	color1 = 'gray',
	color2 = '#1793D1',
	color3 = 'yellow',
	color4 = 'red',

    --========================================================================
};


conky.text = [[
    ${goto 50}${font FontAwesome}${font}${voffset -2} ${mem}/${memmax} ( ${memperc} % ) ${color2}|${color} ${font FontAwesome}${font}${voffset -2} ${swapperc}% ${goto 310}${swapbar 7,305}
    
    ${goto 60}${memgraph 30,555 AAF5D0 00B35B}
    ${goto 40}${color2}${hr 2}${color}
]];
```

Logs don't see anything wrong as there are no errors so I don't think they're necessary here.

DaruSupaHaka:
I'm also having this problem.

OS: `Arch Linux`
Package: `conky`
Version: `conky 1.11.5_pre`

![RAM](https://user-images.githubusercontent.com/49533914/63068382-9a277000-bed8-11e9-8411-4d3291ef8693.PNG)

Config:
```
[...]

PR #878, opened 2019-08-18
Fix for #877

CtrlSequence:
### Description
Race condition happens due to separate calls to `update_meminfo()` with multi-threading enabled, `no_buffers = yes` set in configuration, and using `${top}` objects. Applies only to Linux builds.

A call to `run_all_callbacks()` runs `exec_cb::work()`, which spawns a thread for `update_meminfo()` and one for `update_top()`. Since `update_top()` calls `update_meminfo()`, this means there are two separate threads running `update_meminfo()` in a slightly staggered timing, causing the `info.mem -= info.bufmem` to be applied twice with the same values.

#### How the changes will affect existing behaviour
This should prevent wrong values being reported due to double-booking `info.mem` updates. 
Since `run_all_callbacks()` is responsible for regular updates, separate calls to functions registered as a callback should be avoided. Judging by the comment that I removed, this call seems to be lingering from a time before callbacks were implemented.

#### Describe how you tested and validated your changes.
[Outsourcing](https://github.com/brndnmtthws/conky/issues/877#issuecomment-522187310)

Any sort of configuration that utilizes any `${top}` objects and has `no_buffers = yes` is what causes it. Take: 
```
conky.config = {
    no_buffers = true,
    out_to_x = false,
    out_to_console = true,
    update_interval = 3.0,
    short_units = true,
}
conky.text = [[ RAM ${mem} ${top_mem mem_res 1} ]]
```
Which with `conky 1.11.5` produces:
2019-08-18-104822_398x206_scrot
Compared to this commit version which with the same configuration produces:
2019-08-18-104951_398x146_scrot

### DIFF
--- a/src/top.cc
+++ b/src/top.cc
@@ -390,10 +390,6 @@ static void process_find_top(struct process **cpu, struct process **mem,
 }
 
 int update_top() {
-  // XXX: this was a separate callback. and it should be again, as soon as it's
-  // possible
-  update_meminfo();
-
   process_find_top(info.cpu, info.memu, info.time
 #ifdef BUILD_IOSTATS
                    ,


vmxoxo:
The values briefly drop back to low false values at intervals.

OS: Arch Linux
Package: conky
Version: conky 1.11.5-2

conky -c conky.conf
 RAM 3,91G 2,29G 
 RAM 3,91G 2,29G 
 RAM 186M 2,29G 
 RAM 183M 2,29G 
 RAM 180M 2,29G 
 RAM 196M 2,29G 
 RAM 192M 2,29G 
 RAM 189M 2,29G 
 RAM 187M 2,29G 
 RAM 3,90G 2,29G

PR #892, opened 2019-10-07
add SI unit to the limit value shown in graphs

jbenden:
Currently, when a graph displays the scale value it does not display the value in an easily known SI unit. When simply looking at a value, it could be bytes, kilobytes, megabytes, etc...

This patch resolves this problem. When `format_human_readable = true` and `show_graph_scale = true` are enabled, all graphs will now show the SI unit associated with the value. Even the networking graph will be displayed properly.

This is an image showing off this change (w/short-units):
Screenshot from 2019-10-06 19-29-51

### DIFF
--- a/src/conky.cc
+++ b/src/conky.cc
@@ -584,10 +584,10 @@ void human_readable(long long num, char *buf, int size) {
   }
   if (short_units.get(*state)) {
     width = 5;
-    format = "%.*f%.1s";
+    format = "%.*f %.1s";
   } else {
     width = 7;
-    format = "%.*f%-.3s";
+    format = "%.*f %-.3s";
   }
 
   if (llabs(num) < 1000LL) {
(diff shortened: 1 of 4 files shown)


lasers:
2019-10-07-023813_291x1178_scrot

Low graph `124.2 KiB` vs high graph `5.1 B`. Is the graph right? 

----

2019-10-07-024217_958x1178_scrot
Different screenshot with different numbers.
```
${downspeedgraph eno1}
${downspeedgraph eno1 -t}
${downspeedgraph eno1 -l}
```

jbenden:
I have updated the logic to include log scale graphs.

Thanks,
-Joe

jbenden:
Here's a screen shot showing the top part as my normal display I use, and the graphs below are based on the config snippet sent.

Screenshot from 2019-10-07 22-31-13

src/conky.cc
```diff
@@ -584,10 +584,10 @@ void human_readable(long long num, char *buf, int size) {
   }
   if (short_units.get(*state)) {
     width = 5;
-    format = "%.*f%.1s";
+    format = "%.*f %.1s";
   } else {
     width = 7;
-    format = "%.*f%-.3s";
+    format = "%.*f %-.3s";
```
brndnmtthws: If you're adding an extra character, would you not also need to increase the width?

jbenden: Not sure. It's locally working as far as I can see. I can certainly add the extra, if you'd like.

brndnmtthws: Actually, never mind. I forgot what the `width` argument did with `spaced_print()`, and after review it I can see it applies to the width of the value being printed, so it's fine in this case.

PR #895, opened 2019-10-16
properly format PIDs longer than 5 digits

ndwarshuis:
64-bit Linux allows PIDs up to 4194304 (7 digits), but current code will truncate any PID longer than 5 digits

### DIFF
--- a/data/conky.conf
+++ b/data/conky.conf
@@ -77,7 +77,7 @@ ${color grey}File systems:
 ${color grey}Networking:
 Up:$color ${upspeed} ${color grey} - Down:$color ${downspeed}
 $hr
-${color grey}Name              PID   CPU%   MEM%
+${color grey}Name              PID     CPU%   MEM%
 ${color lightgrey} ${top name 1} ${top pid 1} ${top cpu 1} ${top mem 1}
 ${color lightgrey} ${top name 2} ${top pid 2} ${top cpu 2} ${top mem 2}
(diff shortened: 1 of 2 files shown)


brndnmtthws:
> @brndnmtthws Do we need to change alignments for default `data/conky.conf` too?
> 
> Here's the `default.conf` for 7char PID.
> 
> ```diff
> diff --git a/data/conky.conf b/data/conky.conf
> index e75b5f1..9a244b0 100644
> --- a/data/conky.conf
> +++ b/data/conky.conf
> @@ -77,7 +77,7 @@ ${color grey}File systems:
>  ${color grey}Networking:
>  Up:$color ${upspeed} ${color grey} - Down:$color ${downspeed}
>  $hr
> -${color grey}Name              PID   CPU%   MEM%
> +${color grey}Name              PID     CPU%   MEM%
>  ${color lightgrey} ${top name 1} ${top pid 1} ${top cpu 1} ${top mem 1}
>  ${color lightgrey} ${top name 2} ${top pid 2} ${top cpu 2} ${top mem 2}
>  ${color lightgrey} ${top name 3} ${top pid 3} ${top cpu 3} ${top mem 3}
> ```
> 
> It would be 99% empty. Same as `CPU%` and `MEM%` 99% not reaching `100.00`.

That's a good idea.

@ndwarshuis would you mind updating the original PR accordingly?

PR #907, opened 2019-11-22
Fix segfault on pulseaudio tunnel sinks

paretje:
The code in `pa_sink_info_callback` handling active port information assumes there is always an active port. That's isn't the case for tunnel sinks, resulting in a segmentation fault as `i->active_port` is null.


brndnmtthws:
Thanks!

PR #911, opened 2019-11-28
Remove length limit on curl URLs

adeason:
Currently, the ${curl} directive only uses the first 127 characters in
the given URL, discarding the rest. Change the parsing in
curl_parse_arg to dynamically allocate the URL string instead,
allowing ${curl} to use URLs of any length.

127 chars might seem like a lot, but it's can be easy to hit this when using ${curl} with various REST-y APIs, which I find it useful for :)

Also, apologies if the string manipulation is ugly; I have a C background and I'm not familiar if there are more C++-friendly ways of doing this. There is an sscanf %ms code to dynamically allocate the string instead of restricting its length, but I don't think that code is universally understood, so I assumed that would be undesirable.

### DIFF
--- a/src/ccurl_thread.cc
+++ b/src/ccurl_thread.cc
@@ -166,7 +166,7 @@ class simple_curl_cb : public curl_callback<std::string> {
  */
 
 struct curl_data {
-  char uri[128];
+  char *uri;
   float interval;
 };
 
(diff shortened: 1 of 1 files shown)


AlexApps99:
@adeason `conky/src/ccurl_thread.cc:183:7: error: unused variable ‘argc’ [-Werror=unused-variable]`

adeason:
On Wed, 22 Jan 2020 02:27:20 -0800
AlexApps99 <notifications@github.com> wrote:

> @adeason `conky/src/ccurl_thread.cc:183:7: error: unused variable ‘argc’ [-Werror=unused-variable]`

Thanks for the ping; sorry I didn't realize this had failed. This has
been fixed, and it looks like the builders succeeded this time.

-- 
Andrew Deason
adeason@dson.org

PR #924, opened 2020-01-06
Fix if_mounted & if_running variables for FreeBSD

helmat:
if_mounted and if_running were dysfunctional in FreeBSD. This rather trivial modification enables them again. Tested on my 12-STABLE machine.


madpilot78:
I'm testing this for adding it to the FreeBSD port. I will report back with the results.

This is also being tracked in the FreeBSD bugzilla: https://bugs.freebsd.org/243088

PR #927, opened 2020-01-14
Added lower_case and upper_case commands

AlexApps99:
Each command takes in input and converts it into upper/lower case, to compliment `start_case`.

Examples:
- `${lower_case ABC d e fG}` = "abc d e fg"
- `${nodename}` = "Alex's PC"
- `${lower_case ${nodename}}` = "alex's pc"
- `${upper_case ${nodename}}` = "ALEX'S PC"
- `${upper_case}` = ""

I added these so there is a way to change the case of text individually per-text.

I based them off of `start_case` but simplified them a lot more to make it easier to read and maintain.
I have corrected all formatting mistakes with `make clang-format`.

### DIFF
--- a/doc/variables.xml
+++ b/doc/variables.xml
@@ -3847,7 +3847,27 @@
             </command>
             <option>text</option>
         </term>
-        <listitem>All words capitalized regardless.
+        <listitem>Capitalises the start of each word.
+        <para /></listitem>
+    </varlistentry>
+    <varlistentry>
+        <term>
+            <command>
+                <option>lower_case</option>
+            </command>
+            <option>text</option>
+        </term>
+        <listitem>Converts all letters into lowercase.
+        <para /></listitem>
+    </varlistentry>
+    <varlistentry>
+        <term>
+            <command>
+                <option>upper_case</option>
+            </command>
+            <option>text</option>
+        </term>
+        <listitem>Converts all letters into uppercase.
         <para /></listitem>
     </varlistentry>
     <varlistentry>
--- a/doc/variables.xml
+++ b/doc/variables.xml
@@ -3843,7 +3843,7 @@
     <varlistentry>
         <term>
             <command>
-                <option>start_case</option>
+                <option>startcase</option>
             </command>
             <option>text</option>
         </term>
@@ -3853,7 +3853,7 @@
     <varlistentry>
         <term>
             <command>
-                <option>lower_case</option>
+                <option>lowercase</option>
             </command>
             <option>text</option>
         </term>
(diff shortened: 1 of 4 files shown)


AlexApps99:
If you want, I can also add a `lowercase` option and make the naming more consistent as part of this PR.
Totally up to you though

AlexApps99:
I want to also add a `lowercase` configuration setting, but I don't know how to handle `uppercase` and `lowercase` both being set to `true`.
Is there any other way this could be implemented, such as an enum or string value?
Please share your thoughts on this

AlexApps99:
I rewrote `print_startcase` (formerly `print_cap`) because it was unneccesarily complicated and did not correctly handle words with capital letters in the middle of them. Feedback on this implementation would be welcome, and let me know if it would be better to keep capitals in the middle of words as capitals.

AlexApps99:
The check failed because of a package manager bug :joy: 
https://gitlab.com/brndnmtthws-oss/conky/-/jobs/406322378#L24

brndnmtthws:
@labhub-bot retry

AlexApps99:
It failed again, this time due to a timeout... :man_facepalming:
https://gitlab.com/brndnmtthws-oss/conky/-/jobs/407527135
FreeBSD seems to be quite unreliable on GitLab, having failed twice now

brndnmtthws:
@labhub-bot retry

brndnmtthws:
@AlexApps99 unfortunately the FreeBSD builds are really flaky, so I might just disable them or make the failures warnings instead of errors. If it doesn't work this time, I think we can just ignore the failure and merge anyway.

AlexApps99:
Thanks @brndnmtthws 
I would like some feedback/review [here](https://github.com/brndnmtthws/conky/pull/927#issuecomment-575401081) and [here](https://github.com/brndnmtthws/conky/pull/927#issuecomment-575445702), once that happens we should be able to merge this :)

AlexApps99:
Hello, if you could leave some feedback that'd be great.
Thanks

brndnmtthws:
> I want to also add a `lowercase` configuration setting, but I don't know how to handle `uppercase` and `lowercase` both being set to `true`.
> Is there any other way this could be implemented, such as an enum or string value?
> Please share your thoughts on this

I'd apply KISS and just assume people aren't going to set both. If they do set both, then the last setting should take precedence.

> I rewrote `print_startcase` (formerly `print_cap`) because it was unneccesarily complicated and did not correctly handle words with capital letters in the middle of them. Feedback on this implementation would be welcome, and let me know if it would be better to keep capitals in the middle of words as capitals.

It looks good to me.

PR #949, opened 2020-02-26
Print uid in top if user doesnt exist.

pawciobiel:
The change is necessary for displaying uid instead of blank place if user doesn't exist (in top).
Changed function top.cc `print_top_user()` will add uid (decimal) value to top as opposed to adding blank and printing error to stderr.

#948


pawciobiel:
@brndnmtthws @su8 would this be ok?

PR #973, opened 2020-06-05
make cmake builds reproducible

Segaja:
While working on the [reproducible build](https://reproducible-builds.org/) project, I noticed that conky could not be build reproduced.

This patch makes the output consistent across builds by using [SOURCE_DATE_EPOCH](https://reproducible-builds.org/docs/source-date-epoch/) and `uname -sm` (by omitting the `-r` the kernel version is gone).


brndnmtthws:
Thanks for the PR!

Issue #989, opened 2020-08-01
Bug when using "nvidia fanspeed N"


humanoid2050:
**Issue**

I have dual Nvidia RTX 2080 Super GPUs. Each gpu has two fans, for a total of 4 in the system. I can only query the fans on the first GPU, which are indexed as 0 and 1. Requesting the higher indexed fans (i.e. 2 and 3 on the second GPU) returns the parameter for fan at index 1.

**Information**
The offending variables are of the form 
```
${nvidia fanspeed 0}
${nvidia fanspeed 1}
${nvidia fanspeed 2}
${nvidia fanspeed 3}
```

I believe the issue arises at line 945 of nvidia.cc.I have copied the surrounding code here.

```
if (nvs != nullptr) {
    // Reduce overcommitted GPU number to last GPU
    if (nvs->gpu_id > num_GPU) nvs->gpu_id = num_GPU;

    // Execute switch by query type
    switch (nvs->query) {
      case QUERY_VALUE:
        switch (nvs->attribute) {
          case ATTR_FAN_LEVEL:
            /* falls through */
          case ATTR_FAN_SPEED:
            if (num_COOLER < 1) {
              value = -1;
              break;
            }
            /* falls through */
          default:
            value = get_nvidia_value(nvs->target, nvs->attribute, nvs->gpu_id,
                                     nvs->arg);
            break;
        }
        break;
```
It seems that when the variable is parsed, the numeric arg is assumed to exclusively refer to GPU index and is therefore capped at the number of GPUs detected. In the context of fans, I believe this is incorrect behavior. Unfortunately, I don't have a super clean suggestion for handling this issue, as altering the interpretation of the number could have ripple effects in the code.

PR #1007, opened 2020-09-25
made fan data available for indices higher than the gpu card count, a…

humanoid2050:
…nd added some protection against bad requests

**Descriptions**
* Fan indices are no longer capped at the number of GPUs (helpful if you have multifan GPUs)
* Renamed some internal variables to reflect that they are not strictly gpu id's
* Compiled and ran changes
* ref bug #989 

**Licenses**
* No changes

### DIFF
--- a/src/nvidia.cc
+++ b/src/nvidia.cc
@@ -40,55 +40,55 @@
  *
  * Showcase (conky.conf):
  * --==| NVIDIA | ==--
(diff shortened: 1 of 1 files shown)


humanoid2050:
I just saw that there is a build pipeline error. It appears to be related to an optional dependency on microhttpd if the BUILD_HTTP CMake flag is set ON. Seems more like a pipeline configuration error.

PR #1011, opened 2020-10-03
Fix build with older microhttpd

mmuman:
On Debian Sid with libmicrohttpd-dev 0.9.71-1 this is needed to build.


mmuman:
Of course they added this this year so older version won't have it :-(

mmuman:
Ok, seems they removed a define at the same time so we can ifdef on it.

mmuman:
I noticed you added a fix, without fallback for older versions of the lib, I wonder how many distros out there have an old version…

mmuman:
I see you fixed microhttpd support differently than on this PR, however this might be required for older versions of the library which might still be around for some time?

Issue #1018, opened 2020-10-09
RFE: units spacing


altblue:
Commit dae65af726fab569c1073af16b3033c8813f893a silently adds a blank space before units, inadvertently changing the way such values are displayed across the board.

i.e. where one had `12.3M` now they get `12.3 M`.

A solution to alleviate this mishap might be adding a boolean option similar to `short_units` –possibly named `spaced_units`‒ to control this extraneous `space`. Obviously, for backward (i.e. _before_ that commit) compatibility, its default value should be `false`.

Another would be to allow users to choose whatever _spacer_ they want, e.g. a `units_spacer` string option, defaulting to `""` (_nothing_).

altblue:
b7a0e926566a84e281cf7f1077ab9763f8a793d2 implements `units_spacer`.

Test:

```
$ c=conky-test.lua ; cat $c ; conky -i 3 -c $c ; sed -i 's/--//' $c ; cat $c ; conky -i 3 -c $c
conky.config = {
    out_to_x = false,
    out_to_console = true,
    short_units = true,
    --units_spacer = '··',
}
conky.text = [[${diskio_read}]]
0B
299K
838K
conky.config = {
    out_to_x = false,
    out_to_console = true,
    short_units = true,
    units_spacer = '··',
}
conky.text = [[${diskio_read}]]
0··B
609··K
1.47··M
```

wolfiestyle:
Same issue. Just upgraded and noticed the extra spaces that look quite bad on fixed width fonts.

txtsd:
All my units have a space before them! I'd like the default behavior to be unchanged, with an option to insert spaces if needed.

brndnmtthws:
Can you verify that #1094 fixes this correctly? It's based on @altblue's original patch.

txtsd:
Just tested it. It does get rid of the spaces!

PR #1019, opened 2020-10-09
add units_spacer option for tuning string between values and units

altblue:
A solution for #1018:
- restores the usual output before dae65af726fab569c1073af16b3033c8813f893a, i.e. no _spaces_ between _values_ and _units_;
- offers the ability to customize "spacer", like [thin spaces](https://en.wikipedia.org/wiki/Thin_space).

### DIFF
--- a/doc/config_settings.xml
+++ b/doc/config_settings.xml
@@ -1041,6 +1041,15 @@
     <varlistentry>
         <term>
             <command>
+                <option>units_spacer</option>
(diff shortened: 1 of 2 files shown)


gjs278:
please merge this or revert the addition of these spaces. it has completely ruined my alignment on conky. I'm freezing my package manager to the version before this one, it it is far too distracting.

PR #1028, opened 2020-11-07
Update linux.cc : to show the same used RAM than free

nihilix-melix:
**Descriptions**
The memory available field is describe in /proc man page as : "An estimate of how much memory is available for starting new applications, without swapping."
We can see in page_alloc.c in the linux kernel that : "Not all the page cache can be freed, otherwise the system will start swapping. Assume at least half of the page cache, or the low watermark worth of cache, needs to stay."
There is a difference between the system starts swapping and OOM happens even if there is no swap on the system.
People think they will find the same amount of used RAM with the free program. From man free, we can see :
     used   Used memory (calculated as total - free - buffers - cache)
     free   Unused memory (MemFree and SwapFree in /proc/meminfo)
     buffers Memory used by kernel buffers (Buffers in /proc/meminfo)
     cache  Memory used by the page cache and slabs (Cached and SReclaimable in /proc/meminfo)
So, curmem should be calculate as follow to find the same amount of memory in conky and free.
    curmem = info.memmax - (info.memfree + info.buffers + info.cached + sreclaimable);

**Licenses**
The contributed code is GPLv3 licensed.

### DIFF
--- a/src/linux.cc
+++ b/src/linux.cc
@@ -247,26 +247,26 @@ int update_meminfo(void) {
     */
     curmem -= curbufmem;
     cureasyfree += curbufmem;
-#else
-#   curmem = info.memmax - memavail;
+#else  /* LINUX_VERSION_CODE <= KERNEL_VERSION(3, 14, 0) */
     /* The memory available field is describe in /proc man page as :
-     * "An estimate of how much memory is available for starting new applications, without swapping."
-     * We can see in page_alloc.c in the linux kernel that :
-     * "Not all the page cache can be freed, otherwise the system will start swapping.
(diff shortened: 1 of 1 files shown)


src/linux.cc
```diff
@@ -248,7 +248,23 @@ int update_meminfo(void) {
     curmem -= curbufmem;
     cureasyfree += curbufmem;
 #else
-    curmem = info.memmax - memavail;
+#   curmem = info.memmax - memavail;
```
brndnmtthws, line 251: Remove this line.

brndnmtthws (changes requested):

Issue #236, opened 2016-04-06
Text and image alignment issues


zagortenay333:
It would nice if there was a way to position images relative to lines and not pixels. This way, even when line heights change due to font size or type, the images would remain aligned to specific lines.

Right now, it's enough to change the distro to ruin the alignment because the font rendering will be different and will affect the line heights as well.

nicolasdanelon:
:+1: this is really necessary !!

greatquux:
Ooh - that's true. Now I feel like discounting my report completely because I use a PPA that changes font rendering, and I never thought it would change things subtly enough to affect this. (It's https://launchpad.net/~no1wantdthisname/+archive/ubuntu/ppa which applies infinality's modifications for freetype and fontconfig.) 

Of course, it's also important to remember this project is based on conky and positioning relative to lines as opposed to pixels may not be possible with conky (I don't know). It might just have to be part of the documentation or set up in a way that is more easy to modify.

plikhari:
Why would you want to position images relative to lines and not pixels - do you understand how fonts are rendered ???
Each font size refers to the height of that font in pixels. Your image - unless it is the size of the font will be much larger. 
Your issue is a normal issue on an open system such a Linux. Changes happening up-stream may not be taking into account many many things. This issue is not just with fonts it is with everything with developers who may not be mature enough to understand you have 96 dpi and they figure their 72 dpi is the universe.
Do not loose heart ::D - there is a work around.
```
use_xft = true,
xftalpha = 0.8,
font = 'DejaVu Sans:size=12:style=bold',
```
See above the default font size is declared as 12. If the font rendering engine gets updated - all you need to do is change that font size and things will start falling into place.

zagortenay333:
Why would I need to know how fonts are rendered? I only know what the issue is and that's the main point of my post, not to provide a solution.

Not sure what the rest of your comment has to do with this issue.

zagortenay333:
I suppose it would be a lot of work to fix this.

Nothing short of some css-like rendering could solve this. If I draw some box, it would have to be able to resize depending on the font; otherwise, the text would always end up out of alignment..
This makes it impossible to create widgets like the harmattan conky linked above. You can only do floating text

syco:
I couldn't agree more with this request, a solution for me would be to scrap the `-p x,y` and align the top left corner of the image where the next character would be, this should respect `${goto}`, `${offset}`, and `${voffset}`.
This would be enough to align images with lines and make things more portable.

[...]

PR #1055, opened 2021-01-30
Fix for issue #236

syco:
**Descriptions**
So images defy completely the logic of conky as they don't use the print functions all other shortcodes use.
Because of this I found hard to find a point in `image.cc` where I could ignore `-s x,y` and use instead `cur_x` and `cur_y` from `conky.cc`.

My solution uses a new shortcode that saves the current coordinates into an array, this array can be accessed at any time by `image` to know where the image should be rendered.
Note that the image doesn't hold any size, doesn't move the cursor forward or down, so you still need to play with `goto` and `offset` to align your text.

The configuration file I used to test these changes is this:
```txt
conky.text = [[
${alignc}${time %H:%M}
${alignc}${time %A}
${alignc}${time %d %B, %Y}

${save_coordinates 0}${image /tmp/00.png -i 0 -s 65x65 -n}${goto 95}Today
${goto 95}Rather cloudy, a little rain
${goto 95}Min 6° / Max 7°
${goto 95}Chance of Rain: 55%}${voffset 5}

${hr}
${goto 10}${save_coordinates 1}${image /tmp/00.png -i 1 -s 40x40 -n}${offset 50}Tomorrow
${goto 60}Cloudy, a little rain; breezy
${goto 60}Min 4° / Max 10°

${hr}
${goto 10}${save_coordinates 2}${image /tmp/00.png -i 2 -s 40x40 -n}${offset 50}Wednesday
${goto 60}Mostly cloudy, showers around
${goto 60}Min 4° / Max 8°

${hr}
${goto 10}${voffset -5}${save_coordinates 3}${image /tmp/00.png -i 3 -s 40x40 -n}${voffset 5}${offset 50}Thursday
${goto 60}Mostly cloudy, showers around
${hr}
]]

```

as you can see, before I use `image`, I `save_coordinates` into a positional array, and then tell the image what position I used with the `-i` switch.
This can easily be scripted, I found it very convenient in the script that generate my configuration.
And it removes the need to know beforehand font size or the number of lines.
The only think I noticed on a different pc is that it might require `voffset` to be used before and after the image, as I do in the last one of the configuration. maybe `last_font_height` is not the right value to subtract.

And this is the result:
image

I haven't written any documentation yet, I'm hoping to get some opinions before I spend more time on it.

Thanks.


brndnmtthws:
I think this looks alright. The image feature was kind of shoehorned in, and the rest of the sizing is based on text so yeah, it's not great. I'll take this PR because it seems like an improvement.

trappedinspacetime:
@brndnmtthws @syco Thank you for implementing this feature. I have a problem with it. I have a python moon phase calculator script which prints the path of the moon phase  image by date in local directory. 
I want it to be shown at bottom of the conky, but  at first it appears at top of conky when conky launches  then it's shown at bottom of the conky.
I tried a trick

          ${if_updatenr 3}${save_coordinates 0}${endif}
          ${save_coordinates 0}${execp  python3  ~/.conky/default-screenlet/MPC.py}
but `if_updatenr` doesn't run asynchcronously, it waits to execute whole lines above it including the last line, so it doesn't provide a solution.
But I noticed that `lua` scripts which is loaded via `lua_draw_hook_post = 'main'` provide a working solution.
            local updates = conky_parse('${updates}')
            update_num = tonumber(updates)
            
            if update_num > 5 then
                go_clock_rings(display)
                go_gauge_rings(display)
            end
How can I do it with python script?

PR #1067, opened 2021-02-26
Respect CFLAGS/CXXFLAGS

billie80:
**Descriptions**
In commit bb8723d "add toluapp subtree" support for user defined FLAGS was removed. I do not see a reason why this should not be allowed, so it would be great if this could be added again.

Reference: https://bugs.gentoo.org/772176

### DIFF
--- a/cmake/ConkyBuildOptions.cmake
+++ b/cmake/ConkyBuildOptions.cmake
@@ -37,10 +37,10 @@ if(NOT CMAKE_BUILD_TYPE)
 endif(NOT CMAKE_BUILD_TYPE)
 
 # -std options for all build types
-set(CMAKE_C_FLAGS "-std=c99"
+set(CMAKE_C_FLAGS "-std=c99 ${CMAKE_C_FLAGS}"
     CACHE STRING "Flags used by the C compiler during all build types."
     FORCE)
-set(CMAKE_CXX_FLAGS "-std=c++17"
+set(CMAKE_CXX_FLAGS "-std=c++17 ${CMAKE_CXX_FLAGS}"
     CACHE STRING "Flags used by the C++ compiler during all build types."
     FORCE)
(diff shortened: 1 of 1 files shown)


brndnmtthws:
I do recall there being a reason for that change, but I can't remember precisely what it was. If I were to guess, it may have had to do with the way CMake was behaving when trying to override the variable.

[From the docs](https://cmake.org/cmake/help/latest/command/set.html#set-cache-entry) it sounds like CMake should respect setting the flags, so long as you set them _before_ the value is cached.

billie80:
I don't know if I understood this correctly, but does this actuallt matter if the FORCE option is used?
In the mentioned commit this option was also removed but probably added back at a later time.

If it is really an issue I can probably patch it locally but from what I read it seems like a good idea supporting it. And also to be honest it would spare me some work :-)

The first comments of the Gentoo tracker bug https://bugs.gentoo.org/59506 gives a few reasons why this is a good thing.

brndnmtthws:
Oh, for some reason I didn't see `FORCE` there. I think, in that case, this is probably the right thing to do. Let me test it to make sure it behaves as expected.

billie80:
Great!
FWIW, here it allows me to inject my flags and Gentoo's QA checks are happy in this regard ;-)

brndnmtthws:
I guess the annoying thing about how CMake handles this is that it breaks the caching when you do this. If I specify the cflags, run cmake, then run cmake again without specifying the flags (i.e., relying on the cache) then the value revert to the defaults. This is pretty annoying if you're doing dev work, which is probably why I removed this in the first place.

Anyway, I guess your PR is the lesser of evils. We'll sacrifice a bit of convenience for the sake of making it easier to package.

billie80:
Thank you very much for considering this!

Issue #944, opened 2020-02-11
conky graph does not support HiDPI screen display like Mac Retina


stardiviner:
I installed Arch Linux on MacBook Pro 2015 (with Retina screen).

I followed Arch Linux wiki https://wiki.archlinux.org/index.php/HiDPI to setup for HiDPI. I setted Xresources `Xft.dpi` etc, seems Conky font folloed those settings. But Conky graph does not. I have to manully change graph size parameters. Is there a way for Conky to auto support HiDPI screen?

PR #1083, opened 2021-04-08
Scale UI elements according to Xft.dpi

livanh:
As in the title, this PR changes the size of all UI elements (graphs, sizes, offsets...) depending on the Xft.dpi X resource. Xft.dpi is the recommended way to scale application graphics on HiDPI displays, but it currently breaks conky: only xft text is scaled correctly, everything else is not. This results in many inconsistencies when Xft.dpi is set to anything other than the default value (96).See also issue #944 and my screenshots below. 

With this PR, any size in pixels is scaled by the appropriate amount before being used for rendering. A DPI value of 96 means no scaling, 192 is 2x scaling, and so on. Scaling by non-integer amounts is rounded to the closest pixel size, e.g. 2 px scaled by 160% is 3.2 px, so it is rounded to 3 px.

If use_xft is false, then text is not scaled, so I also kept the UI elements unscaled. Alternatively, it should be possible to calculate an equivalent scaled size for the non-xft font, and thus scale both text and UI elements even when use_xft = false. Please let me know if the non-xft case should be handled in this way. Also, if any other change is considered necessary, just tell me.

Screenshots follow, to demonstrate the problem and how it is solved. They all use the same configuration file, so that the differences among them are only due to the Xft.dpi setting and to the way it is handled.

<details>
<summary>Standard dpi (96):</summary>

unscaled
</details>
<details>
<summary>Increased dpi (160), without this PR:</summary>

scaled_wrong
</details>
<details>
<summary>Increased dpi (160), with this PR:</summary>

scaled_ok
</details>

### DIFF
--- a/src/conky.cc
+++ b/src/conky.cc
@@ -58,6 +58,9 @@
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wvariadic-macros"
 #include <X11/Xutil.h>
+#ifdef BUILD_XFT
+#include <X11/Xlib.h>
+#endif /* BUILD_XFT */
 #pragma GCC diagnostic pop
 #include "x11.h"
 #ifdef BUILD_XDAMAGE
@@ -495,6 +498,19 @@ int calc_text_width(const char *s) {
 #endif /* BUILD_X11 */
 }
 
+#ifdef BUILD_X11
+#ifdef BUILD_XFT
+int xft_dpi_scale(int value) {
+    if (use_xft.get(*state)) {
+        return (value * xft_dpi + (value>0?48:-48)) / 96;
+    } else {
+        return value;
+    }
+}
+#endif /* BUILD_XFT */
+#endif /* BUILD_X11 */
+
+
 /* formatted text to render on screen, generated in generate_text(),
  * drawn in draw_stuff() */
 
@@ -835,8 +851,8 @@ int get_string_width(const char *s) { return *s != 0 ? calc_text_width(s) : 0; }
 
 #ifdef BUILD_X11
 static inline int get_border_total() {
-  return border_inner_margin.get(*state) + border_outer_margin.get(*state) +
-         border_width.get(*state);
+  return DPI_SCALE(border_inner_margin.get(*state)) + DPI_SCALE(border_outer_margin.get(*state)) +
+         DPI_SCALE(border_width.get(*state));
 }
 
 static int get_string_width_special(char *s, int special_index) {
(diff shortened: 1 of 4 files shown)


src/conky.h
```diff
@@ -301,6 +301,18 @@ void set_updatereset(int);
 int get_updatereset(void);
 int get_total_updates(void);
 
+#ifdef BUILD_X11
+#ifdef BUILD_XFT
+static int xft_dpi;
+int xft_dpi_scale(int value);
+#define DPI_SCALE(a) xft_dpi_scale(a)
```
brndnmtthws, line 308: Rather than using a macro, this can just be a plain C function.

src/conky.cc
```diff
@@ -495,6 +498,19 @@ int calc_text_width(const char *s) {
 #endif /* BUILD_X11 */
 }
 
+#ifdef BUILD_X11
+#ifdef BUILD_XFT
+int xft_dpi_scale(int value) {
```
brndnmtthws, line 503: Change the inner body to something like this to avoid the macro:

```c
#if defined(BUILD_X11) && defined(BUILD_XFT)
    if (use_xft.get(*state)) {
        return (value * xft_dpi + (value>0?48:-48)) / 96;
    } else {
        return value;
    }
#else /* defined(BUILD_X11) && defined(BUILD_XFT) */
    return value;
#endif /* defined(BUILD_X11) && defined(BUILD_XFT) */
```

brndnmtthws (changes requested):

brndnmtthws (approved):

sven337:
Note that this change broke existing configuration files for high DPI, so while it's certainly a good idea, it's also annoying for end users.
You may want to cover that in the documentation somewhere?

nrudzis:
Downgraded for now due to this. The drawings look better in high res. Even after adjusting the sizing, everything looks chunky. A switch to turn off would be a nice feature.

LinuxOnTheDesktop:
Can someone say please whether any of this related to issue #1097?

genes1122:
They are the same.

dubrowgn:
Is there any chance this scaling value is exposed to lua somehow? While this seems like the right direction long term, drawing with cairo becomes problematic without knowing what scaling value conky is using internally.

ferdnyc:
@dubrowgn 

> Is there any chance this scaling value is exposed to lua somehow? While this seems like the right direction long term, drawing with cairo becomes problematic without knowing what scaling value conky is using internally.

I don't think it has to, though. Cairo has an API to handle that. The problem is that the API isn't exposed to the Lua bindings.

Since Cairo 1.14, two functions [`cairo_surface_set_device_scale()`](https://www.cairographics.org/manual/cairo-cairo-surface-t.html#cairo-surface-set-device-scale) and [`cairo_surface_get_device_scale()`](https://www.cairographics.org/manual/cairo-cairo-surface-t.html#cairo-surface-get-device-scale) have been available to control the internal scaling of the Cairo surface being drawn on.

Problem is, if I'm understanding this correctly those functions are missing from Conky's (very old) [cairo.pkg](https://github.com/brndnmtthws/conky/blob/main/lua/cairo.pkg) definition for the Cairo Lua bindings. So there's no way to query or adjust Cairo's notion of scaling using the exposed API. But the functionality to do so is present in Cairo.

If the Cairo surface were adjusted to scale the same way Xft is (and Conky is now doing internally), everything should go back to matching up without having to be manually scaled.

dubrowgn:
For anyone else struggling with this, my current work around is to measure the actual size of something I control, compared to what I expected. For example, in my config, I set window width to a constant, and then use the following to get a ratio:

```lua
local dpr = (conky_window.text_width - 1) / expected_width
cairo_scale(cairo, dpr, dpr)
```

PR #1086, opened 2021-04-17
Add ClassName hint to own window even when the Override_redirect is true

zalohasa:
Currently, when the own_window is true and override is the type, there is no WM_CLASS assigned to the window. This causes some problems with composers, when you want to apply a special treatment for the conky window. 

In my case, I use i3wm with picom as the composer. In order for conky to appear, I need to set the own_window to true, but this is not enough, as the window will be always on top of the others, so I also need to enable the override_redirect, assigning 'override' to the own_window_type. This works with the window manager, but then, I cannot write any rules to disable rounding corners, shadows, or blurring the background for the conky window, as there is no class for that window.

The change I propose is to add the own_window_class to the own window, even when override_redirect is true, so the composers could find the window and apply the user rules to them. 

I've tested the change empirically, by writing picom rules, and also by the use of xprop, to verify the conky window has the right class assigned to it.

### DIFF
--- a/src/x11.cc
+++ b/src/x11.cc
@@ -673,6 +673,18 @@ static void init_window(lua::state &l __attribute__((unused)), bool own) {
     /* Sanity check to avoid making an invalid 0x0 window */
     if (b == 0) { b = 1; }
 
+    XClassHint classHint;
+
+    // class_name must be a named local variable, so that c_str() remains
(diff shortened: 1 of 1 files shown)


brndnmtthws (approved):

zalohasa:
Thanks!

PR #1089, opened 2021-04-20
Added Intel backlight support

rogierreerink:
I added support for displaying Intel backlight information from sysfs. Such a function does not yet seem to exist in Conky and although it shouldn't be too difficult to write a Lua script for it, I think a native implementation is just a bit nicer. Unfortunately, it only works with Intel backlight because I don't know how other backlight drivers use sysfs.

As far as I can see this PR shouldn't affect other behaviour in Conky.

Steps that were taken to find errors:
1. I compiled and tested the code with and without valid sysfs paths to confirm that a failed `fopen` would be handled correctly.
2. I ran Intel backlight from 0 to _full_ to check that no zero-division was done and that the whole range was covered well.
3. I configured conky.conf with 0 to multiple instances of `intel_backlight` to check that either nothing would happen (in case of 0) or that each instance would open, read and close the backlight files well (using additional log messages, which I removed after confirmation.)

Peek 2021-04-20 23-16

This is my first commit here, so I hope that it's all in good shape. Please, let me know when I've missed something!

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -24,7 +24,7 @@ affinity <affy at users dot sourceforge dot net>
 akash <akash at users dot sourceforge dot net>
   battery_percent and battery_bar
 
-Asbj�rn Zweidorff Kj�r <bunjiboys at users dot sourceforge dot net>
+Asbjørn Zweidorff Kjær <bunjiboys at users dot sourceforge dot net>
   support for EVE-Online skill monitoring
 
 Aseem Mohanty <amohanty at myrealbox dot com>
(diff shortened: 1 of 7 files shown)


AUTHORS
```diff
@@ -24,7 +24,7 @@ affinity <affy at users dot sourceforge dot net>
 akash <akash at users dot sourceforge dot net>
   battery_percent and battery_bar
 
-Asbj�rn Zweidorff Kj�r <bunjiboys at users dot sourceforge dot net>
+Asbjørn Zweidorff Kjær <bunjiboys at users dot sourceforge dot net>
```
brndnmtthws, line 27: Revert all these changes.

brndnmtthws (changes requested):
Looks good minus the changes to AUTHORS. There are a couple formatting issues but I'll clean that up after.

brndnmtthws:
One more thing I forgot: can you update https://github.com/brndnmtthws/conky/blob/main/doc/config_settings.xml with details on this?

AUTHORS
rogierreerink: Thanks!

rogierreerink: Thanks! I've reverted those changes in my latest commit.

rogierreerink: Thanks! I've reverted them.

brndnmtthws (approved):

PR #1092, opened 2021-04-22
Add a `legacyfree` variable for `free` compat.

### LINKED ISSUE #1090: regression: conky 1.12.0 intentionally broke Linux memory reporting
**Issue**

As of Conky 1.12.0 intentionally broke Linux memory reporting

**Information**

Memory reporting in Conky under Linux was finally fixed with issue #859 in git commit https://github.com/brndnmtthws/conky/commit/11695b076abf72d797879c7c3b16fe89f3b68ca9 , after many (7) years of doing it incorrectly and so distributions dropped their patches fixing it themselves.

However then for largely unknown reasons Conky later intentionally broke memory reporting again under Linux in Feb 2021.
See git commits https://github.com/brndnmtthws/conky/commit/91a3f2ed871b7200fcf7a53f284bcd98f1fc49aa and https://github.com/brndnmtthws/conky/commit/c2c9e07afc7ac27e348ce12879f3b1e1b7382ce6 .

As is explained in detail in this kernel [git commit](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=34e431b0ae398fc54ea69ff85ec700722c9da773)  the proper way to report used memory is:

`curmem = info.memmax - memavail`

which was how it was being done in Conky before being reverted.

Sidenote: curmem is used as 'mem' in Conky which is defined as "Amount of memory in use." which fits this definition exactly.

The only explanation in the Conly source for why it was broken again is:

_**"* People think they will find the same amount of used RAM with the free program."**_

So Conky was doing it correctly then intentionally broke it supposedly because some people don't know how to read 'free' output correctly, yet didn't mention an opened issue by this group of people who don't know how memory actually works under Linux.

However nearly every other program, besides free, have already fixed memory reporting in a released version or in their next release (for example: htop, inxi, neofetch, screenfetch, etc) to match how it should be done with modern Linux.

In addition to the above mentioned fact that conky is now reporting the data incorrectly it also actually makes a **HUGE** difference as the numbers themselves are nowhere even close to each other.

Currently Conky is reporting my system is only using 8.16 GiB of ram while every other memory reporting program shows the real usage is nearly double that at 15.83 GiB.

**In addition Conky currently has no way to actually report the correct memory usage at all.**

**Perhaps the best solution to this problem is to have 'mem' report the correct memory usage, like prior to 1.12.0, and add a separate 'memused' variable to show the incorrect numbers like 'free' does if users still want to see the old bad data. Or maybe just add in the 'mem' variable documentation that 'free' reports used memory incorrectly.**

brndnmtthws:
It seems there is some confusion between the way Conky reports free
memory, versus other programs. I'm adding a new variable, `legacyfree`,
to provide the same value that is reported on older programs that don't
report memory the way recommended by newer kernels.

For more details, see:
 * https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=34e431b0ae398fc54ea69ff85ec700722c9da773
 * https://github.com/brndnmtthws/conky/pull/859
 * https://github.com/brndnmtthws/conky/pull/1028
 * https://github.com/brndnmtthws/conky/issues/1090

This resolves #1090.

### DIFF
--- a/doc/variables.xml
+++ b/doc/variables.xml
@@ -2314,6 +2314,15 @@
     <varlistentry>
         <term>
             <command>
+                <option>legacymem</option>
+            </command>
+        </term>
+        <listitem>Amount of memory used, calculated the same way as in the `free` program.
+        <para /></listitem>
+    </varlistentry>
+    <varlistentry>
+        <term>
+            <command>
                 <option>memgauge</option>
             </command>
             <option>(height),(width)</option>
--- a/src/common.cc
+++ b/src/common.cc
@@ -395,6 +395,7 @@ double cpu_barval(struct text_object *obj) {
 PRINT_HR_GENERATOR(mem)
 PRINT_HR_GENERATOR(memwithbuffers)
 PRINT_HR_GENERATOR(memeasyfree)
+PRINT_HR_GENERATOR(legacymem)
 PRINT_HR_GENERATOR(memfree)
 PRINT_HR_GENERATOR(memmax)
 PRINT_HR_GENERATOR(memdirty)
--- a/src/common.h
+++ b/src/common.h
@@ -105,6 +105,7 @@ double cpu_barval(struct text_object *);
 void print_mem(struct text_object *, char *, unsigned int);
 void print_memwithbuffers(struct text_object *, char *, unsigned int);
 void print_memeasyfree(struct text_object *, char *, unsigned int);
+void print_legacymem(struct text_object *, char *, unsigned int);
 void print_memfree(struct text_object *, char *, unsigned int);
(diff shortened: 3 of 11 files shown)


ohnonot:
Thank you for explaining and addressing this issue!
Shouldn't other mem-related variables also have a "legacy" version?
Like many people I use something like this:

~~~
MEM: ${mem}/${memmax} - ${memperc}%
~~~

Assuming that `memmax` is the same in both scenarios, there should at least by a `legacymemperc`, too?
Essentially what @Tomunu said [here](https://github.com/brndnmtthws/conky/pull/859#issuecomment-507476227).

PR #1093, opened 2021-04-24
Fix exec blocking

brndnmtthws:


stevenxxiu:
If you undo that commit, then I think you can undo my https://github.com/brndnmtthws/conky/commit/415c730133476d6652f479242b0638496b5f673b too as it's no longer relevant.

PR #1104, opened 2021-05-17
make `gw_iface` work with net stat

stevenxxiu:
**Descriptions**

Currently `gw_iface` doesn't work with the network speed (unless it's in a Lua function like `${lua rjust ${downspeed ${gw_iface}} 8}`), and doesn't wok not the network graph.

```
conky.text = [[
${gw_iface}\
${downspeed ${gw_iface}}\
${upspeed ${gw_iface}}
${downspeedgraph ${gw_iface} 60,230 7500000}${alignr}\
${upspeedgraph ${gw_iface} 60,230 3000000}
]]
```

This change lets it work with both:

The `downspeed` portion of the code appears to check for this, but it's buggy as it doesn't run `update_gateway_info()`.

Screenshot_2021-05-17_20-08-10

I'm not sure why this broke, as I remember it did work once. This could be related to [Partially revert change from c352069b31b2dbb4d. · brndnmtthws/conky@da2ddb0](https://github.com/brndnmtthws/conky/commit/da2ddb00ed05a772003f4d4ae5fc1156099dc66f), I'm not sure.

### DIFF
--- a/src/net_stat.cc
+++ b/src/net_stat.cc
@@ -134,6 +134,9 @@ void parse_net_stat_arg(struct text_object *obj, const char *arg,
   if (arg == nullptr) { arg = DEFAULTNETDEV; }
 
   if (0 == (strcmp("$gw_iface", arg)) || 0 == (strcmp("${gw_iface}", arg))) {
+#if defined(__linux__)
+    update_gateway_info();
+#endif
     arg = e_iface;
   }
 
--- a/src/net_stat.cc
+++ b/src/net_stat.cc
@@ -335,6 +335,14 @@ void parse_net_stat_graph_arg(struct text_object *obj, const char *arg,
 
   // default to DEFAULTNETDEV
   if (buf != nullptr) {
+#if defined(__linux__)
+    if (0 == (strcmp("$gw_iface", buf)) || 0 == (strcmp("${gw_iface}", buf))) {
+      update_gateway_info();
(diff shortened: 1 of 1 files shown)


src/net_stat.cc
```diff
@@ -133,9 +132,12 @@ void parse_net_stat_arg(struct text_object *obj, const char *arg,
 
   if (arg == nullptr) { arg = DEFAULTNETDEV; }
 
+#if defined(__linux__)
```
brndnmtthws, line 135: Move this `#if` so that it _only_ wraps the call to `update_gateway_info()`.

src/net_stat.cc
```diff
@@ -53,7 +53,6 @@
 #if defined(__linux__)
 #include "linux.h"
 #else
-static char e_iface[50] = "empty";
```
brndnmtthws, line 56: Why is this removed? I still see it referenced below.

brndnmtthws (changes requested):

src/net_stat.cc
stevenxxiu: Ah must be because I didn't put the `#if` above in the right place.

src/net_stat.cc
stevenxxiu: Fixed now.

stevenxxiu:
@brndnmtthws Mind looking at this again?

PR #1124, opened 2021-08-05
Fix SEGFAULT caused by out-of-bounds write

arkamar:
This PR fixes `SEGFAULT` caused by out-of-bounds write which I was experiencing after update to newest version, see the commit message for broader explanation.

I also replaced `graph_id` variable with `g->id` in other commit.


brndnmtthws:
Thanks and sorry for the delay!

arkamar:
No problem, thanks for merge!

PR #1135, opened 2021-08-22
avoid multiple running instances

movinator:
**Avoid multiple running instances**
* I like to use saved session on kde and it happens, that on startup multiple instances of conky will be created. I don't know, whether this behaviour is related to my number of virtual desktops or whatever, so I changed conky to not start, if it is already running.
* Standard behaviour is not changed. Changed app just refuses to start if there is an instance already running
* Tested: from terminal start one instance with "conky -d" and then start conky without -d switch. conky will respond, that it is already running and quits.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -260,6 +260,27 @@ static void print_help(const char *prog_name) {
          prog_name);
 }
 
+int already_running() {
+  int rv = 0;
+  char buf[128];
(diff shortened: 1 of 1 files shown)


Nimesh-Srivastava (approved):

simotek:
Yeah for now I also run 7-8 instances each with different configs, I don't think many people would want to run 2 instances with identical configs so maybe thats an approach you could look at.

PR #1148, opened 2021-10-04
Fix diskio graph scale being 3 orders of magnitude too small

missing-semi-colon:
The scale shown in diskiograph with `show_graph_scale = true` was 3 orders of magnitude too small (it shows "B" instead of "KiB" etc)

Before:
diskio_error

After:
diskio_correct


missing-semi-colon:
Made a mistake with the PR. See https://github.com/brndnmtthws/conky/pull/1157 for the new one.

PR #1166, opened 2022-01-10
fix #1165 by applying the padding function to nvidia objects

### LINKED ISSUE #1165: nvidia percentages don't get a padding with config `pad_percents`
**Issue**

nvidia print callbacks are registered as print callbacks `obj->callbacks.print = &print_nvidia_value;` (line 1869, src/core.cc).

And in `src/conky.cc` (line 722)
```
    if (obj->callbacks.print != nullptr) {
      (*obj->callbacks.print)(obj, p, p_max_size);
```
the callbacks are called.

For percentages to be printed correctly we would need to call (line 722, src/conky.cc), using `callbacks.percentage`.
```
    } else if (obj->callbacks.percentage != nullptr) {
      percent_print(p, p_max_size, (*obj->callbacks.percentage)(obj));
    }
```

**Information**

Source version at the time of writing `8f5014b9`.

kistlin:
**Descriptions**
PR tries to fix #1165.

A field `is_percentage` was added to the class `nvidia_s`. By default it is `false`. If set to `true` in `set_nvidia_query`, during the call to `print_nvidia_value`, padding will be added. If left `false`, the current behavior is maintained.

Little test with
```
conky.config = {
  nvidia_display = ":0",  -- https://github.com/brndnmtthws/conky/issues/520
  out_to_console = true,
  out_to_x = false,
  pad_percents = 3,
  short_units = true,
  update_interval = 1,
  use_spacer = "left"
}

conky.text = [[
gpu util ${nvidia gpuutil}
]]
```

resulted in
```
> gpu util  10
```
with `use_spacer = "right"` (note the trailing whitespace)
```
> gpu util 13 
```
without `pad_percents = 3`
```
gpu util 10
```

### DIFF
--- a/src/nvidia.cc
+++ b/src/nvidia.cc
@@ -327,7 +327,8 @@ class nvidia_s {
         attribute(ATTR_GPU_TEMP),
         token(0),
         search(SEARCH_FIRST),
-        target_id(0) {}
+        target_id(0),
+        is_percentage(false) {}
   const char *command;
   const char *arg;
   QUERY_ID query;
@@ -337,6 +338,7 @@ class nvidia_s {
   SEARCH_ID search;
   //  added new field for GPU id
   int target_id;
+  bool is_percentage;
 };
 
 // Cache by value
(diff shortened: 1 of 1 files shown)


Fizie92 (approved):
gpu util 10

PR #1169, opened 2022-01-30
Detect that nvidia display exists in each refresh

rlalik:
**Descriptions**

I use laptop with dual graphics. The nvidia is used on demand with bumbleebee. Currently if the optirun as not used, the conky will abort due to missing nvidia display (typically :8) or if started when optirun running (nvidia display present) then will crash after the optirun finish due to invalid pointer.

In my change the `XOpenDisplay` is call each time when the nvidia desktop is requested. As I have no experience with X server I don't know what impact it might have on the performance.

This change should not affect current behaviour in any way.

I tested it with my laptop trying various orders of starting conky, optirun and stoping optirun to find any issues. None found.

This contribution is GPLv3 licensed.

### DIFF
--- a/src/nvidia.cc
+++ b/src/nvidia.cc
@@ -101,6 +101,8 @@
 #include "temphelper.h"
 #include "x11.h"
 
+#include <memory>
+
 // Separators for nvidia string parsing
 // (sample: "perf=0, nvclock=324, nvclockmin=324, nvclockmax=324 ; perf=1,
 // nvclock=549, nvclockmin=549, nvclockmax=549")
@@ -366,6 +368,14 @@ struct nvidia_c_string {
 const int MAXNUMGPU = 64;
 
 namespace {
+
+// Deleter for nv display to use with std::unique_ptr
+void close_nvdisplay(Display * dp) {
+  XCloseDisplay(dp);
+}
+
+using unique_display_t = std::unique_ptr<Display, decltype(&close_nvdisplay)>;
+
 class nvidia_display_setting
     : public conky::simple_config_setting<std::string> {
   typedef conky::simple_config_setting<std::string> Base;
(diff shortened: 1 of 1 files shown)


src/nvidia.cc
```diff
@@ -376,30 +374,40 @@ class nvidia_display_setting
...
-    }
-  }
+  nvdisplay = do_convert(l, -1).first;
 
   ++s;
 }  // namespace
 
+Display *nvidia_display_setting::get_nvdisplay() {
+  if (!nvdisplay.empty()) {
+    Display *nvd = XOpenDisplay(nvdisplay.c_str());
```
brndnmtthws, line 396: I'm not seeing the corresponding calls to `XCloseDisplay` to release the resources from `XOpenDisplay`.

Alternatively, you could wrap the pointer in a small struct or class and use RAII to call `XCloseDisplay` in the destructor.

brndnmtthws (changes requested):

src/nvidia.cc
rlalik: Yeah, it was bad I forgot about releasing (I made incorrect assumptions about the XDisplay without checking specs). I updated the code with smart pointer to handle this now.

src/nvidia.cc
```diff
@@ -366,6 +368,14 @@ struct nvidia_c_string {
 const int MAXNUMGPU = 64;
 
 namespace {
+
+// Deleter for nv display to use with std::unique_ptr
+void close_nvdisplay(Display * dp) {
+  XCloseDisplay(dp);
+}
+
+using unique_display_t = std::unique_ptr<Display, decltype(&close_nvdisplay)>;
```
brndnmtthws, line 377: This is a great way to handle this!

src/nvidia.cc
```diff
@@ -366,6 +368,14 @@ struct nvidia_c_string {
 const int MAXNUMGPU = 64;
 
 namespace {
+
+// Deleter for nv display to use with std::unique_ptr
+void close_nvdisplay(Display * dp) {
```
brndnmtthws, line 373: 👍

brndnmtthws (approved):
Looks great, thanks!

PR #1186, opened 2022-05-02
added ', ' delimiter to user_names

smjoseph:
**Descriptions**
$user_names was not adding any kind of delimiter between names, so it is ending up as 'useruseruseruser' when there is more than one user. I added a comma and space.


src/users.cc
```diff
@@ -42,6 +42,7 @@ static void user_name(char *ptr) {
     if (usr->ut_type == USER_PROCESS) {
       if (strlen(ptr) + strnlen(usr->ut_name, UT_NAMESIZE - 1) + 1 <= BUFLEN) {
         strncat(ptr, usr->ut_name, UT_NAMESIZE);
+        strncat(ptr, ", ", 2);
```
brndnmtthws, line 45: You need to subtract the 2 extra characters from `UT_NAMESIZE` so this doesn't overflow.

PR #1204, opened 2022-09-04
Use another port for HTTP, one not blocked by Firefox and Chrome

### LINKED ISSUE #1061: Firefox bans "unsafe" ports by default
Firefox started to restrict which ports it can connect to. A Firefox user trying to connect to the Conky by HTTP on the default port would get this message:

> This address is restricted
> This address uses a network port which is normally used for purposes other than Web browsing. Firefox has cancelled the request for your protection.

With the only option to "Try Again" which wouldn't work.

[The workaround is to create](https://support.mozilla.org/en-US/questions/1083282) a `network.security.ports.banned.override` key in `about:config` to list the ports to allow.

We should at least mention this in the documentation.

mmuman:
Fixes #1061.

### DIFF
--- a/cmake/ConkyBuildOptions.cmake
+++ b/cmake/ConkyBuildOptions.cmake
@@ -246,7 +246,7 @@ option(BUILD_IRC "Enable if you want IRC support" false)
 
 option(BUILD_HTTP "Enable if you want HTTP support" false)
 if(BUILD_HTTP)
-  set(HTTPPORT "10080" CACHE STRING "Port to use for out_to_http")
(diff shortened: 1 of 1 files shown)


brndnmtthws:
I think we should really make this configurable, instead of changing the default value (which will break this feature for anyone who already uses it). Would you like to take a stab at that?

mmuman:
Hmm since it touches code I moved around in the X11 PR I'd rather wait on this.
I suppose we could add a setting in the default config files for a while before changing the default value.

mmuman:
We cal probably close this one.

PR #1205, opened 2022-09-04
Http port config

mmuman:
This allows setting the port to listen to for HTTP, to work around #1061. Even seems to update correctly on reload.

Depends on #664 as I didn't want to rebase it again over this.


npyl (commented):
looks good

npyl (approved):

PR #1220, opened 2022-09-20
Avoid referencing underfined symbol

madpilot78:
When building without ncurses out_to_ncurses is not defined. It is also not needed to check for ncurses if it is not compiled in.

The error I'm getting when building without ncurses is:

```
--- src/CMakeFiles/conky.dir/display-console.cc.o ---
/wrkdirs/usr/ports/sysutils/conky/work/conky-1.13.1/src/display-console.cc:58:8: error: use of undeclared identifier 'out_to_ncurses'
      !out_to_ncurses.get(*state)) {
       ^
1 error generated.
*** [src/CMakeFiles/conky.dir/display-console.cc.o] Error code 1
```

### DIFF
--- a/src/display-console.cc
+++ b/src/display-console.cc
@@ -54,8 +54,11 @@ display_output_console::display_output_console(const std::string &name_)
 }
 
 bool display_output_console::detect() {
-  if ((out_to_stdout.get(*state) || out_to_stderr.get(*state)) &&
-      !out_to_ncurses.get(*state)) {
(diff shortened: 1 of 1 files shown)


brndnmtthws (approved):

mmuman:
Thanks, I didn't try all combinations :sweat_smile:

madpilot78:
> Thanks, I didn't try all combinations sweat_smile

I did, but just because I was updating the FreeBSD port.

PR #1223, opened 2022-09-24
Linux: show the CPU scaling governor with $cpugovernor

pghvlaans:
**Descriptions**
* For Linux, display the CPU scaling governor for any thread with a variable `$cpugovernor`. This will help users who change their scaling governors on the fly to keep track of what they're doing.
* No existing behavior is changed.
* I tested `$cpugovernor` with various combinations of CPU scheduling governors, as well as the default value (`cpu0`) and invalid values (which revert to the default).
* Screenshot: 
cpugovernors

**Licenses**
* There are no new files.


su8:
`atoi` can introduce undefined behaviour to conky, replace it with `strtol`

pghvlaans:
It works with `strtol`. Everything else in `core.cc` uses `atoi`, though.

brndnmtthws:
> It works with `strtol`. Everything else in `core.cc` uses `atoi`, though.

I understand there's a lot of existing crufty bits in the code, but if you're adding new code it's preferred to use C functions that (at a minimum) reduce the possibility of bugs such as reading past the end of buffers.

pghvlaans:
OK, understood! I'll go ahead and push.

brndnmtthws (approved):
This is great, thank you kindly.

PR #1235, opened 2022-10-04
use HCL color space (Rec.2020 for luma) instead of HSV

caiofreitaso:
## Description

HSV Gradient brought some unpleasant colors (either too bright or too saturated) with my conky config because of how the HSV color space works. I changed it to HCL (Hue-Chroma-Luma) so that colors blend with a better distribution of luminosity.

Using HCL color space:
image

Using HSV color space:
image


brndnmtthws:
First of all, thanks for doing this.

Is there a reason why you removed the HSV code? I think it would make sense to add HCL, which does indeed look better to me, and keep HSV so it doesn't break existing configs. It seems like it should be fairly straight forward to have both HSV and HCL. What do you think?

brndnmtthws (changes requested):
This looks good, thanks for fixing HSV gradient.

Now we have 3 different ways to colour the graphs, so we should probably add a config setting and document the available methods. Something like `graph_griadent_method = (rgb|hsv|hcl)`. Do you want to take a stab at that?

Also, please unstage `tests/catch2/catch.hpp`, that's a 3rd party library and I'd like to keep the original.

caiofreitaso:
That seems nice. I’ll try 😊

Can I fix those errors I mentioned on HSV or should I leave it as it is?

And I will unstage the header! Sorry!

brndnmtthws:
> Can I fix those errors I mentioned on HSV or should I leave it as it is?

Yeah, that would be great!

caiofreitaso:
Done! Now the config file can declare which mode with `graph_gradient_mode = 'hcl'`, for example.

RGB:
image

HSL:
image

HCL:
image

doc/config_settings.yaml
```diff
@@ -433,6 +433,10 @@ values:
     desc: Shows the time range covered by a graph.
   - name: show_graph_scale
     desc: Shows the maximum value in scaled graphs.
+  - name: graph_gradient_mode
+    desc: |-
+      Changes the color space used for interpolation. Arguments are hcl, hsv,
+      and rgb (default).
```
brndnmtthws, line 439: Can you add `args` here, like:

```yaml
    args:
      - (rbg|hcl|hsv)
```

src/CMakeLists.txt
```diff
@@ -68,6 +68,8 @@ set(conky_sources
     exec.h
     fs.cc
     fs.h
+    gradient.cc
```
brndnmtthws, line 71: Since we're including the sources like this, we can remove all references to `BUILD_HSV_GRADIENT` from the CMake files and source. The build option is defined in [ConkyBuildOptions.cmake](https://github.com/brndnmtthws/conky/blob/d2f81cee7ee0bedb4b15b306709216358665e7d0/cmake/ConkyBuildOptions.cmake#L270).

brndnmtthws (changes requested):
Overall this looks much better. Thanks for combining all the code into one file, I think that's cleaner.

It looks like the tests were removed, I'm not sure if that was intentional or not, but we keep & update the tests so they work with this code. I don't generally require people include tests with submissions, but we shouldn't be removing tests.

cmake/ConkyBuildOptions.cmake
```diff
@@ -267,8 +267,6 @@ option(BUILD_PULSEAUDIO
 option(BUILD_INTEL_BACKLIGHT
        "Enable support for Intel backlight" false)
 
-option(BUILD_HSV_GRADIENT "Enable gradient in HSV colour space" true)
```
brndnmtthws, line 270: Thanks!

caiofreitaso:
Hey, could you explain what is the `ON` and `OFF` that appear on those builds? I want to reproduce the errors here, but I can't understand what exactly those are. Is it a no GUI or no X11 flag? D:

brndnmtthws:
Good question, in this case `OFF` is referring to the `BUILD_X11` flag. I will update the build so it prints the build command.

The full command should be as follows:

```shell
           cmake ..                               \
             -DBUILD_AUDACIOUS=ON                 \
             -DBUILD_HTTP=ON                      \
             -DBUILD_ICAL=ON                      \
             -DBUILD_ICONV=ON                     \
             -DBUILD_IRC=ON                       \
             -DBUILD_IRC=ON                       \
             -DBUILD_JOURNAL=ON                   \
             -DBUILD_LUA_CAIRO=ON                 \
             -DBUILD_LUA_IMLIB2=ON                \
             -DBUILD_LUA_RSVG=ON                  \
             -DBUILD_MYSQL=ON                     \
             -DBUILD_NVIDIA=ON                    \
             -DBUILD_PULSEAUDIO=ON                \
             -DBUILD_RSS=ON                       \
             -DBUILD_TESTS=ON                     \
             -DBUILD_WLAN=ON                      \
             -DBUILD_X11=OFF                      \
             -DBUILD_XMMS2=ON                     \
             -DCMAKE_C_COMPILER=$CC               \
             -DCMAKE_CXX_COMPILER=$CXX            \
             -DMAINTAINER_MODE=ON
```

brndnmtthws:
Oh I guess you can also expand the substituted commands if you click the `Run ...` log line at the top of the step.

brndnmtthws:
Are you still working on this PR? I think we're very close to completing it.

caiofreitaso:
> Are you still working on this PR? I think we're very close to completing it.

I was. Sorry for the delay! D:

src/gradient.cc
```diff
@@ -0,0 +1,335 @@
...
+  }
+
+  for (int i = 0; i < 3; i++) {
+    mask[i] = mask[i] << shift[i];
+  }
+}
+
+void gradient_factory::setup_colour_depth() {
+#ifdef BUILD_X11
+  if (state == nullptr) { // testing purposes
```
brndnmtthws, line 84: This is okay for now, but it would be better to initialize the state with appropriate values as part of the setup for the tests.

PR #1241, opened 2022-10-08
power draw added to conky

brightkill:
Add a new option $battery_power_draw to see battery power draw in watts

### DIFF
--- a/doc/variables.yaml
+++ b/doc/variables.yaml
@@ -175,6 +175,13 @@ values:
     default: BAT0
     args:
       - (num)
+  - name: battery_power_draw
+    desc: |-
+      Battery power draw in watts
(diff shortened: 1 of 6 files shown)


brndnmtthws (approved):
I don't love C string functions, but otherwise looks good to me.

:shipit:

brndnmtthws:
Looks like one small build issue on macOS, once that's addressed we should be good to go. You can just copy one of the stubs like this if you want to punt on implementing it: https://github.com/brndnmtthws/conky/blob/30cad6afcaa97bad885ab4cfceb2534b164f98c3/src/darwin.mm#L1034-L1037

brightkill:
done

PR #1243, opened 2022-10-12
Fix man page formatting.

pghvlaans:
## Description
* The new markdown-based man page was being rendered without the `.TH` metadata by `pandoc-2.19.2`, which resulted in line breaks and other formatting not appearing in the final product. To fix this, add `%` statements with the original title and date and run `pandoc` with options `--standalone -f markdown -t man man.md`. The instructions in `README.md` have also been updated to that effect.
* Only the man page is affected.
* I rebuilt `conky` and verified the formatting of the man page.
* Here are screenshots of the current and fixed man pages:
man-page-before
man-page-after

### DIFF
--- a/doc/CMakeLists.txt
+++ b/doc/CMakeLists.txt
@@ -27,7 +27,7 @@ if(BUILD_DOCS)
     )
     add_custom_target(conky.1
         ALL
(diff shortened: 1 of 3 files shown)


brndnmtthws:
Very lovely, thanks for catching this!

PR #1248, opened 2022-10-12
Use cmake macros for cxx standards and compiler/linker options

rlalik:
In the current cmake files, compiler flags are always appended to itself, leading to multiplication of the same flags in each cmake run, giving flags like `-std=c++17 -std=c++17 -std=c++17 -std=c++17 ...`.

This fix replaces this with ore appropriate tools and macros from cmake.

This change should not influence current builds. Tested with existing tests and github actions.


brndnmtthws:
Makes sense. These settings have changed a bit since that code was originally written.

PR #1250, opened 2022-10-13
More X11 cleanup

mmuman:
I needed those to get something displayed with SDL. There are still some more around but I wasn't sure how to deal with them.

Maybe the NULL test is not always required, but I haven't tried removing them yet. Maybe we should introduce some helper function instead to lower the verbosity?

### DIFF
--- a/src/conky.cc
+++ b/src/conky.cc
@@ -772,7 +772,6 @@ static int get_string_width_special(char *s, int special_index) {
   if (display_output() == nullptr || !display_output()->graphical()) {
     return strlen(s);
   }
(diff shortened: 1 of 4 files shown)


mmuman:
We can probably resolve the rest later.

brndnmtthws (approved):
Looks good, nice work 🙂

bi4k8:
Is there anything holding this up from being merged? I'd like to move forward with #1310 as soon as possible.

brndnmtthws:
I'm going to assume this is ready to go.

PR #1251, opened 2022-10-13
SDL GUI backend

mmuman:
The idea is to be able to run Conky on ancient OSes and hardware (thinking about MiNT (not the GNU/Linux distro, the Atari OS) or AmigaOS), in a regular window but with minimal coding required, and also to allow testing the GUI backend without X11. I used SDL 1.2 because SDL2 only runs on Windows, OSX and Linux, which limits the interest for old platforms.

For now it only draws text with a hardcoded font. It also still needs to be compiled with X11 due to some variables here and there.


mmuman:
I added some way to locate and load TTF files by the font name, it kinda works.
Screenshot_20221129_223721

mmuman:
Just rebased, had to change some `#ifdef` info `#if defined(BUILD_WAYLAND) || defined(BUILD_SDL)` so maybe we'd need some `BUILD_NONX11` define around to avoid these lists everywhere?

brndnmtthws:
> Just rebased, had to change some `#ifdef` info `#if defined(BUILD_WAYLAND) || defined(BUILD_SDL)` so maybe we'd need some `BUILD_NONX11` define around to avoid these lists everywhere?

That could work, though it seems messy. There's also `BUILD_GUI`, but it sounds like that won't help here.

mmuman:
The recent changes on color stuff just got rid of the `#ifdef` thing.

mmuman:
And we have colors :-)

Caellian:
Hi, I changed how registration of output displays works to avoid linker magic, you can check out how other backends are registered on `main` branch for reference.

Main differences are:
- `sdl-output.cc` should be guarded by build flag in `CMakeLists.txt`
- You need to provide a log message template fallback in `display-output.cc` for feature gated outputs

mmuman:
Rebased, updated to last renaming, and should be up to @Caellian 's changes too.
Still not finished.

mmuman:
(of course, pushing from wrong machine…)

mmuman:
I'd like to have the default settings working but it seems it's black over black so nothing shows up… Otherwise the basic functions are there.

brndnmtthws:
> I'd like to have the default settings working but it seems it's black over black so nothing shows up… Otherwise the basic functions are there. Should we merge it in already?

Black over black? The default colour should be grey & white, for text at least.

https://github.com/brndnmtthws/conky/blob/612cecfecc30deca6ff9a7e0ac9d21bb634e616d/data/conky.conf#L58-L77

brndnmtthws:
By the way, if you think it's ready to merge, click the "Ready for review" button so others know it's time to review it.

mmuman:
> Black over black? The default colour should be grey & white, for text at least.

I believe it's because only `out_to_x` is set in the default config, which doesn't exist if I don't build with it, so I report the output as available (I fake one locally to read it), but it's not used (`out_to_sdl` defaults to false), so nothing is drawn to any GUI output, and it doesn't even parse colors.

I think gui.cc should have a fallback `out_to_x` as a catchall `out_to_gui` or something.

mmuman:
Yeah that fixed it.

mmuman:
Default config now renders just fine.
Copie d'écran_20250126_032910

brndnmtthws:
I fixed the font loading on macOS, and added some basic font caching.

However, I get a blank window on macOS with the default config:

Screenshot 2025-01-26 at 07 48 48

brndnmtthws:
Not sure why the Nix build is failing, but it's fine to ignore it for now.

mmuman:
Hmm probably something similar with how we accept GUI settings or not.

Btw, it fails to build on Haiku because `::map` is not in `<unordered_map>`.

mmuman:
Just to make sure, you might want to trace calls to `set_foreground_color` to see if it actually gets proper colors or just black.

brndnmtthws:
Is there any reason we can't use SDL2 rather than 1.2? It seems like SDL 1.2 isn't as well supported by the distros.

Ubuntu doesn't appear to have SDL_ttf 1.2: https://packages.ubuntu.com/search?suite=noble&arch=any&searchon=names&keywords=libsdl-ttf

brndnmtthws:
FWIW there's a decent migration guide here: https://wiki.libsdl.org/SDL2/MigrationGuide

mmuman:
Yes, the idea is to run on very old OSes like Atari MiNT.

brndnmtthws:
> Yes, the idea is to run on very old OSes like Atari MiNT.

Ah, right, I see now that's in your original comment, but I had forgotten about it.

The SDL_ttf library is pretty tiny, I think we could just bundle it as a 3rd party source. It's literally just 2 source files (`SDL_ttf.c` and `SDL_ttf.h`). Then it can be used with the SDL 1.2 compat layer, which is widely available. What do you think of that?

brndnmtthws:
One more thought: maybe we should make it more clear that we intentionally only support SDL 1.2, so that in the future we can support newer versions as well.

brndnmtthws:
> The SDL_ttf library is pretty tiny, I think we could just bundle it as a 3rd party source. It's literally just 2 source files (`SDL_ttf.c` and `SDL_ttf.h`). Then it can be used with the SDL 1.2 compat layer, which is widely available. What do you think of that?

Actually, this isn't as easy as I thought. SDL_ttf 1.2 depends on freetype 1, which is also somewhat deprecated.

At this point, I think it's better to just not support distros that don't ship with SDL1.2, or people can compile SDL1.2 (and its dependencies) themselves if they really want it in a modern distro that doesn't include all the SDL1.2 deps.

mmuman:
I suppose we could support both with some `#ifdef` spaghetti.

brndnmtthws:
> I suppose we could support both with some `#ifdef` spaghetti.

This is less than ideal. It becomes really hard to maintain, and there are already a million different build options as it is.

I'd prefer to keep it simple, and since GitHub actions uses Ubuntu by default for the Linux CI, it would be good to make it compile on Ubuntu (and macOS too if possible). At least that way we can catch compiler errors.

Getting it to compile with Nix and macOS is pretty easy, but Ubuntu is a pain. Doing it in Docker would work, but I don't love that.

I guess we could use the [ExternalProject](https://cmake.org/cmake/help/latest/module/ExternalProject.html) feature of CMake, then it wouldn't be too hard to include _both_ freetype 1 and SDL_ttf 1.2 as 3rdparty dependencies. While I don't like including 3rdparty code this way, in this case it makes some sense.

mmuman:
Or just have a separate sdl2 display.

Caellian:
Just to check, did rebasing to renamed files cause issues or did git figure out how to move things properly? I tried really hard to reduce the impact so I'm wondering whether it made any difference.

> Or just have a separate sdl2 display.

This seems like the cleanest option from all angles (code, CMake, CI) - [`SDL_VERSION_MAJOR`](https://cmake.org/cmake/help/latest/module/FindSDL.html) from `FindSDL` can be used to switch between includes and sources so it would only add one variable and work seamlessly for both.

Last (somewhat insane) option is to port SDL_ttl 1.2 to use TrueType 2, it's supported ifaict. :smile:

mmuman:
> Just to check, did rebasing to renamed files cause issues or did git figure out how to move things properly? I tried really hard to reduce the impact so I'm wondering whether it made any difference.

Just had to `git mv` the SDL ones.

> This seems like the cleanest option from all angles (code, CMake, CI) - [`SDL_VERSION_MAJOR`](https://cmake.org/cmake/help/latest/module/FindSDL.html) from `FindSDL` can be used to switch between includes and sources so it would only add one variable and work seamlessly for both.

I'll look into that later.

PR #1343, opened 2022-12-25
Update doc/man/md.j2 about auto sensing config changes

ZjYwMj:
Partially fix bug #1301. 
Add to the man page that in many cases, if the user just waits a few seconds after a new configuration has been written to disk, conky is likely to sense its configuration has changed. And reload the new configuration automatically.

### DIFF
--- a/doc/variables.yaml
+++ b/doc/variables.yaml
@@ -58,10 +58,10 @@ values:
   - name: adt746xfan
     desc: Fan speed from therm_adt746x.
   - name: alignc
+    default: 0
     desc: |-
       Align text to centre, with a shift of N (num) pixels. A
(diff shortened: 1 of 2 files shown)


brndnmtthws (approved):

doc/variables.yaml
```diff
@@ -58,7 +58,10 @@ values:
   - name: adt746xfan
     desc: Fan speed from therm_adt746x.
   - name: alignc
-    desc: Align text to centre.
+    desc: |-
```
brndnmtthws, line 61:
```suggestion
    default: 0
    desc: |-
      Align text to centre, with a shift of N (num) pixels. A
      negative N (num) shifts to the right.
```

PR #1402, opened 2023-02-15
Add an above-all overlay mode for `own_window_type`, `utility`

tranquillity-codes:
# Checklist
- [ ] I have linked to any relevant GitHub issues, if applicable

## Description

Added the `utility` option for `own_window_type`, where conky functions as an overlay on top of all windows
No existing features were changed
Validated by checking whether the behaviour of other `own_window_type` modes changed, and whether the behaviour of the `utility` `own_window_type` behaves as expected
image


brndnmtthws (changes requested):
Looks good, but you need to add libxfixes to the macos build here: https://github.com/brndnmtthws/conky/blob/main/.github/workflows/build-and-test-macos.yaml#L22-L33

After that, I think we're good to go.

tranquillity-codes:
Now it seems to be failing at code that I did not touch

tranquillity-codes:
Seems that the build system does not work as I expected, and BUILD_XFIXES is not defined even on Linux. I have done some more testing, and discovered this fact, currently trying to figure out exactly what is going on.

tranquillity-codes:
I tracked the issue down to `#cmakedefine BUILD_XFIXES 1` in `cmake/config.h.in`, testing in a clean repo to be sure.

tranquillity-codes:
Okay, that should be fixed, tested again and it works.

brndnmtthws:
When I merge #1403, that should fix the mac build. I guess something changed recently on the github runners and pkg-config is no longer installed by default.

tranquillity-codes:
I see, thanks for the info & work on this ^^ Any more input necessary from me?

PR #1447, opened 2023-03-05
Fix and test graph parsing

### LINKED ISSUE #1428: [Bug]: downspeedgraph and upspeedgraph require color and width after update
### What happened?

After upgrading from 1.17.0 to 1.18.1 through the Arch Linux repos I noticed my `downspeedgraph` and `upspeedgraph` objects were throwing errors. Upon digging through documentation and some trial and error I found that I needed to now include the width of the graph and 2 color values.

In my config before the 2 lines were as follows:
`${downspeedgraph wlp7s0 80,}`
`${upspeedgraph wlp7s0 80,}`

But now they must be these 2 lines as follows to work like they used to:
`${downspeedgraph wlp7s0 80,224 #5498f8 #5498f8}`
`${upspeedgraph wlp7s0 80,224 #5498f8 #5498f8}`

Oddly enough passing only the size of the graph will produce the correct size, and passing only the interface will produce a working graph of said interface, but when both are passed to it I get the following output when running `conky` in my terminal:
```
conky: desktop window (1c0001d) is subwindow of root window (6c7)
conky: window type - desktop
conky: drawing to created window (0x5400002)
conky: can't parse X color 'wlp7s0' (6)
conky: can't parse X color '80,' (3)
conky: can't parse X color 'wlp7s0' (6)
conky: can't parse X color '80,' (3)
conky: forked to background, pid is 49288
```

Also note that when passing only the size `80,` will work, but when passing that as the size with the colors as arguments I still get the errors noted above. An example of such would look like:
`${downspeedgraph wlp7s0 80, #5498f8 #5498f8}`

Hopefully this is enough information to shed some light on the issue. If you need me to test anything else let me know and I will happily mess with my config.

The config I will attach below is my full config that worked in 1.17.0, but does not fully work now.

### Version

1.18.1

Arch Linux

### Conky config

```Lua
conky.config = {
background = true,
double_buffer = true,
alignment = 'top_right',
border_width = 1,
cpu_avg_samples = 2,
default_color = 'black',
default_outline_color = 'black',
default_shade_color = 'black',
default_bar_width = 200,
draw_borders = false,
draw_graph_borders = true,
draw_outline = false,
draw_shades = false,
gap_x = 8,
gap_y = 32,
minimum_width = 170,
minimum_height = 0,
net_avg_samples = 2,
no_buffers = true,
out_to_console = false,
out_to_stderr = false,
extra_newline = false,
own_window = true,
own_window_type = 'desktop',
own_window_transparent = false,
own_window_colour = '#000000',
own_window_argb_value = 180,
stippled_borders = 0,
update_interval = 2.0,
uppercase = false,
use_spacer = 'none',
show_graph_scale = false,
show_graph_range = false,
use_xft = true,
xftalpha = 0.1,
color0 = '#5498f8',
color1 = '#ee7c44',
color2 = '#db4238',
color3 = '#f5683a',
color4 = '#ef4447'
}

conky.text = [[
${voffset 1}$alignc${font DejaVu Sans:size=14}${color1}≻   ${color3}Status   ${color1}≺${font Ubuntu:size=12}
${voffset 12}${color2}OS $alignr ${color0} ${exec lsb_release -d | cut -f 2| tr "[:upper:]" "[:lower:]" | cut -d ' ' -f1-2}
${color2}Kernel $alignr ${color0} $kernel

${color2}CPU ${color0}${alignr}$cpu%
${cpubar cpu1 5,224}
${cpubar cpu2 5,224}
${cpubar cpu3 5,224}
${cpubar cpu4 5,224}
${cpubar cpu5 5,224}
${cpubar cpu6 5,224}
${cpubar cpu7 5,224}
${cpubar cpu8 5,224}
${top name 1} $alignr ${top cpu 1}%
${top name 2} $alignr ${top cpu 2}%
${top name 3} $alignr ${top cpu 3}%
${top name 4} $alignr ${top cpu 4}%
${top name 5} $alignr ${top cpu 5}%
${top name 6} $alignr ${top cpu 6}%

${color2}RAM ${color0}${alignr} $mem / $memmax - $memperc%
${membar 5,224}
${color2}SWAP ${color0}${alignr} $swap / $swapmax - $swapperc%
${swapbar 5,224}
${top_mem name 1} $alignr ${top_mem mem_res 1}
${top_mem name 2} $alignr ${top_mem mem_res 2}
${top_mem name 3} $alignr ${top_mem mem_res 3}
${top_mem name 4} $alignr ${top_mem mem_res 4}
${top_mem name 5} $alignr ${top_mem mem_res 5}
${top_mem name 6} $alignr ${top_mem mem_res 6}

${color2}Temperatures
${color0}Intel i7 4770 $alignr ${exec sensors | grep -E "^Package id 0:" | cut -d ' ' -f 5-5 | cut -d '+' -f 2-}
${color0}AMD RX 6600 $alignr ${exec /home/syphist/.programs/gputemp-conky.sh}

${color2}Download ${color0}${alignr}${downspeedf wlp7s0} KiB/s
${downspeedgraph wlp7s0 80,}
${color2}Upload ${color0}${alignr}${upspeedf wlp7s0} KiB/s 
${upspeedgraph wlp7s0 80,}

${color2}Local IP${color0}${alignr}${addrs wlp7s0}
${color2}Public IP${color0}${alignr}${execi 3600 wget -q -O /dev/stdout http://checkip.dyndns.org/ | cut -d : -f 2- | cut -d \< -f -1}
]]
```

### Stack trace

_No response_

### Relevant log output

_No response_

bi4k8:
Fixes #1428.

# Checklist

## Description

First I did a couple cleanups (adding comments and moving an enveloping `if` into an early-return) to `scan_graph` to make it easier to read and edit. Then I reorganized the cases to correct the misparsing that was introduced in 5e98c49c. But I still wasn't happy with how the code looked, so I wanted to simplify it further. I added a test case that exercises the parser on its own, then separated out parsing of a quoted command prefix from `scan_graph`, as it's a separate concern and not all callers even accept a command prefix (and some consider one optional, such as the net graphs). This cuts the number of cases that `scan_graph` has to handle in half.

I tested against a few conkyrc files from #1428 to verify that this PR corrects the regression, but more testing would be appreciated as it seems this impacted quite a few people (sorry!). This PR shouldn't change any behavior other than fixing the bug, but because it does touch all users of graphs, it would be good to give it some fairly thorough testing.

### DIFF
--- a/src/common.cc
+++ b/src/common.cc
@@ -352,10 +352,7 @@ void print_no_update(struct text_object *obj, char *p,
 
 #ifdef BUILD_GUI
 void scan_loadgraph_arg(struct text_object *obj, const char *arg) {
-  char *buf = nullptr;
-
-  buf = scan_graph(obj, arg, 0);
-  free_and_zero(buf);
+  scan_graph(obj, arg, 0);
 }
 
 double loadgraphval(struct text_object *obj) {
--- a/src/core.cc
+++ b/src/core.cc
@@ -725,11 +725,9 @@ struct text_object *construct_text_object(char *s, const char *arg, long line,
   DBGP2("Adding $cpubar for CPU %d", obj->data.i);
 #ifdef BUILD_GUI
   END OBJ(cpugraph, &update_cpu_usage) get_cpu_count();
-  char *buf = nullptr;
   SCAN_CPU(arg, obj->data.i);
-  buf = scan_graph(obj, arg, 1);
+  scan_graph(obj, arg, 1);
   DBGP2("Adding $cpugraph for CPU %d", obj->data.i);
-  free_and_zero(buf);
   obj->callbacks.graphval = &cpu_barval;
   obj->callbacks.free = &free_cpu;
   END OBJ(loadgraph, &update_load_average) scan_loadgraph_arg(obj, arg);
@@ -1244,13 +1242,9 @@ struct text_object *construct_text_object(char *s, const char *arg, long line,
   END OBJ(memwithbuffersbar, &update_meminfo) scan_bar(obj, arg, 1);
   obj->callbacks.barval = &mem_with_buffers_barval;
 #ifdef BUILD_GUI
-  END OBJ(memgraph, &update_meminfo) char *buf = nullptr;
-  buf = scan_graph(obj, arg, 1);
-  free_and_zero(buf);
+  END OBJ(memgraph, &update_meminfo) scan_graph(obj, arg, 1);
   obj->callbacks.graphval = &mem_barval;
-  END OBJ(memwithbuffersgraph, &update_meminfo) char *buf = nullptr;
-  buf = scan_graph(obj, arg, 1);
-  free_and_zero(buf);
+  END OBJ(memwithbuffersgraph, &update_meminfo) scan_graph(obj, arg, 1);
   obj->callbacks.graphval = &mem_with_buffers_barval;
 #endif /* BUILD_GUI*/
 #ifdef HAVE_SOME_SOUNDCARD_H
@@ -1829,8 +1823,9 @@ struct text_object *construct_text_object(char *s, const char *arg, long line,
   END OBJ_ARG(
       lua_graph, nullptr,
       "lua_graph needs arguments: <function name> [height],[width] [gradient "
-      "colour 1] [gradient colour 2] [scale] [-t] [-l]") char *buf = nullptr;
-  buf = scan_graph(obj, arg, 100);
+      "colour 1] [gradient colour 2] [scale] [-t] [-l]") auto [buf, skip] =
+      scan_command(arg);
+  scan_graph(obj, arg + skip, 100);
   if (buf != nullptr) {
     obj->data.s = buf;
   } else {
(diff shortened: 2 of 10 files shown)


miblodelcarpio:
Nice work, @bi4k8. This lets me omit the width / height from my `${upspeedgraph}` and `${downspeedgraph}` declarations and have those values picked up from `default_graph_height` and `default_graph_width` as they were before.

I do, however, find that variables nested within those declarations fail to parse. Is this intended? Here are example declarations that fail:

`${downspeedgraph enp0s31f6 000000 ${color4} 10000 -t -l}`
`${downspeedgraph enp0s31f6 000000 $color4 10000 -t -l}`

The new behaviour is actually better than 1.17.0-1 in that it draws the graph in a default red, rather than throwing an `X Error: Display 55db4fdabe20` and displaying garbage.

brndnmtthws (approved):

bi4k8:
> Nice work, @bi4k8. This lets me omit the width / height from my `${upspeedgraph}` and `${downspeedgraph}` declarations and have those values picked up from `default_graph_height` and `default_graph_width` as they were before.
> 
> I do, however, find that variables nested within those declarations fail to parse. Is this intended? Here are example declarations that fail:
> 
> `${downspeedgraph enp0s31f6 000000 ${color4} 10000 -t -l}` `${downspeedgraph enp0s31f6 000000 $color4 10000 -t -l}`
> 
> The new behaviour is actually better than 1.17.0-1 in that it draws the graph in a default red, rather than throwing an `X Error: Display 55db4fdabe20` and displaying garbage.

Unfortunately, you can't arbitrarily substitute Conky "variables" the way you might expect from the name. `${color1}` is more of a command than a variable, so it doesn't work to place it where another command expects a color name. It certainly isn't a bad idea to support this, as it would be intuitive, but I don't think this functionality existed so far.

A workaround for what you're doing is to use templates, which *are* substituted the way you would expect:
```
conky.config = {
...
  template1 = "00ff00",
}
conky.text = [[
${downspeedgraph wlan0 50,200 000000 ${template1} 1000 -t}
]]
```

miblodelcarpio:
Wonderful, cheers for the info, @bi4k8. And this template workaround is golden! I guess pretty much all of us from #1428 may as well benefit from this at least for the interface name.

Issue #1017, opened 2020-10-09
Add support for right aligned columns


maxdraki:
**Enhancement**

Allow text and objects to be right aligned at specific points in conky - similar to ${goto 20} but right aligned instead of left aligned

So output can be something like:

| Column A | Column B |
|---------|---------:|
| Normal | Right |
| Row 2 | Value 2 |
| Row 3 | Value 3 |

maxdraki:
bump

Issue #940, opened 2020-02-11
$alignr, $alignc in combination of ${goto x} or ${offset x} possible?


beta-tester:
**Issue**

$alignr, $alignc in combination of ${goto x} or ${offset x} possible? 

**Information**

i want to arcive that in one line there multople text elements appear alinged to the right at a specific position.
for example:
```
text1        text2              test3
▲                ▲                  ▲
|                |                  |<-aligned right
|                |<-aligned right at position x
|<-(alinged to the left)

                 |<-----------------| (alignr + offset)
    |----------->| (goto + alignr)

```
i want that text1 is alinged to the very left (thist is no problem)
text3 sould be aligned to the very right (this is no problem)
but text2 should stay aligned right at a specific position so that the last character will stay on a fixed position.
this is the problem. because when text3 changes its length, text2 will shift in its position to the left by the amount of the length of text3.

```lua
conky.config = {
...
    minimum_width = 250,
    maximum_width = 250,
...
};

conky.text = [[
...
${if_mounted /media/99-USB}
99-USB $alignr ${offset -95} ${diskio_read /dev/sda}▲ $alignr ${diskio_write /dev/sda}▼
${diskiograph_read /dev/sda 24,120} ${goto 130}${diskiograph_write /dev/sda 24,120}
/media/99-USB $alignc ${fs_used /media/pi/99-USB} / ${fs_size /media/99-USB} $alignr ${fs_used_perc /media/99-USB} %
${fs_bar /media/99-USB}
$endif
...
]];
```

beta-tester:
still alive from my side

Issue #75, opened 2014-12-26
RTL languages not supported


moceap:
![10864042_420651738084229_4584166603244457579_o](https://cloud.githubusercontent.com/assets/4345211/5558284/f8ca160e-8d31-11e4-9226-7b84f757abef.jpg)

Should be like الجمعة ، ديسمبر , there is a way here to fix this issue :

https://github.com/faisal-oead/conky-arabic-support

I think this issue must fix immedetly. Thank you.

moceap:
@Tucos 

I try to create a pullrequest but I can't due to changes on conky... I hope if you create a new patch and ask a pullrequest.

Thanx

Tucos:
Sorry, I will not bother with pull requests myself. The patch should still
apply cleanly to 1.9.1, as there's only been 1 commit in an unrelated file.
The master branch is conky 2.x, which is a rewrite of pretty much
everything. Either somebody other than me would have to rebase the patch
for that branch, or you need to make sure to checkout 1.9.1 before trying
to apply the patch.

brndnmtthws:
Can you open a PR against master?

su8:
@lasers we need @brndnmtthws as I don't understand cmake how to find linkage dependencies.

PR #1501, opened 2023-04-12
Better Text Rendering for cairo / lua

### LINKED ISSUE #900: Can I use the Imlib2 Lua bindings for font rendering?
TL;DR Trying to use font related `imlib_…` functions in Lua apparently does nothing.

I've been creating a fairly elaborate system of cairo based lua widgets for conky [here](https://github.com/philer/polycore). However, cairo's text rendering is extremely limited as explained in the [official documentation](https://www.cairographics.org/manual/cairo-text.html#cairo-text.description). Hence, I was hoping to get some better font rendering via Imlib2, which conky provides Lua bindings for. Unfortunately, they apparently don't work – or I am doing it wrong.

### Details

I've tried this on both version 1.11.5_pre (Lua 5.3), compiled today, and on version 1.10.8 (Lua 5.1) installed via package manager on Linux Mint 19. Imlib2 is listed in the Lua bindings section of `conky -v`.

`require "imlib2"` works. All the expected globals are present. I can call any number of `imlib_…` functions from within my code, except none of the ones I'm interested do or return anything. For instance the following code:

    local number_return
    local font_paths = imlib_list_font_path(number_return)
    print(font_paths, number_return)

outputs `nil	nil`. The [function's documentation](https://docs.enlightenment.org/api/imlib2/html/imlib2_8c.html#6e5611447a520cbf1966fef9b274b19e) is quite clear that it *should* return a list of strings, which I understand would need to be converted to a Lua table before they can be used.

As a further example, if I run

    local font = imlib_load_font("ubuntu/10")
    imlib_context_set_font(font)
    imlib_text_draw(0, 0, "hello world")

I get an error message:

    ***** Imlib2 Developer Warning ***** :
	This program is calling the Imlib call:
	imlib_text_draw_with_return_metrics();
	With the parameter:
	font
	being NULL. Please fix your program.

which is expected, since `font` is also `nil`. Meanwhile the "ubuntu" font is most definitely available in TrueType format on my system.

So now I'm facing two issues:

1. I have no idea if it's possible to use imlib from within my Lua code.
2. Assuming it's possible I still can't find any documentation telling me how. There is apparently a `cairo_draw_image` function defined in a cairo/imlib2 helper but I'd still have to rely on a lot of trial-and-error to figure out if/how I can use that to solve my problem.

I'd be quite grateful for some hints here. Let me know if I should provide any further details.

simotek:
# Description
## Introduction
I have attempted to resolve #900 by creating a new lua binding specifically for doing font rendering in Cairo using Freetype and Harfbuzz, fontconfig is also used to locate the correct fonts. I have chosen to follow the approach of the `imlib2_helper` binding and not expose functions from the headers of each of these libraries but rather provide simple functions that "internally" use each of these libraries.

At the time of creating this PR I have no fixed thoughts on the API other then it works for my use case and tries to balance the simple act of rendering an English string  while also allowing complex options such as RTL text as requested in #75 as well as being able to center and right align text #940  and #1017.  So your thoughts on the API topic would be much appreciated, once the API is finalized I will add appropriate documentation. currently this is the last major step for be before being able to move to a lua only config.

## Current API Description (Happy to make changes here)
Loading a font from a file is a slow enough operation that we probably don't want to do it multiple times so I separated loading fonts and freeing memory at close into separate functions.

```
/* font is a string containing the name of the font to load */
/* this will be matched on a best effort basis and FontConfig will fall back to a similar font if the specified font isn't found */
FontData * cairo_text_hp_load_font(const char *font, int font_size);
void cairo_text_hp_destroy_font(FontData *font);
```

The next set of functions cover the "Simple" English use cases, I could have added a parameter for alignment but I was trying to keep the common Left case as simple as possible.

```
void cairo_text_hp_simple_show(cairo_t *cr, int x, int y, const char *text, FontData *font);
void cairo_text_hp_simple_show_center(cairo_t *cr, int x, int y, const char *text, FontData *font);
void cairo_text_hp_simple_show_right(cairo_t *cr, int x, int y, const char *text, FontData *font);
```
The next function is the "catch all" with the main implementation and allows tweeking of the internalization options such as text direction in Harfbuzz for people looking to use non latin languages.
```
/* 
 * Direction calls hb_direction_from_string example values are LTR and RTL
 *   https://harfbuzz.github.io/harfbuzz-hb-common.html#hb-direction-from-string
 * Script is an ISO 15924 4 character string, "Zyyy" can be used for "Common" and "Zinh"
 *  for "Inherited". 
 *  https://harfbuzz.github.io/harfbuzz-hb-common.html#hb-script-from-string
 * Language is a BCP 47 language tag. eg "en" or "en-US"
 */
void cairo_text_hp_intl_show(cairo_t *cr, int x, int y, cairo_text_alignment_t alignment, const char *text, 
                                                    FontData *font,  const char *direction, const char *script, const char *language);
```
Finally when doing layouting sometimes its useful to know how wide the final string will be so the following find the text width,  again there is a simple latin version and an internationalized version. In the future I'll also provide a Height function which will be important for supporting vertical text.

```
int cairo_text_hp_simple_text_width(const char *text, FontData *font);
int cairo_text_hp_intl_text_width(const char *text, FontData *font, const char *direction, const char *script, const char *language);
```

## Screenshot
In the following screenshot the "Cairo" text is using Cairo's built in basic text rendering, the "Conky" text is using Conky's standard text rendering and the "Freetype" strings are using this new API with Freetype and Harfbuzz rendering the text in Cairo.  All these cases are just using the default settings as found on openSUSE Tumbleweed.

shot-2023-04-11_21-13-49


.github/workflows/build-and-test-linux.yaml
brndnmtthws, line 1: We could probably enable this in the macos build too, if you want to take a stab at that.

lua/libcairo_text_helper.h
brndnmtthws, line 1: The functions in here to be documented in `conky/doc/lua.yml`.

brndnmtthws (changes requested):
Looks great overall, just need to add documentation to https://github.com/brndnmtthws/conky/blob/main/doc/lua.yaml for the new functions.

simotek:
Thanks for the feedback, i'll update the docs and finish of implementing the Top to Bottom features then finish up a couple of configs to use them before sending through some updates. But its good to know i'm heading in the right direction.

moceap:
After trying this patches, bug #75 still exist!

Screenshot_٢٠٢٣٠٧٠٤_٠٣٢٨٤١

simotek:
> After trying this patches, bug #75 still exist!
> 
> Screenshot_٢٠٢٣٠٧٠٤_٠٣٢٨٤١

Yes that is currently to be expected, the WIP patch just had the LTR support so I could get feedback on the API, RTL is almost there I just need to finish writing the tests.

Also this change only works with the lua API, however it is the first step in my project to add a toolkit to make using the lua API much simpler. Although that is further from completion.

simotek:
@moceap Can you confirm that the following screenshot looks ok, it should be the days of the week.
shot-2023-07-20_15-27-09

I'll probably do just a little more testing of my actual real world usecases before sending through an updated pull request

simotek:
shot-2023-07-20_15-51-45 

This one uses Noto Sans Arabic instead.

moceap:
No it's not OK.

Letters in Arabic should be connected:

You write:

الإثنين
يوم الثلاثاء
الأربعاء
يوم الخميس
جمعة
السبت
الأحد

Look at them here in the browser, they are connected.

simotek:
> No it's not OK.
> 
> Letters in Arabic should be connected:
> 
> You write:
> 
> الإثنين يوم الثلاثاء الأربعاء يوم الخميس جمعة السبت الأحد
> 
> Look at them here in the browser, they are connected.

Thanks for the confirmation, that's what I thought i'll have to have a closer look at whats going wrong, the left to right seems to be ok.

lua/libcairo_text_helper.h
```diff
@@ -0,0 +1,281 @@
...
+  hb_language_t text_language = hb_language_from_string (language, -1);
+  
+  /* Draw text */
+  /* Create a buffer for harfbuzz to use */
+  hb_buffer_t *buf = hb_buffer_create();
+
+  //alternatively you can use hb_buffer_set_unicode_funcs(buf, hb_glib_get_unicode_funcs());
+  hb_buffer_set_unicode_funcs(buf, hb_glib_get_unicode_funcs());
+  hb_buffer_set_direction(buf, text_direction); /* or LTR */
+  hb_buffer_set_script(buf, text_script); /* see hb-unicode.h */
```
Caellian, line 154: Which `text_script` are you using to test out arabic? It should be "Arab" as defined in [`hb-common.h`](https://github.com/harfbuzz/harfbuzz/blob/fd5eb10d3b4cfee1bb10614ff6719378c949492e/src/hb-common.h#L526C17-L526C17). I'm asking because you're referencing `hb-unicode.h` which might be the reason why HarfBuzz is not shaping the arabic example properly.

lua/libcairo_text_helper.h
```diff
@@ -0,0 +1,281 @@
...
+                             const char *direction, const char *script, const char *language)
+{  
+  /* FIXME: Support others */
+  hb_direction_t text_direction = hb_direction_from_string(direction, -1);
+  if (text_direction == HB_DIRECTION_INVALID) {
+    text_direction = HB_DIRECTION_LTR;
+  }
+  hb_script_t text_script = hb_script_from_string(script, -1);
+  if (text_script == HB_SCRIPT_UNKNOWN) {
+    text_script = HB_SCRIPT_COMMON;
```
Caellian, line 142: Try setting this to `HB_SCRIPT_ARABIC` (string literal) instead of `HB_SCRIPT_COMMON` and see if it works (though it will break everything else).

lua/libcairo_text_helper.h
```diff
@@ -0,0 +1,281 @@
...
+  hb_language_t text_language = hb_language_from_string (language, -1);
+  
+  /* Draw text */
+  /* Create a buffer for harfbuzz to use */
+  hb_buffer_t *buf = hb_buffer_create();
+
+  //alternatively you can use hb_buffer_set_unicode_funcs(buf, hb_glib_get_unicode_funcs());
+  hb_buffer_set_unicode_funcs(buf, hb_glib_get_unicode_funcs());
+  hb_buffer_set_direction(buf, text_direction); /* or LTR */
+  hb_buffer_set_script(buf, text_script); /* see hb-unicode.h */
```
Caellian, line 257: This should be `hb-common.h` as well. Just pointing it out (so it's easier for others to pick up), not that these comments will affect anything.

lua/libcairo_text_helper.h
```diff
@@ -0,0 +1,281 @@
...
+  hb_direction_t text_direction = hb_direction_from_string(direction, -1);
+  if (text_direction == HB_DIRECTION_INVALID) {
+    text_direction = HB_DIRECTION_LTR;
+  }
+  hb_script_t text_script = hb_script_from_string(script, -1);
+  if (text_script == HB_SCRIPT_UNKNOWN) {
+    text_script = HB_SCRIPT_COMMON;
+  }
+    
+  hb_language_t text_language = hb_language_from_string (language, -1);
```
Caellian, line 145: Is the language set to "ar"?

Caellian (commented):
Never used HarfBuzz, but from what I picked up googling it, the way you're using it seems fine.

lua/libcairo_text_helper.h
```diff
@@ -0,0 +1,281 @@
...
+  cairo_set_source_rgba (cr, 1.0, 1.0, 1.0, 1.0);
+  cairo_set_font_face(cr, font->cairo_ft_face);
+  cairo_set_font_size(cr, font->font_size);
+  cairo_show_glyphs(cr, cairo_glyphs, glyph_count);
+
+  free(cairo_glyphs);
+  hb_buffer_destroy(buf);
+}
+
+void cairo_text_hp_simple_show(cairo_t *cr, int x, int y, const char *text, FontData *font)
```
Caellian, line 220: Note that there's no need for these `cairo_text_hp_simple_show_*` convenience functions. toluapp supports default values and harfbuzz can infer defaults too.

You can make all of the properties optional by:
- adding a call to `hb_buffer_guess_segment_properties (hb_buffer);` in `cairo_text_hp_intl_show` after _specified properties_ are set.
- setting argument order to be: `language` (because it's inferred from system (least smart)), `script` (inferred from buffer contents), `direction` (inferred from script)
  - as documented [here](https://harfbuzz.github.io/harfbuzz-hb-buffer.html#hb-buffer-guess-segment-properties)
  - that order would minimize the need for supplying well inferred values to set the one without a good default
- specifying argument defaults to be `nullptr`/`NULL` in the header
  - and not calling `hb_*_from_string` if the value is null (i.e. empty string).

Caellian: In the same line of reasoning I'd position cairo `alignment` argument after the `font` argument and set it to default to `CAIRO_TEXT_ALIGN_LEFT`.

Caellian: Also, the binding function should be able to load and cache a font. The cache can leak memory (no need to call free) as it would be weird to manage such things from a lua script and fonts will likely be needed for each draw call.

So the signature of `cairo_text_hp_intl_show` in .pkg would finally look like: `void cairo_text_hp_intl_show(cairo_t *cr, int x, int y, const char *text, const char *font = NULL, int font_size = 12, cairo_text_alignment_t alignment = CAIRO_TEXT_ALIGN_LEFT, const char *language = NULL, const char *script = NULL, const char *direction = NULL)`.

moceap:
@simotek Take a look https://github.com/HOST-Oman/libraqm, It may fast the process.

lua/libcairo_text_helper.h
simotek: Sorry I didn't see this until now, I'll certainly swap to the idea of default arguments that seems much better. Leaking fonts seems reasonable to me as well so i'm happy to just add a cache.

lua/libcairo_text_helper.h
simotek: Yep, the full setting was {font_family="Noto Sans Arabic", font_size=16, align=CAIRO_TEXT_ALIGN_LEFT, font_direction="RTL", font_script="HB_SCRIPT_ARABIC", font_language="ar"}

simotek:
> @simotek Take a look https://github.com/HOST-Oman/libraqm, It may fast the process.

One thing I noticed at https://github.com/HOST-Oman/libraqm/blob/9d7ca9d59dba38e777bdcd096870776dab5e6e66/src/raqm.c#L1793C1-L1793C48 is that for the RTL case they seem to have hardcoded a -1 into the calculation for character position, So I might have to have a play with that later. 

It seemed like hardcoding a -1 into the width of each character would solve the issue but I didn't want to just do it without seeing it documented, but the fact this other codebase is suggests its probably the right solution

lua/libcairo_text_helper.h
simotek: I am implementing most of this, with the exception of `language` which i'm going to default to english if not explicitly set. The reason being if I create a config / script for conky and want to distribute it, I don't want it to start breaking in random ways because other people are using a different language. Also much of conky's outputs such as process names CPU percentages etc work best in English. In this case if you are doing something that requires an explicit language you'll need to set it as well.

simotek:
So far I've done all the interface changes, other then caching fonts, which is next on my list.

As for the RTL codepaths, subtracting 1/10th of the font size seems to get much closer then -1 especially for larger font sizes. Its not perfect for all fonts at all sizes though, but is significantly better, i'll follow this up with upstream closer

@moceap I presume this screenshot is getting much closer. At this point I have no idea if the text is actual words.

shot-2024-02-15_11-53-40

simotek:
Also if anyone has any idea what Ubuntu packages are needed to get cmake to find freetype that would be much appreciated so that I don't need to spin up a VM and test.

lua/libcairo_text_helper.h
simotek: I've had a shot at doing this in a way that hopefully makes sense, for some reason i couldn't generate docs locally but that doesn't seem related to any of my changes.

If your happy with that i'll probably do similar for the imlib2 bindings that also aren't currently documented.

.github/workflows/build-and-test-linux.yaml
simotek: Once I figure out why the ubuntu builds are failing i'll have a shot at enabling it.

doc/lua.yaml
```diff
@@ -184,3 +184,44 @@ values:
...
+      | font_size           | The font size.                                                                       |
+      | alignment           | One of `CAIRO_TEXT_ALIGN_LEFT`, `CAIRO_TEXT_ALIGN_RIGHT`, `CAIRO_TEXT_ALIGN_CENTER`. |
+      |                     | Default value: `CAIRO_TEXT_ALIGN_LEFT`                                               |
+      | language            | A string containing a BCP 47 language tag.                                           |
+      |                     | Default value: `en`                                                                  |
+      | script              | A string containing a ISO 15924 script tag.                                          |
+      |                     | Default value: auto-detect from text                                                 |
+      | direction           | A string representing text direction eg `LTR`, `RTL` or `TTB`                        |
+      |                     | Default value: auto-detect from text otherwise `LTR`                                 |                                                               |
```
Caellian, line 206: Extra pipe (`|`) at the end.

doc/lua.yaml
```diff
@@ -184,3 +184,44 @@ values:
...
+      | cr                  | The `cairo_t` to render to.                                                          |
+      | text                | The text to render.                                                                  |
+      | font                | The name of the font to be used, `Fontconfig` is used to search for the font.        |
+      | font_size           | The font size.                                                                       |
+      | language            | A string containing a BCP 47 language tag.                                           |
+      |                     | Default value: `en`                                                                  |
+      | script              | A string containing a ISO 15924 script tag.                                          |
+      |                     | Default value: auto-detect from text                                                 |
+      | direction           | A string representing text direction eg `LTR`, `RTL` or `TTB`                        |
+      |                     | Default value: auto-detect from text otherwise `LTR`                                 |
```
Caellian, line 223: Instead of putting default value into the next row, it might be better to put it in the same cell and use `<br/>` so that it looks like this:

| Argument | Description |
|-|-|
| language | A string containing a BCP 47 language tag. <br/>  Default value: `en` |

Caellian (commented):
Some inputs on tables in docs.

Caellian:
> Also if anyone has any idea what Ubuntu packages are needed to get cmake to find freetype that would be much appreciated so that I don't need to spin up a VM and test.

Should be `libfreetype6-dev`, see: https://askubuntu.com/questions/1347131/what-is-the-reason-of-cmake-error-when-using-freetype2-header
- CMake is crappy at locating dependencies sometimes (amongst a rather long list of other things), if you have that package installed, send the output of `apt-file list libfreetype6-dev` so we can patch the [Find script](https://cmake.org/cmake/help/latest/module/FindFreetype.html).

simotek:
> > Also if anyone has any idea what Ubuntu packages are needed to get cmake to find freetype that would be much appreciated so that I don't need to spin up a VM and test.
> 
> Should be `libfreetype6-dev`, see: https://askubuntu.com/questions/1347131/what-is-the-reason-of-cmake-error-when-using-freetype2-header
> 
>     * CMake is crappy at locating dependencies sometimes (amongst a rather long list of other things), if you have that package installed, send the output of `apt-file list libfreetype6-dev` so we can patch the [Find script](https://cmake.org/cmake/help/latest/module/FindFreetype.html).

I swapped it and the XFT implementation to using cmake's `find_package(Freetype REQUIRED)`, the old XFT implementation worked fine for me on openSUSE, but obviously it had issues on ubuntu

doc/lua.yaml
simotek: This also should be fixed.

simotek:
I think this is finally getting close and i've finally got it doing everything I need from it.

All the text in the below screenshot is using the new conky lua API including the stiff with borders and Gradients etc. The conky memory leak is unrelated and fixed in #2083

.github/workflows/build-and-test-linux.yaml
simotek: this seems to be working

lua/libcairo_text_helper.h
simotek: Also now done.

Dockerfile
```diff
@@ -40,11 +43,25 @@ RUN apt-get update \
   libxml2-dev \
   libxmmsclient-dev \
   libxnvctrl-dev \
+  make \
   ninja-build \
   patch \
   && apt-get clean \
   && rm -rf /var/lib/apt/lists/*
 
+# Compile CMake, we need the latest because the bug here (for armv7 builds):
```
brndnmtthws, line 52: Is adding this back intentional? It was removed in https://github.com/brndnmtthws/conky/commit/a533bcb8e41c57fdfa7a4aec7f3c2cd6c34734fe, so I'm wondering if this is just a merge issue.

cmake/ConkyBuildOptions.cmake
```diff
@@ -26,15 +26,15 @@ if(NOT CMAKE_BUILD_TYPE)
     set(
       CMAKE_BUILD_TYPE Debug
       CACHE
-      STRING
-      "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel."
+        STRING
```
brndnmtthws, line 29: It looks like the only thing that changed in this file is the formatting. I think it would be better to preserve the formatting here.

cmake/ConkyPlatformChecks.cmake
```diff
@@ -499,32 +494,32 @@ set(conky_libs ${conky_libs} ${LUA_LIBRARIES})
 set(conky_includes ${conky_includes} ${LUA_INCLUDE_DIR})
 include_directories(3rdparty/toluapp/include)
 
-# Check for libraries used by Lua bindings
-if(BUILD_LUA_CAIRO)
+  # Check for libraries used by Lua bindings
```
brndnmtthws, line 497: Preserve formatting.

lua/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@
 #
 # Please see COPYING for details
 #
-# Copyright (c) 2005-2024 Brenden Matthews, et. al. (see AUTHORS) All rights
+# Copyright (c) 2005-2021 Brenden Matthews, et. al. (see AUTHORS) All rights
```
brndnmtthws, line 6: Preserve this, along with the formatting.

lua/libcairo_text_helper.h
```diff
@@ -0,0 +1,447 @@
+/*
+ *
+ * Conky, a system monitor, based on torsmo
+ *
+ * Please see COPYING for details
+ *
+ * Copyright (c) 2005-2021 Brenden Matthews, Philip Kovacs, et. al.
```
brndnmtthws, line 7: 2024 (not a big deal, just noticed it).

brndnmtthws (changes requested):
The code looks pretty reasonable to me. A few minor nit picks (mostly formatting related). I'm not really familiar with the implementation details or harfbuzz, so I'll trust your judgement on that.

Not related to this PR, but we should probably separate the helper header declarations from the implementation code (putting the latter into separate files).

simotek:
I will write a blog post to go with the release but here is a couple of simple and more complex examples if anyone wants to have a play with them.  I'm going to use it as part of a larger lua framework that's creating some of the above screenshots [https://github.com/simotek/conky-bubbles](https://github.com/simotek/conky-bubbles) which will also get a blog post and probably a conference talk or two.

shot-2025-10-29_17-41-29

```
function conky_text_example()
    -- The script may be called before the window is created.
    if conky_window == nil then
        return nil
    end

    -- Create Surface and cairo object for drawing
    local cs = cairo_xlib_surface_create(conky_window.display,
                                         conky_window.drawable,
                                         conky_window.visual,
                                         conky_window.text_width,
                                         conky_window.text_height)
    ---
    -- Simple example displaying a date
    ---
    local cr = cairo_create(cs)
    local date_time = os.date("%Y-%m-%d %H:%M:%S")
    local font_family = "SUSE Thin"
    local font_size = 16
    local w, h = cairo_text_hp_text_size(date_time, font_family, font_size)
    local x = (conky_window.text_width - w)/2
    cairo_set_source_rgba(cr, 1.0,1.0,1.0,1.0)
    cairo_set_line_width(cr, 0.2)
    cairo_text_hp_show(cr, x, 10, date_time, font_family, font_size)

    cairo_stroke(cr)

    ---
    -- Complex fancy example
    ---

    -- Create a mesh with a background
    local title_gradient = cairo_pattern_create_mesh()
    cairo_mesh_pattern_begin_patch(title_gradient)
    cairo_mesh_pattern_line_to(title_gradient, 00, 10)
    cairo_mesh_pattern_line_to(title_gradient, 160, 5)
    cairo_mesh_pattern_line_to(title_gradient, 150, 55)
    cairo_mesh_pattern_line_to(title_gradient, -10, 65)
    cairo_mesh_pattern_line_to(title_gradient, 00, 10)

    -- Unfortunately Cairo uses values between 0.0 and 1.0 for rgb values
    cairo_mesh_pattern_set_corner_color_rgba(title_gradient, 0, 0.20, 0.59, 0.84, 1.00)
    cairo_mesh_pattern_set_corner_color_rgba(title_gradient, 1, 0.55, 0.26,	0.67, 1.00)
    cairo_mesh_pattern_set_corner_color_rgba(title_gradient, 2, 0.20, 0.59, 0.84, 1.00)
    cairo_mesh_pattern_set_corner_color_rgba(title_gradient, 3, 0.55, 0.18, 0.14, 1.00)
    cairo_mesh_pattern_end_patch(title_gradient)
    cairo_set_source(cr, title_gradient)

    -- Render the text
    cairo_text_hp_show(cr, 10, 40, "TESTING 1234", "Sixteen", 16)
    -- Border Line Width and Color
    cairo_set_line_width(cr, 1.2)
    cairo_set_source_rgba(cr, 0.00, 0.32, 0.52, 1.00)
    cairo_stroke(cr)

    ---
    -- Some Right to left and Top to bottom text
    ---
    cairo_set_source_rgba(cr, 1.0,1.0,1.0,1.0)
    
    local w, h = cairo_text_hp_text_size("السبت", "Noto Sans Arabic", 16, 
                                         CAIRO_TEXT_ALIGN_RIGHT, "ar", "HB_SCRIPT_ARABIC", "RTL")
    
    cairo_text_hp_show(cr, conky_window.text_width - w -25, 60, "السبت", "Noto Sans Arabic", 16,
                       CAIRO_TEXT_ALIGN_RIGHT, "ar", "HB_SCRIPT_ARABIC", "RTL")
    -- Use the same formatting for this top to bottom text
    cairo_text_hp_show(cr, 10, 70, "這是一些中文", "Source Han Sans TW", 16, 
                       CAIRO_TEXT_ALIGN_LEFT, "ch", "HB_SCRIPT_HAN", "TTB")

    -- Cleanup Memory usage
    cairo_destroy(cr)
    cairo_surface_destroy(cs)
end
```

simotek:
@brndnmtthws I think this is ready for final review again

brndnmtthws (approved):

Issue #1495, opened 2023-04-08
[Bug]: Mouse clicks aren't consumed when llua_mouse_hook returns true. (X11)


ErskaConti:
### What happened?

I noticed the (awesome) implementation of Mouse Events recently added to Conky.

but I think due to this being so new documentation does not exist... so i printed all the args passed to lua and explored stuff...

now I also love the fact that I can click-through the Conky window. but I'll want to consume^(aka, skip pass-through) the clicks from time to time.

I was expecting the event to be consumed by `return true;` in the hook-function. (I also tried `return false;`)

On a sidenote, each click-through also triggers Leave-Enter events (at least on KDE) as the desktop is selected so as to trigger the click on it.

### Version

conky-cairo 1.19.0-1 (AUR)

### Which OS/distro are you seeing the problem on?

Arch Linux

ErskaConti:
I upgraded my computer, which lead to re-installation of stuff.

now I'm running in OpenBox (ran in KDE^(why the f did I forget to say that) before) .

```
function mouse_hook(args)

  return true;
end
```

does consume clicks. 

so seems to work as expected under OpenBox.

***

**conky-cairo**  from AUR
v 1.19.1-1

Caellian:
Thanks for pointing out lack of documentation, I [added it to the wiki](https://github.com/brndnmtthws/conky/wiki/Lua:-Mouse-Events) so people can find out about it.

> I was expecting the event to be consumed by return true; in the hook-function. (I also tried return false;)

The way I implemented it, if the hook returns true it [should stop the propagation](https://github.com/brndnmtthws/conky/pull/955/files#diff-5da03e30dd643b06945321a0509384d16ad7783650cc3ca2f967dad551858cf7R471). Given that it works on _some_ WMs, I'd guess that it has to do with how KDE handles events.

Maybe if an event isn't captured by any floating windows KDE notifies the desktop code, it's impossible to tell without the config though. Please provide the config so we can know how conky window is mounted.

> click-through also triggers Leave-Enter events (at least on KDE) as the desktop is selected so as to trigger the click on it.

This _is_ a bug in the merged [PR](https://github.com/brndnmtthws/conky/pull/955) and I'm certain it has to do with how the feature is implemented. Passing events through probably causes X11 to think the desktop is now under the cursor. This has to be handled by conky by making it remember it passed the event through and preventing the callback trigger for enter/leave events to avoid quirks.

ErskaConti:
[...]

PR #1672, opened 2023-11-05
Add mouse events & friends for Wayland

Caellian:
## Description

This PR is a supplement to #955 (X11 mouse events) - it adds support for the mouse event hook to Wayland.

Progress of this PR:
- [x] Fix mouse events referencing X11 code
- [x] Attach event listeners to Wayland
- [x] Fill in missing data and pass all events to Lua
- [x] Test mouse hook for all event types
- [x] Fix broken X11 code
- [x] Clean up code
  - [x] Debug propagation focus flickering
- [x] Test again
- [x] Update documentation in `doc/`

Optional:
- [x] See whether #1495 can be fixed
  - Fixed

### Leave surface event doesn't contain cursor position
Can be handled by statically storing the last known position from the move event in the window. I won't store it in `window` struct to avoid its layout changing based on build flags but instead create a new static storage in `mouse-events.cc`.

### Keyboard modifier keys aren't provided with button press/release events
**I'm not providing modifier keys in Wayland** (the table will be empty). I looked into it and supporting that requires me to add dependency on `XKB`, map scan codes through it to keys, maybe even support keyboard layouts, and storing the state of all tracked modifiers statically.
I feel this functionality should be provided through an additional feature for keyboard support (e.g. `BUILD_KEYBOARD_EVENTS`) because it basically adds everything needed for that. As such, it might be better to implement that in a separate PR.
- That PR should probably disable `mods` on X11 as well if the flag isn't enabled and handle modifiers better (e.g. left vs. right <kbd>Shift</kbd> key).
- Keyboard support requires [special handling](https://wayland.app/protocols/wlr-layer-shell-unstable-v1#zwlr_layer_surface_v1:request:set_keyboard_interactivity) and isn't supported the best by some compositors (Hyprland: setting interactivity to [`on_demand`](https://wayland.app/protocols/wlr-layer-shell-unstable-v1#zwlr_layer_surface_v1:enum:keyboard_interactivity:entry:on_demand) completely steals keyboard focus until the layer closes)

### Mouse button event produces platform dependent event codes instead of a semantic value

This is explained in detail in the next section:

## Breaking changes

- `event.button` previously mapped to integers representing mouse buttons because it was passing those values from X11 directly, now `event.button` is a string value (e.g. `left`, `right`, `middle`) representing a semantic name of a button.
  - `left`, `right`, `middle` correspond to previous `1`, `2` and `3` values.
  - `event.button` will be `nil` when no semantic name is hardcoded in conky.
  - `event.button_code` is now provided and contains raw event codes
    - value will always be present for all pointer buttons, but `event.button` is less platform dependent so `event.button_code` should only be used as fallback.

- `event.mods` no longer reports buttons/keys that aren't modifiers.
- renamed `mod1` to `super`.

While the previous values could've been kept, I feel this new format will stand the test of time much better and the scripts will be more stable/portable.

## Choices

I'm not processing pointer events on Wayland `frame`s because it would make the implementation much more complicated while providing no benefit - just like `last_known_positions` is static, a frame queue would have to be as well. We don't need complex axis event handling so we can greedily process all events.
- This might cause larger overhead for movement events due to higher number of Lua dispatches, but I don't think optimizing those warrants the additional processing complexity.

## Out of scope

Errors are now handled better in `llua.cc`. i.e. `lua_L` will no longer be initialized if `lua_load` property points to a file that doesn't exist which will make many of the functions in the file return early because they're referencing a blank lua context anyway (and would've failed at a later point).

Bumped `toluapp` minimum cmake version so CMake stops complaining about the old one no longer being supported.

Allowed Lua libraries to be included in build on Wayland.

Added a macro for ConkyBuildOptions that lets the user running cmake know which of the selected options were disabled by the script and why.

Disabled requirement for `BUILD_ARGB` feature on Wayland as it's supported out of the box (by almost all compositors I can think of). Even `OWN_WINDOW` should work as Wayland allows surfaces to be transparent.

### X11 Fixes

Enabled propagation of additional events to desktop window such as mouse movement and expose event. Provided a function which handled propagation for most core X11 events properly.

Added optional Xinput2 dependency that enables better (though more CPU intensive) tracking for whether cursor is over conky by capturing global movement events.
- It's double optional as it falls back to old behavior if not present at runtime. 😄

### X11 Error handling

Removed `abort()` call when X11 errors are received.

Fixing current cursor area detection bug where mouse leave wasn't being detected if a window was overlapping conky proved to be a nuisance - `XQueryPointer` didn't return correct window under the cursor so I implemented a manual lookup function that checks window properties though `XGetWindowAttributes`. Call to that function was failing when it was being executed on windows querried directly before by `XQueryTree`.

Conky should handle X11 errors without closing as they can fail for various reasons which don't mean the application is in invalid state. Instead of calling `abort()` from `x11_error_handler`, calls that are crucial for conky to function properly should check for `Success` at call site and invoke `CRIT_ERR` with an appropriate message. [Documentation](https://www.x.org/releases/X11R7.6/doc/xproto/x11protocol.html#errors) also notes that:
> when a request terminates with an error, the request has no side effects

I added detection for `xcb-util-errors` which now enables printing more descriptive error messages (name of failed object (error_code) & request (major, minor)) and added a fallback for error_code when `xcb-util-errors` isn't installed on a machine.

## License

All added code is licensed under GPLv3.


src/display-wayland.cc
```diff
@@ -1084,7 +1160,7 @@ struct window *window_create(struct wl_surface *surface, struct wl_shm *shm,
   window->rectangle.y = 0;
   window->rectangle.width = width;
   window->rectangle.height = height;
-  window->scale = 0;
+  window->scale = 1;
```
Caellian, line 1163: I seemed to have had to change initial scale of window to 1 because I was otherwise getting:
```
wl_surface@3: error 0: Specified scale value (0) is not positive
```
, the conky window didn't show and the process didn't respond to `SIGINT`.

This only seems to have happened after `78ea7b69782e2be6b024d6f408d08d9b7ecd25d8` so I should check that out.

Caellian:
Did testing with a modified script from last PR (to account for new data):

<details>
<summary>New script</summary>
<pre>
function echo (what)
    os.execute("echo -n '" .. what .. "'")
end

local function print_mods (mods)
    echo("mods:\n")
    for k,v in pairs(mods) do
        if (v) then
            echo("\t- " .. k .. "\n")
        end
    end
end

function conky_mouse (event)
    if (event.type == "mouse_move") then
        return false
    end

    echo("type: " .. event.type .. ",\n")
    echo("x: " .. tostring(event.x) .. ", ")
    echo("y: " .. tostring(event.y) .. ", ")
    echo("x_abs: " .. tostring(event.x_abs) .. ", ")
    echo("y_abs: " .. tostring(event.y_abs) .. ",\n")
    echo("time: " .. tostring(event.time) .. ",\n")
    
    if (event.type == "button_down") then
        print_mods(event.mods)
        echo("button: " .. tostring(event.button) .. "\n")
        echo("button_code: " .. tostring(event.button_code) .. "\n")
    elseif (event.type == "button_up") then
        echo("button: " .. tostring(event.button) .. "\n")
        echo("button_code: " .. tostring(event.button_code) .. "\n")
    elseif (event.type == "mouse_scroll") then
        echo("direction: " .. tostring(event.direction) .. "\n")
    elseif (event.type == "mouse_move") then
        print_mods(event.mods)
    elseif (event.type == "mouse_enter") then
        -- pass
    elseif (event.type == "mouse_leave") then
        -- pass
    elseif (event.type == nil) then
        echo("Unknown event type.")
    end
    return false
end
function echo (what)
    os.execute("echo -n '" .. what .. "'")
end

local function print_mods (mods)
    echo("mods:\n")
    for k,v in pairs(mods) do
        if (v) then
            echo("\t- " .. k .. "\n")
        end
    end
end

function conky_mouse (event)
    if (event.type == "mouse_move") then
        return false
    end

    echo("type: " .. event.type .. ",\n")
    echo("x: " .. tostring(event.x) .. ", ")
    echo("y: " .. tostring(event.y) .. ", ")
    echo("x_abs: " .. tostring(event.x_abs) .. ", ")
    echo("y_abs: " .. tostring(event.y_abs) .. ",\n")
    echo("time: " .. tostring(event.time) .. ",\n")
    
    if (event.type == "button_down") then
        print_mods(event.mods)
        echo("button: " .. tostring(event.button) .. "\n")
        echo("button_code: " .. tostring(event.button_code) .. "\n")
    elseif (event.type == "button_up") then
        echo("button: " .. tostring(event.button) .. "\n")
        echo("button_code: " .. tostring(event.button_code) .. "\n")
    elseif (event.type == "mouse_scroll") then
        echo("direction: " .. tostring(event.direction) .. "\n")
    elseif (event.type == "mouse_move") then
        print_mods(event.mods)
    elseif (event.type == "mouse_enter") then
        -- pass
    elseif (event.type == "mouse_leave") then
        -- pass
    elseif (event.type == nil) then
        echo("Unknown event type.")
    end
    return false
end
</pre>
</details>

preview

All events except `mouse_enter` and `mouse_leave` seem to work.

src/display-wayland.cc
Caellian: I added the call to `wl_surface_set_buffer_scale` in this PR. It's not necessary... so I'll remove it.

3rdparty/toluapp/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@
 # Please note that the package source code is licensed under its own license.
 
 project ( toluapp C )
-cmake_minimum_required ( VERSION 3.4 )
+cmake_minimum_required ( VERSION 3.15 )
```
Caellian, line 8: 3.15 was released in 2019 (says ChatGPT), so all targeted platforms should have it and it removes the annoying warning that 3.4 won't be supported.

cmake/ConkyPlatformChecks.cmake
```diff
@@ -469,16 +469,15 @@ if(BUILD_X11)
     set(luaimlib2_libs ${IMLIB2_LIBS} ${IMLIB2_LDFLAGS} ${LUA_LIBRARIES})
     set(luaimlib2_includes
       ${IMLIB2_INCLUDE_DIRS}
-      ${LUA_INCLUDE_DIR}
-      ${X11_INCLUDE_DIR})
```
Caellian, line 473: Not sure why luaimlib2 had X11 in includes, it should be separate. All of these Lua features are now under BUILD_GUI instead of BUILD_X11 as they don't require/depend on X11 at all.

src/display-x11.cc
```diff
@@ -227,6 +230,29 @@ bool display_output_x11::shutdown() {
   return true;
 }
 
+inline void propagate_unconsumed_event(XEvent &ev, bool is_consumed) {
```
Caellian, line 233: This function was tested thoroughly on KDE plasma with background widgets and such.

Caellian: This code portion is now additionally invoked on cursor movement so that e.g. desktop shortcuts get proper on-hover highlighting.

src/display-x11.cc
```diff
@@ -227,6 +230,29 @@ bool display_output_x11::shutdown() {
...
+    ev.xmotion.x = ev.xmotion.x_root;
+    ev.xmotion.y = ev.xmotion.y_root;
+    XSendEvent(display, window.desktop, False, capture_mask, &ev);
+    if (ev.type == ButtonPress) {
+      XSetInputFocus(display, window.desktop, RevertToNone, ev.xmotion.time);
+    }
+  } else {
+    XGrabPointer(display, window.window, TRUE, capture_mask,
+      GrabModeAsync, GrabModeAsync, window.window, None, ev.xbutton.time);
+    XSetInputFocus(display, window.window, RevertToParent, ev.xmotion.time);
```
Caellian, line 252: Now if the event was consumed by conky, it will capture focus and grab pointer so that the previously focused app needs to be focused again - same as if they clicked on the desktop, but we don't want desktop to be focused.

brndnmtthws:
This looks great so far, just let me know when you're ready for a review.

src/display-x11.cc
Caellian: Removed all pointer grabbing, we don't need it _at all_ AFAICT

Caellian:
X11 errors now give very detailed error messages when compiled with `xcb-util-errors` (autodetected, only has to be present during build, not an error if it's not).
pow

I'll print them with `DBGP` though because they can be very spammy now with `BUILD_MOUSE_EVENTS`+`BUILD_XINPUT` and curor tracking.

Caellian:
@brndnmtthws Ready for review :smiling_face_with_tear:

Caellian:
> I'll print them with `DBGP` though because they can be very spammy now with `BUILD_MOUSE_EVENTS`+`BUILD_XINPUT` and curor tracking.

I reverted back to `XQueryPointer` (needed to look up last descendant), so this error doesn't occur anymore, but I didn't revert removing `abort()`.

Caellian:
Added nice notices to CMake so features don't just disappear.

2023-11-10-122627_hyprshot

Caellian:
All tweaks are done, can't seem to force CI to use the new x11.cc file that was pushed like 10 commits ago, and also reformatted with clang-format since then (so hash should be different).

This ended up being a bit larger PR than I thought :laughing:, but it tackles a lot of issues with both the previous implementation and Wayland support.

Let me know whether there's anything you'd like me to change @brndnmtthws.

brndnmtthws (approved):
Thanks again for the PR. In the future it would be nice to have separate changes (like the formatting or VS Code extension stuff) broken out into separate PRs just to make the review process a little easier, but I understand that is more work as well.

brndnmtthws:
This is merged from afa5f326e850636e484f751b3d8053364814506a on, but GitHub seems to have been unable to handle the merge.

PR #1729, opened 2024-02-22
Lua imlib2 fixes

simotek:
# Checklist

## Description

A range of fixes for the cairo_imlib2 lua helper, after using it for a couple of days

#### 1. Significantly improve error reporting
Previously if you had an error in your code you would get an error message for each call to imlib2 which lead to alot of spam. Instead we now detect and print our own error message for common errors then return to help with this.

#### 2. Implement Stride in a standard way
This doesn't actually fix anything but I noticed it could be improved when I was debugging something else.

#### 3. Scale Alpha masks correctly.
We weren't scaling the output of the alpha mask to the size of the image we were displaying, which lead to the ugly artifacts that can be seen below. (Ignore the general incompleteness of the actual config its half done)

** Broken Image: **
shot-2024-02-22_17-36-46

** Fixed Image: **
shot-2024-02-22_17-35-14

####  4. Add Docs
Adds similar docs to the one's being worked on for cairo-text-helper


doc/lua.yaml
```diff
@@ -185,3 +185,17 @@ values:
...
+      | Argument            | Description                                                                          |
+      |---------------------|--------------------------------------------------------------------------------------|
+      | file                | Path to the image to render.                                                         |
+      | cs                  | The `cairo_surface_t` to render to.                                                  |
+      | x,y                 | Position to render the image.                                                        |
+      | scale_x, scale_y    | The amount to scale the image, 1.0 provides no scaling, 2.0 for twice the size and   |
+      |                     | 0.5 for half size etc.                                                               |
+      |                     | Default value: No Scaling (1.0,1.0)
+
+      require('cairo_imlib2_helper') in your lua file.                                                  |
```
brndnmtthws, line 201: This formatting is messed up, check the output here: https://deploy-preview-1729--conkyweb.netlify.app/lua#ret_scale_x,ret_scale_y:cairo_draw_image(file,%20cs,%20x,%20y,%20scale_x,%20scale_y)

brndnmtthws (changes requested):
This is a good start, but can you use the logging functions from `logging.h` instead of `printf`? I think [`NORM_ERR`](https://github.com/brndnmtthws/conky/blob/51f6f0576cd1b924e83ba11ca6dfaa96e3b37f96/src/logging.h#L76-L81) should do what you need.

doc/lua.yaml
simotek: I couldn't use the functions in `logging.h` directly as unfortunately the lua bindings only currently have access to a C compiler.  So I added a simple C re implementation and used that, if this PR is accepted i'll also swap to that for the lua-text PR.

It looks like the docs issue was just a typo.

simotek:
To make this really usable for my use case I suspect i'll have one more feature commit to add to this PR later today but as this stands hopefully it is an improvement on the existing codebase.

doc/lua.yaml
brndnmtthws: Ah, I guess that makes sense.

Want to try this patch?: https://github.com/brndnmtthws/conky/pull/1740

I think cmake should do the right thing and assume C++ if we simply change the filename.

simotek: Yep Ill give it a shot when I find some time in the next couple of days.

simotek: Yep, that's worked thanks

simotek:
I've added one more function `cairo_place_image` which splits the existing function in half, keeping the existing API but also adding another that is more useful in some cases. Hopefully i'll get to fixing up that nix build tomorrow.

brndnmtthws (changes requested):

simotek:
@brndnmtthws I have no idea why `Nix build / build-and-check (macos-latest) (pull_request) ` is failing, it seems not to find imlib2 or something, but all the other builds seem fine.

brndnmtthws:
> @brndnmtthws I have no idea why `Nix build / build-and-check (macos-latest) (pull_request) ` is failing, it seems not to find imlib2 or something, but all the other builds seem fine.

Yeah that's odd, let me see if I can figure it out.

brndnmtthws:
`imlib2.h` and `Imlib2.h` conflict because we are now including `src/`, and I guess macos is less strict about case sensitivity.

lua/cairo_imlib2_helper.pkg
```diff
@@ -1,5 +1,11 @@
...
 $#include <libcairo_imlib2_helper.h>
 
-void cairo_draw_image(const char *, cairo_surface_t *, int, int, double, double,
+/* Paints the image onto the cairo_surface_t */
+void cairo_draw_image(const char *, cairo_surface_t *, int, int,
+                      double scale_x=1.0, double scale_y=1.0,
                       double * return_scale_w, double * return_scale_h);
+
+/* Places an image onto a cairo_t but doesn't call cairo_paint */
+void cairo_place_image(const char *file, cairo_t *cr, int x, int y,
```
brndnmtthws, line 10: Can you add documentation for this function? After that, I think we're good to go.

brndnmtthws (changes requested):

lua/cairo_imlib2_helper.pkg
simotek: Sorry I guess I left that in a different branch, ill dig it up tomorrow and commit it here.

simotek: Should be there now, sorry for the delay work's been busy.

brndnmtthws (approved):

PR #1745, opened 2024-02-24
Stop walking window tree if we get zero

brndnmtthws:
It looks like some of the child windows are zero, which causes the next call to XQueryTree to blow up.

cc @Caellian

### DIFF
--- a/src/x11.cc
+++ b/src/x11.cc
@@ -1384,8 +1384,13 @@ inline Window last_descendant(Display *display, Window parent) {
   while (
       XQueryTree(display, current, &_ignored, &_ignored, &children, &count) &&
       count != 0) {
-    current = children[count - 1];
+    auto next = children[count - 1];
     XFree(children);
+    if (next != 0) {
+      current = next;
+    } else {
+      break;
+    }
   }
 
   return current;


Caellian:
What's your WM and how do you mount conky (I'm assuming on root). I want to test this locally because no `parent` should return 0 (root window) as child from `XQueryTree`. I'd like to install your WM and debug `last_descendant()` a bit because 0 being returned as child is weird.

It might be better to check whether conky is drawing on root (if that's your config) from [code that's calling `last_descendant()`](https://github.com/brndnmtthws/conky/blob/56a81ef4fd1b9a994b6157bb94a69a660a27b684/src/display-x11.cc#L411) before calling `last_descendant` in order to skip to bound checking - `query_result` being default initialized would work the same for that case without the extra X11 calls.

The reason that function was needed is because plasma created 2/3 nested nodes for conky with the following settings:
```
own_window = true,
own_window_type = 'normal',
```
so there was no simpler way to check whether the cursor was over conky, because the window stored in globals points to the deepest descendant, and one of container windows is returned by `XQueryPointer`.

I think it's also redundant to check for all three of `root`, `window`, `desktop` from `conky_x11_window` in the [if statement below](https://github.com/brndnmtthws/conky/blob/56a81ef4fd1b9a994b6157bb94a69a660a27b684/src/display-x11.cc#L414), so I'd like to clean up/simplify that check.

brndnmtthws:
> I'm going to test this locally because no `parent` should return 0 (root window) as child from `XQueryTree`. I'd like to install your WM and debug `last_descendant()` a bit because 0 being returned as child is weird.

From the stack trace, it's calling `XQueryTree` with `current=0`, like this:

```
#13 0x00007ffff7dfac3c in XQueryTree (dpy=0x5555557b4c10, w=0, root=0x7fffffffdac0, parent=0x7fffffffdac0, children=0x7fffffffdad0, nchildren=0x7fffffffdaac) at /usr/src/debug/libx11/libX11-1.8.7/src/QuTree.c:47
```

So if `current` is 0, that implies it's either 1) calling with 0 on the first pass, or 2) eventually getting 0 as a child.

Caellian:
Found the cause of the crash from the issues; it's is in `x11_error_handler` and it crashes because I incorrectly calculated `size` for `error_name` when calling `snprintf`. Fixed it [here](https://github.com/brndnmtthws/conky/pull/1748).

Calling `XQueryTree` on root (0) happens when the cursor is on desktop. `query_x11_window_at_pos` should check if `last == 0` and return 0 instead of calling `last_descendant`. This isn't an error, though `XQueryTree` fails because of it. I added this check in the linker PR.

brndnmtthws:
Closing this in favour of #1748

Issue #1751, opened 2024-02-25
[Bug]: segfault when zwlr_layer_shell_v1 not available on Wayland


brndnmtthws:
### What happened?

Conky segfaults when zwlr_layer_shell_v1 is not available, but it should handle this gracefully. At the moment we call `zwlr_layer_shell_v1_get_layer_surface()` before checking we have `layer_shell`:

https://github.com/brndnmtthws/conky/blob/98087f9fc49ba5b0b70aab20f55ced81cb9269bb/src/display-wayland.cc#L585-L587

### Version

main

### Which OS/distro are you seeing the problem on?

Ubuntu

### Stack trace

```GDB
Starting program: /home/ppsmol/conky/build/src/conky -c conkyrc
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/aarch64-linux-gnu/libthread_db.so.1".
conky: drawing to single buffer
conky: invalid setting of type 'table'
[New Thread 0xfffff61fee60 (LWP 17344)]

Thread 1 "conky" received signal SIGSEGV, Segmentation fault.
0x0000fffff7a679d0 in wl_proxy_get_version () from /lib/aarch64-linux-gnu/libwayland-client.so.0
(gdb) r
The program being debugged has been started already.
Start it from the beginning? (y or n) y
Starting program: /home/ppsmol/conky/build/src/conky -c conkyrc
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/aarch64-linux-gnu/libthread_db.so.1".
conky: drawing to single buffer
conky: invalid setting of type 'table'
[New Thread 0xfffff61fee60 (LWP 17357)]

Thread 1 "conky" received signal SIGSEGV, Segmentation fault.
0x0000fffff7a679d0 in wl_proxy_get_version () from /lib/aarch64-linux-gnu/libwayland-client.so.0
(gdb) bt full
#0  0x0000fffff7a679d0 in wl_proxy_get_version () at /lib/aarch64-linux-gnu/libwayland-client.so.0
#1  0x0000aaaaaab1910c in zwlr_layer_shell_v1_get_layer_surface(zwlr_layer_shell_v1*, wl_surface*, wl_output*, unsigned int, char const*)
    (zwlr_layer_shell_v1=0x0, surface=0xaaaaaad462b0, output=0x0, layer=1, namespace_=0xaaaaaab7cf1f "conky_namespace")
    at src/wlr-layer-shell-client-protocol.h:230
[...]

PR #1752, opened 2024-02-25
Add wlr layer shell check

Caellian:
Closes: #1751

When a global is registered with `wl_registry_add_listener` it receives events for all supported extensions in bulk. `wl_display_roundtrip` call blocks until the initial events are exchanged so it's safe to check whether `wl_globals.layer_shell` is initialized right after that.


Caellian:
This needs testing. I don't have a WM without layer shell installed so I just made a quick PR as it was quicker than describing the change in the issue.

Issue #1477, opened 2023-03-30
[Bug]: raspberry pi ${color anycolor} turns text black 


rschell:
### What happened?

I have three Debian x86_64 machines on 6.2.9 which display as expected, with argb_visual set to true.  
![image](https://user-images.githubusercontent.com/1888300/228881591-138abe62-3a90-4158-bb74-2f6aa2b17010.png)

I also have four raspberry pi 4b machines where, if I place any ${color xxxx} text in the text field, the text colors turn black and cpu/memory graphs disappear.
![image](https://user-images.githubusercontent.com/1888300/228881862-e1f17130-9a40-4d13-90b0-44112bac1a11.png)

My workaround is to remove all references to ${color} in the text field of .conkyrc and use a default_color setting in the config field, which results in:
![image](https://user-images.githubusercontent.com/1888300/228883037-d64131a0-6ace-47ce-af44-57e3722b3d7a.png)

I don't know was is missing from the Raspberry Pi software configuration (running 6.1.19 kernels), any clues?  These .conkyrc setups where working just fine prior to 1.18.

When started in a terminal window, no error messages are produced when the .conkyrc file is reloaded:
conky: '/home/ron/.conkyrc' modified, reloading...
conky: desktop window (1200028) is subwindow of root window (3a2)
conky: window type - normal
conky: drawing to created window (0x3000002)
conky: drawing to double buffer

### Version

1.18.1+ currently 1.18.3

### Which OS/distro are you seeing the problem on?

Linux (other) [edit] all running xfce4 using xfwm4 compositor

### Conky config

```Lua
conky.config = {
	background = true, total_run_times = 0,
	double_buffer = true,
	update_interval = 1.0,
	alignment = 'top_right', gap_x = 10, gap_y = 10,
	minimum_width = 250, maximum_width = 400, minimum_height = 5,
	use_xft = true, font = '123:size=8',
	draw_shades = false, draw_outline = false, draw_borders = false, draw_graph_borders = false,
	own_window = true,
	own_window_type = 'normal',
	own_window_transparent = false,
	own_window_argb_visual = true, own_window_argb_value = 120,
	own_window_hints = 'undecorated,below,sticky,skip_taskbar,skip_pager',
	no_buffers = true,
	cpu_avg_samples = 2, net_avg_samples = 1,
	default_color = "gainsboro",
};

conky.text = [[
${font Arial:size=20}Pi-CNC
${font Arial:bold:size=10}SYSTEM ${hr 2}
[...]

PR #1768, opened 2024-03-03
Define text_object.data.l as "long long"

livanh:
# Checklist
- [X] I have linked to any relevant GitHub issues, if applicable

## Description

The definition of `text_object.data.l` is changed from `long` to `long long`. In 32-bit environments, `long` may be as short as 32 bits, and this can cause minor issues. Specifically, I was hitting #1477.

This change solved the issue for me, and it doesn't appear to cause new issues, even on a 64-bit platform. However, it may not be the best fix in the long run, as `int` and `long long` are still platform-specific and can hide similar bugs.

Fixed-size definitions (e.g. `uint64_t`) should be safer, but I'm not sure what the sizes should be (the defaults on x86_64 could be a good starting point). I can update the PR if this solutions turns out to be preferred.

### DIFF
--- a/src/text_object.h
+++ b/src/text_object.h
@@ -109,7 +109,7 @@ struct text_object {
     void *opaque; /* new style generic per object data */
     char *s;      /* some string */
     int i;        /* some integer */
-    long l;       /* some long integer */
+    long long l;  /* some long integer */
   } data;
 
   void *special_data;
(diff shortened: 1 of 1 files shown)


brndnmtthws:
I suggest using `int64_t`, less ambiguous and it's part of the C++ standard since C++11.

brndnmtthws (approved):

Issue #1767, opened 2024-03-03
[Bug]: Conky 1.19.7 makes tint2 crash


brunoGenX:
### What happened?

Hello,
Conky 1.19.7-1 crash and makes crash tint2 (taskbar).
Conky 1.19.7-2 work fine (issue 1748) but makes crash tint2
Conky and tint2 works fine with conky 1.19.6

`mars 03 09:22:00 archlinux kernel: tint2[2856]: segfault at 40 ip 000061e7396ec2f5 sp 00007ffd87bcfbf0 error 4 in tint2[61e7396c2000+3f000] likely on CPU 3 (core 1, socket 0)
mars 03 09:22:00 archlinux kernel: Code: 77 04 39 d6 7f e1 01 f1 31 c0 39 d1 0f 9d c0 c3 0f 1f 44 00 00 f3 0f 1e fa 41 56 49 89 fe 41 55 41 89 d5 41 54 41 89 f4 55 53 <49> 8b 5e 40 48 85 db 75 1b eb 6c 44 89 ea 44 89 e6 48 89 ef ff d1
mars 03 09:22:00 archlinux systemd[1]: Started Process Core Dump (PID 2860/UID 0).
mars 03 09:22:00 archlinux systemd-coredump[2861]: [🡕] Process 2856 (tint2) of user 1000 dumped core.
                                                   
                                                   Stack trace of thread 2856:
                                                   #0  0x000061e7396ec2f5 find_area_under_mouse (tint2 + 0x3d2f5)
                                                   #1  0x000061e7396cf216 n/a (tint2 + 0x20216)
                                                   #2  0x000061e7396cf796 handle_x_events (tint2 + 0x20796)
                                                   #3  0x000061e7396cf8f5 run_tint2_event_loop (tint2 + 0x208f5)
                                                   #4  0x000061e7396cfa35 tint2 (tint2 + 0x20a35)
                                                   #5  0x000061e7396c20ae main (tint2 + 0x130ae)
                                                   #6  0x00007546d57fdcd0 n/a (libc.so.6 + 0x25cd0)
                                                   #7  0x00007546d57fdd8a __libc_start_main (libc.so.6 + 0x25d8a)
                                                   #8  0x000061e7396c2805 _start (tint2 + 0x13805)
                                                   
                                                   Stack trace of thread 2857:
                                                   #0  0x00007546d58de88d syscall (libc.so.6 + 0x10688d)
[...]

PR #1809, opened 2024-04-10
Tweak XSendEvent arguments

Caellian:
Set propagate to `True`, to not propagate by default, as discussed in #1767.

XSendEvent `propagate` argument actually means "don't force propagation", so True value means the event will only get propagated if it's not handled, and False means the event will _always_ get propagated.

This is a followup PR to #1802.

### DIFF
--- a/src/x11.cc
+++ b/src/x11.cc
@@ -1410,7 +1410,7 @@ void propagate_x11_event(XEvent &ev) {
   }
 
   XUngrabPointer(display, CurrentTime);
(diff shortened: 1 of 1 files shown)


Caellian:
> Nice catch. Gotta love Xlib.

Xlib is comparable to door knobs with razors - looks straightforward to use until you try.

Thanks to @lineage-of-roots for checking the fine print on the docs and pointing out the mismatch between the name and the meaning. I remembered being stuck on this particular argument before, and changing it to `False` which apparently made `tint2` crash.

PR #1807, opened 2024-04-08
Forward XInput events

Caellian:
This PR fixes propagation of mouse events to gdk applications (e.g. caja).

It builds on top of #1802.

PR #1813, opened 2024-04-11
Fix missing y_abs argument copy

Caellian:


src/mouse-events.h
```diff
@@ -123,7 +123,7 @@ struct mouse_positioned_event : public mouse_event {
 
   mouse_positioned_event(mouse_event_t type, std::size_t x, std::size_t y,
                          std::size_t x_abs, std::size_t y_abs)
-      : mouse_event(type), x(x), y(y), x_abs(x_abs), y_abs(){};
+      : mouse_event(type), x(x), y(y), x_abs(x_abs), y_abs(y_abs){};
```
Caellian, line 126: Defaulted to 0 previously, found while debugging #1807.
Made a separate PR so #1807 can be squashed in order to make tracking down bugs easier in future.

brndnmtthws (approved):

PR #1818, opened 2024-04-11
Improve documentation for window types and hints

Caellian:
These options seem to cause a lot of confusion in the issues so I made them clearer in hope to reduce the number of issues asking why specific values don't work with their WM/DE.


brndnmtthws (approved):
Lovely, thanks!

PR #1819, opened 2024-04-12
Refactor x11 event handling

Caellian:
This PR refactors event handling into separate functions in order to make the event handling code more readable. It's separated out of #1807 in order to make that future PR cleaner.

### DIFF
--- a/src/display-x11.cc
+++ b/src/display-x11.cc
@@ -423,16 +423,30 @@ bool display_output_x11::main_loop_wait(double t) {
   return true;
 }
 
-#define EV_HANDLER(name)                                                     \
-  bool _conky_ev_handle_##name(conky::display_output_x11 *surface,           \
-                               Display *display, XEvent &ev, bool *consumed, \
-                               void **cookie)
-#define NOOP_EV_HANDLER(name) \
-  EV_HANDLER(name) { return false; }
+enum x_event_handler {
+  XINPUT_MOTION,
+  MOUSE_INPUT,
+  PROPERTY_NOTIFY,
+
+  EXPOSE,
+  REPARENT,
+  CONFIGURE,
+  BORDER_CROSSING,
+  DAMAGE,
+};
+
+template <x_event_handler handler>
+bool handle_event(conky::display_output_x11 *surface, Display *display,
+                  XEvent &ev, bool *consumed, void **cookie) {
+  return false;
+}
 
 #ifdef OWN_WINDOW
 #ifdef BUILD_XINPUT
-EV_HANDLER(xinput_motion) {
+template <>
+bool handle_event<x_event_handler::XINPUT_MOTION>(
+    conky::display_output_x11 *surface, Display *display, XEvent &ev,
+    bool *consumed, void **cookie) {
   if (ev.type != GenericEvent || ev.xcookie.extension != window.xi_opcode)
     return false;
 
@@ -490,7 +504,10 @@ EV_HANDLER(xinput_motion) {
   return true;
 }
 #endif /* BUILD_XINPUT */
-EV_HANDLER(mouse_input) {
+template <>
+bool handle_event<x_event_handler::MOUSE_INPUT>(
+    conky::display_output_x11 *surface, Display *display, XEvent &ev,
+    bool *consumed, void **cookie) {
   if (ev.type != ButtonPress && ev.type != ButtonRelease &&
       ev.type != MotionNotify)
     return false;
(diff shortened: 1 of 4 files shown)


src/display-x11.cc
```diff
@@ -679,6 +423,314 @@ bool display_output_x11::main_loop_wait(double t) {
...
+                                                   ev.xmotion.x_root,
+                                                   ev.xmotion.y_root, mods));
+    }
+  }
+#else  /* BUILD_MOUSE_EVENTS */
+  // always propagate mouse input if not handling mouse events
+  *consumed = false;
+#endif /* BUILD_MOUSE_EVENTS */
+
+  if (!own_window.get(*state)) return true;
```
Caellian, line 536: This is the only part that changes functionality.

src/x11.cc
```diff
@@ -1411,6 +1413,9 @@ void propagate_x11_event(XEvent &ev) {
 
   XUngrabPointer(display, CurrentTime);
   XSendEvent(display, i_ev->common.window, True, ev_to_mask(i_ev->type), &ev);
+  if (focus) {
+    XSetInputFocus(display, i_ev->common.window, RevertToParent, CurrentTime);
```
Caellian, line 1417: I moved `XSetInputFocus` here, after sending the event, if it's a ButtonPress.

src/display-x11.cc
```diff
@@ -374,266 +377,7 @@ bool display_output_x11::main_loop_wait(double t) {
...
-#endif /* BUILD_MOUSE_EVENTS */
-        if (own_window.get(*state)) {
-          /* if an ordinary window with decorations */
-          if ((own_window_type.get(*state) == TYPE_NORMAL) &&
-              !TEST_HINT(own_window_hints.get(*state), HINT_UNDECORATED)) {
-            /* allow conky to hold input focus. */
-            break;
-          }
-        }
-        break;
```
Caellian, line 578: These consecutive breaks made no sense to me, so I tracked down that the window type check was intended to avoid propagation of events (inner break) in the [original code](https://github.com/Caellian/conky/commit/7fbdfbd4d45a99b06def7065d66e02337323066d#diff-5da03e30dd643b06945321a0509384d16ad7783650cc3ca2f967dad551858cf7L466).

Over several commits, I seem to have made it do nothing.

src/display-x11.cc
```diff
@@ -679,6 +423,314 @@ bool display_output_x11::main_loop_wait(double t) {
...
+  *consumed = false;
+#endif /* BUILD_MOUSE_EVENTS */
+
+  if (!own_window.get(*state)) return true;
+  switch (own_window_type.get(*state)) {
+    case window_type::TYPE_NORMAL:
+    case window_type::TYPE_UTILITY:
+      // decorated normal windows always consume events
+      if (!TEST_HINT(own_window_hints.get(*state), HINT_UNDECORATED)) {
+        *consumed = true;
```
Caellian, line 542: So now, it still passes the events to Lua, but force-disables propagation in case of normal decorated window, or desktop. Also added comments to make it clear in future.

src/display-x11.cc
```diff
@@ -679,6 +423,314 @@ bool display_output_x11::main_loop_wait(double t) {
   return true;
 }
 
+#define EV_HANDLER(name)                                                     \
```
brndnmtthws, line 426: I'm not a fan of adding new macros. They make it difficult to debug things, and they're not type safe. Probably better to just write out the whole function definition, or you could get fancy with templates (but I think in this case it's not worth the effort).

brndnmtthws (changes requested):

src/display-x11.cc
```diff
@@ -423,16 +423,30 @@ bool display_output_x11::main_loop_wait(double t) {
...
+  CONFIGURE,
+  BORDER_CROSSING,
+  DAMAGE,
+};
+
+template <x_event_handler handler>
+bool handle_event(conky::display_output_x11 *surface, Display *display,
+                  XEvent &ev, bool *consumed, void **cookie) {
+  return false;
+}
```
Caellian, line 442: Replaced `EV_HANDLER` macro with a templated function.

Nice perk with this is that there's no need for explicit `NOOP_EV_HANDLER`.

src/display-x11.cc
```diff
@@ -675,33 +708,30 @@ EV_HANDLER(damage) {
...
 bool process_event(conky::display_output_x11 *surface, Display *display,
                    XEvent ev, bool *consumed, void **cookie) {
-#define HANDLE_EV(handler)                                                \
-  if (_conky_ev_handle_##handler(surface, display, ev, consumed, cookie)) \
-  return true
+#define HANDLE_EV(event)                                                   \
+  if (handle_event<x_event_handler::event>(surface, display, ev, consumed, \
+                                           cookie)) {                      \
+    return true;                                                           \
+  }
```
Caellian, line 722: Can I keep a macro here?

It's right next to invocation, won't be called anywhere else, and quarters the code below. I would use a constexpr for loop to loop over all `x_event_handler` values, but those don't exist in C++.

src/display-x11.cc
Caellian: Replaced with a template - it's more readable as it reduces the need for `#else`.

I used macros to simplify changing handler arguments (one source of truth), but I guess those won't be changing that often.

Issue #1836, opened 2024-04-18
[Bug]: conky doesn't compile on MacOS with default flags


Caellian:
### Discussed in https://github.com/brndnmtthws/conky/discussions/1835

<div type='discussions-op-text'>

<sup>Originally posted by **Buitragox** April 18, 2024</sup>
I am trying to compile the source code following the [instructions](https://github.com/brndnmtthws/conky/wiki/Installation) page. I managed to make the `cmake ..` step successful, but the `make install` keeps failing.

```
[ 17%] Building CXX object src/CMakeFiles/conky.dir/display-x11.cc.o
/Users/jhbuitrago/Study/conky/src/display-x11.cc:765:14: error: no member named 'xi_opcode' in 'conky_x11_window'
  if (window.xi_opcode != 0) return true;  // handled by mouse_input already
      ~~~~~~ ^
/Users/jhbuitrago/Study/conky/src/display-x11.cc:767:55: error: no member named 'x' in 'conky_x11_window'
  bool not_over_conky = ev.xcrossing.x_root <= window.x ||
                                               ~~~~~~ ^
/Users/jhbuitrago/Study/conky/src/display-x11.cc:768:55: error: no member named 'y' in 'conky_x11_window'
                        ev.xcrossing.y_root <= window.y ||
                                               ~~~~~~ ^
/Users/jhbuitrago/Study/conky/src/display-x11.cc:769:55: error: no member named 'x' in 'conky_x11_window'
                        ev.xcrossing.x_root >= window.x + window.width ||
                                               ~~~~~~ ^
/Users/jhbuitrago/Study/conky/src/display-x11.cc:770:55: error: no member named 'y' in 'conky_x11_window'
                        ev.xcrossing.y_root >= window.y + window.height;
                                               ~~~~~~ ^
/Users/jhbuitrago/Study/conky/src/display-x11.cc:775:44: error: use of undeclared identifier 'mouse_event_t'
        ev.xcrossing.type == EnterNotify ? mouse_event_t::AREA_ENTER
                                           ^
/Users/jhbuitrago/Study/conky/src/display-x11.cc:776:44: error: use of undeclared identifier 'mouse_event_t'
                                         : mouse_event_t::AREA_LEAVE,
                                           ^
/Users/jhbuitrago/Study/conky/src/display-x11.cc:975:10: error: no member named 'x' in 'conky_x11_window'
  window.x = x;
  ~~~~~~ ^
[...]

Issue #1822, opened 2024-04-13
[Feature request] Support a small subset of CSS (or CSS-like) rules


rubyFeedback:
I would like to propose that conky supports, in addition to the current way, CSS-like properties. (If this is the case it would be nice to indicate this via the main README).

Not all CSS properties make sense, but things such as colors would be nice; and perhaps more effects if conky supports
these. Would be even fun to see conky support hovering properties, and neon-like glow effects, but I don't want to make this too fancy, as someone has to implement that, and that may take time; and time is a finite resource. So, I would suggest CSS-like properties.

One important part of this proposal is to support external files - here, I mean local files, but something like:

    conky_colours.css

This can also be used on a website, by the way, so we get both: support on the www, AND support within conky itself. People could also more easily share their conky-rc files, e. g. the skeleton where the logic resides, and then shuffle in or out different .css files (or, rather, the SUBSET that conky may support; again I am not saying all of CSS should be supported, only the parts that may make sense for conky).

The benefit here would be that it lowers the entry-threshold for new users; and another benefit is that we can share and re-use the main design (excluding the skeleton, although it would be nice if conky could support HTML widget-like properties too, but for the purpose of THIS issue request, I really only want to focus on CSS-like properties. Someone else would have to propose HTML-like widget support. Buttons would be cool, but here I focus only on CSS).

That's mostly the gist of it. Perhaps ID (identifiers) may have to be supported too, so we could arbitrarily name the conky properties, such as foo1, foo2, foo3, and within the CSS identifier also refer to them e. g. #foo1. Note that I am not necessarily suggesting conky supports such sub-elements either; for a first step it should suffice if conky can work on perhaps simple:

    .pink  { color: pink }

Or something like that. (Or even RGB colours if that can be parsed); and the ability to tag them onto conky-elements, such as "for this header, I want to use green colour and padding 8px and underlined text". Something like that, but usable via CSS properties IN ADDITION to the current way that works. So this is an additional use case, not at all meant to REMOVE the existing way.

Caellian:
I see this idea often suggested on similar projects, but I disagree and here's why:

### CSS is just a different (less powerful) syntax to do the same

CSS isn't Turing complete, Lua is. There's still space to improve Lua things in conky (e.g. supporting change in `conky.config` after initial load). On the other hand, CSS is completely static (by design).

[...]

PR #1559, opened 2023-05-30
Implement vgoto

bi4k8:
# Checklist
- [x] I have described the changes
- [x] I have linked to any relevant GitHub issues, if applicable
- [x] Documentation in `doc/` has been updated
- [x] All new code is licensed under GPLv3

## Description

This implements a vertical `goto`, to match the existing horizontal one. This has been requested by a number of users. Closes #450, and is based in part on the patch there by @easysid.
  
Tested against the following conkyrc:
```
conky.config = {
	out_to_wayland = true,
	use_xft= true,
	font= 'DejaVu Sans:style=medium:size=9'
}
conky.text = [[
foo bar${vgoto 0}${goto 0}text after vgoto

line 1
line 2
another
line 4
line 5
]]
```

This is still a draft because it seems to cause some issues calculating height--with the above conkyrc, on the first iteration, everything looks as expected, but at the beginning of the second iteration, the whole displayed output moves downward, and the bottom line gets cut off. I'll have to debug what's happening there. I have seen similar jumps previously, so it seems more likely that this is exacerbating an existing misbehavior than introducing a completely new one.

PR #1841, opened 2024-04-19
Cleanup build flags, global namespace and includes

Caellian:
# Changes

## Flags
- Switch `BUILD_MOUSE_EVENTS` dependency from `OWN_WINDOW` to `BUILD_X11`.
- Remove `BUILD_XINPUT` dependency on `BUILD_MOUSE_EVENTS` (now `BUILD_X11` directly).

## Refactor
- Removed weird linker magic for display output init and replaced it with simpler (and maybe faster?) template functions that behave the same.
  - This separates display output compilation units so they don't have to be all built _even when their features are disabled_.
  - Replaced `disabled_display_output` classes with log messages.
- Add explicit errors when feature dependent headers get included where they shouldn't.
- Completely remove use of X11 from unrelated code.
- Rename `special_t` to `special_node` and `special_type` to `text_node_t`.
- Added `#undef` for `HANDLE_EV` macro at the end of the function.
- Added `axis_align` for single axis alignment matching.

## Mouse event fixes/improvements
- As flag is now separated from XInput one, conky uses newer XInput instead of basic X11 events.
- Guarded `XI_Motion` which means that building without `BUILD_MOUSE_EVENTS` produces a more performant binary (due to not tracking global (root) mouse movements)

## Documentation
- Add documentation links to sample configs.

# Related issues
- #1836 - has shown that disabling `OWN_WINDOW` doesn't compile
- #1822 - changes to `colour.cc` make it a lot easier to _string_ (huehue) on new color parsers

# Testing

- [x] Tested for inclusion errors with various build flag permutations
  - BUILD_WAYLAND, BUILD_X11, OWN_WINDOW, BUILD_MOUSE_EVENTS, BUILD_XINPUT
- [x] Tested expected behavior for independent XInput handling
- [x] Tested display output registration

# Remaining work

- Demonstrate `include` improvement using visualization


Caellian:
Checked for conflicts, PRs that would require some changes are:
- #1559 - adding a variable without enum scope
- #1251 - registering a backend using the removed linker dependency injection

src/gui.h
```diff
@@ -25,13 +25,104 @@
...
+  MIDDLE_LEFT = 0b1001,
+  MIDDLE_MIDDLE = 0b1010,
+  MIDDLE_RIGHT = 0b1011,
+  BOTTOM_LEFT = 0b1101,
+  BOTTOM_MIDDLE = 0b1110,
+  BOTTOM_RIGHT = 0b1111,
+};
+constexpr uint8_t operator*(alignment index) {
+  return static_cast<uint8_t>(index);
+}
```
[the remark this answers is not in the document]
Caellian: Also, purely matter of taste, `window_type::NORMAL` looks much nicer than `TYPE_NORMAL` (and is more descriptive).

data/conky-dfly-smp.conf
```diff
@@ -36,7 +39,8 @@ conky.config = {
     extra_newline = false,
     own_window = false,
     own_window_class = 'Conky',
-    own_window_type = 'desktop',
+    own_window_type = 'normal',
+    own_window_hints = 'undecorated,sticky,below,skip_taskbar,skip_pager',
```
Caellian, line 43: Changed this to `normal` because it's the type that most commonly works, on WMs/DEs that are most commonly used (with desktops).

src/colour-names.cc
Caellian, line 1: This file is generated by gperf, it might be a good idea to add its generation to build process, but that adds an additional build dependency. I left it as it is, as I don't expect it to change that often. There's a comment at the top with command and arguments used to generate it from `colour-names.gperf.in`.

src/display-output.cc
```diff
@@ -27,36 +27,44 @@
...
-
-}  // namespace
-
-// HACK: force the linker to link all the objects in with test enabled
-extern void init_console_output();
-extern void init_ncurses_output();
-extern void init_file_output();
-extern void init_http_output();
-extern void init_x11_output();
-extern void init_wayland_output();
```
Caellian, line 59: And HACK gets replaced by [explicit template specialization](https://en.cppreference.com/w/cpp/language/template_specialization), which is a well supported language mechanism.

Caellian (commented):
Explanation of changes to display output registration.

src/display-x11.cc
```diff
@@ -324,39 +323,11 @@ bool display_output_x11::main_loop_wait(double t) {
...
-          case MIDDLE_RIGHT: {
-            sidenum = 1;
-            break;
-          }
-
-          case NONE:
-          case MIDDLE_MIDDLE: /* XXX What about these? */;
-        }
-
-        set_struts(sidenum);
```
Caellian, line 359: `sidenum` is an intermediate variable we do matching on later in `set_struts` as well.

src/x11.cc
```diff
@@ -1205,61 +1084,83 @@ void print_desktop_name(struct text_object *obj, char *p,
...
+}
+
 /* reserve window manager space */
-void set_struts(int sidenum) {
-  Atom strut;
-  if ((strut = ATOM(_NET_WM_STRUT)) != None) {
+void set_struts(alignment align) {
+  // Middle and none align don't have least significant bit set.
+  // Ensures either vertical or horizontal axis are start/end
+  if ((*align & 0b0101) == 0) return;
```
Caellian, line 1110: I added a check to verify `alignement` is actually valid, i.e. would reserve space on one of 4 screen sides.

This is a cheap check (no jumps), and works because both left/top and right/right have least significant bit set to `1`, while none and middle don't. `alignment` is `vertical << 2 + horizontal`, so we can just check `0x0101` pattern to see whether _either_ is set.

src/x11.cc
```diff
@@ -1205,61 +1084,83 @@ void print_desktop_name(struct text_object *obj, char *p,
...
+
+  Atom strut = ATOM(_NET_WM_STRUT);
+  if (strut != None) {
     /* reserve space at left, right, top, bottom */
-    signed long sizes[12] = {0};
+    uint32_t sizes[*x11_strut::COUNT] = {0};
     int i;
 
-    /* define strut depth */
-    switch (sidenum) {
```
Caellian, line 1217: Here, instead of matching again on intermediate sidenum, we can just match actual alignment values. This is clearer, **and** generates much less assembly (though a lot more verbose).

src/x11.cc
```diff
@@ -1205,61 +1084,83 @@ void print_desktop_name(struct text_object *obj, char *p,
...
-    for (i = 0; i < 12; i++) {
-      if (sizes[i] < 0) {
-        sizes[i] = 0;
-      } else {
-        if (i <= 1 || i >= 8) {
-          if (sizes[i] > display_width) { sizes[i] = display_width; }
-        } else {
-          if (sizes[i] > display_height) { sizes[i] = display_height; }
-        }
-      }
```
Caellian, line 1255: We can also skip this loop (12 jumps) entirely because we've already handled constraints for each side separately.

src/x11.cc
```diff
@@ -1205,61 +1084,83 @@ void print_desktop_name(struct text_object *obj, char *p,
...
+            std::min(window.x + window.width, display_width);
+        sizes[*x11_strut::LEFT_START_Y] = window.y;
+        sizes[*x11_strut::LEFT_END_Y] =
+            std::min(window.y + window.height, display_height);
+      case axis_align::END:
+        sizes[*x11_strut::RIGHT] =
+            window.x < display_width ? display_width - window.x : 0;
+        sizes[*x11_strut::RIGHT_START_Y] = window.y;
+        sizes[*x11_strut::RIGHT_END_Y] =
+            std::min(window.y + window.height, display_height);
```
Caellian, line 1151: In effect, we only have one jump on the value.

I checked with godbolt,
- `window.x < display_width ? display_width - window.x : 0` turns into a saturating sub instruction on achitectures where it's supported.
- `std::min(window.y + window.height, display_height)` doesn't generate jumps either AFAICT, allows the compiler to better optimize the code than clamping in a loop later.

The only downside is that it looks like there's a lot more going on in the final code, but it's just that enum names take up a lot of space. That's IMO better because I couldn't get what `5 + (sidenum * 2)` is mapping to even while looking at variant values.

src/x11.cc
Caellian: This check avoids calls to X11 and jumps if the `alignment` doesn't do anything.

src/gui.h
```diff
@@ -25,13 +25,104 @@
...
+  /// Top or left alignment
+  START = 0b01,
+  /// Middle alignment
+  MIDDLE = 0b10,
+  /// Bottom or right alignment
+  END = 0b11,
+};
+constexpr uint8_t operator*(axis_align index) {
+  return static_cast<uint8_t>(index);
+}
```
Caellian, line 51: I added `axis_align` to simplify matching.

Caellian (commented):
Explanation of changes to `alignment` code and `set_struts`.

Caellian:
Good to see labeler changes working :laughing:

Removed color parsing into a stacked PR: #1848

Caellian:
Moved alignment changes into #1849.

Caellian:
Did something... for visualization of includes, does need a lot more work though. Not sure if it will be useful before finishing this commit, but I'll be able to create _some_ visualization of it at least.

include_graph

Caellian:
> There's a lot in this PR, but overall I like it :)

Thank you. I'll maybe tweak it a tiny bit more before merging (some added checks aren't needed). I wanted to change those things (like enums) all at once so it's just a single merge for others.

Caellian:
There's a linking error with a specific build flag configuration, holding this off until I resolve it. Maybe it was there before, maybe it wasn't.

Finished the [tool for generating include graphs](https://github.com/Caellian/cdepgraph) for now, I'll try to connect it up to treeshaker in order to collect information about function exports and how they're being used and then check why the functions aren't linking (even though cc file is enabled and header included). Some time soon... (up to a month max.)

PR #1848, opened 2024-04-20
Use perfect hash (gperf) for parsing color names

Caellian:
Replaces X11 color name lookup with something that's easier to maintain.

# Performance

Possibly faster than previous implementation - previous used binary search (`O(log n)`) on sorted color names, new uses a perfect hash (`O(1)`).

# Extensibility

The main selling point of this PR is that it replaces code that was hard to extend with something where people can easily add their own colors.

While writing this PR, I realized that the previous could could've just been simplified to achieve same extensibility, but I guess perf. hash scales better and doesn't require colors to be sorted in advance (though that can be automated).

# Testing

Tested by preexisting tests.


src/colours.cc
```diff
@@ -50,58 +42,15 @@ Colour Colour::from_argb32(uint32_t argb) {
   return out;
 }
 
-#ifdef BUILD_X11
-unsigned long Colour::to_x11_color(Display *display, int screen,
-                                   bool transparency, bool premultiply) {
```
Caellian, line 55: Moved to `x11-color.cc` where previous matching function was.

brndnmtthws (approved):

LinuxOnTheDesktop:
This commit sounds good (so, thank you!) but just what does it mean for the writer of conkies? How should we give colour names now? Is the new system backwards compatible?

PR #1849, opened 2024-04-20
set_struts improvements

Caellian:
This PR modifies `set_struts` function to accept `alignment` directly instead of using an intermediate value.

This commit makes code easier to follow and improves performance a tiny bit because it inlines a loop.

# Testing

- [x] bottom and middle alignment seem to not be working on Openbox. Need to be tested against `main`, it's possible this is a bug in Openbox because set values seem fine.
  - Used incorrect type for values.

### DIFF
--- a/src/display-x11.cc
+++ b/src/display-x11.cc
@@ -313,37 +313,8 @@ bool display_output_x11::main_loop_wait(double t) {
 
       /* update struts */
       if ((changed != 0) && own_window_type.get(*state) == window_type::PANEL) {
-        int sidenum = -1;
-
         NORM_ERR("defining struts");
-
-        alignment align = text_alignment.get(*state);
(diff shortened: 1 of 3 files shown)


Caellian:
@brndnmtthws Try re-running CI when you get to this, I tested it and it should work. CI is failing on package install for some reason.

brndnmtthws (approved):
PR looks good, the CI issue looks temporary.

brndnmtthws:
Seems to be a known issue, just need to wait for it to be resolved: https://www.githubstatus.com/incidents/7txcg0j03kkg

My guess is that someone screwed up when they were trying to fix [this problem](https://www.bleepingcomputer.com/news/security/github-comments-abused-to-push-malware-via-microsoft-repo-urls/) :)

Issue #1852, opened 2024-04-21
[Bug]: Querying windows every event causes significant CPU usage


Caellian:
> With fluxbox as WM, a right click on the root window leads to 3 second delay waiting for the WM right-click menu to appear, with 100% CPU in the meantime.
I bisected the problem to f4b3229fc34e1acc9a2e814abee1d886049a4560

_Originally posted by @jborme in https://github.com/brndnmtthws/conky/issues/1821#issuecomment-2068056821_

Workaround: disable `BUILD_XINPUT` flag.

lineage-of-roots:
I tried on Fluxbox. I can confirm that there is a significant delay on the hover events. And Conky's cpu usage jumps really high as well.  
The Fluxbox menu delay is not a problem by itself. It seems its the accumulated hover events before the click that are causing the problem.
By leaving the mouse cursor on Conky, right clicking and pressing escape to remove the menu...and repeat, I saw that the menu reappears without delay.
The menu has no delay if you wait about 3 seconds before right clicking.
I will try later on other wms

````
  return propagate_xinput_event(
        reinterpret_cast<const conky::xi_event_data *>(cookie));

````

Seems like this is where it starts going wrong.

Caellian:
Hmm... Most work happens while trying to figure out which window to send the event to.

Which also, if that's the problematic part, won't be fixed by disabling BUILD_XINPUT.

Mouse click is the simplest event to handle as well, so it doesn't make sense for it to be held up there.

Personal note while fixing this: Added conky atoms for tweaking XInout behavior might not be needed in most cases - normal windows get correct relative/absolute values and only other types (myb root as well) seem to always get absolute values on Open box (check other WMs, might be an Openbox thing).

lineage-of-roots:
Update:
So I tested across many Window Managers and a couple of Desktop Environments
There is a  significant jump in cpu usage when hovering on conky. It's consistent across all WMs and DEs, except one outlier that I found.

Conky also makes Xorg jump really high in cpu usage. They both jump together.


The outlier I found was jwm.  It does not seem care about the hover events. No cpu usage jump and  the menu appears normally as well. I don't yet know what it's doing different. Might be a clue.

Fluxbox seems like the only one with more delay in the hover events, therefore making it look like the menu is delayed. 


I testsed on MATE and KDE Plasma, same high cpu usage on hover events.
I am guessing other DEs like Gnome and Cinnamon will also see jumps in cpu usage.  

[...]

Issue #1528, opened 2023-05-09
[Bug]: `use_xft = true` breaks window sizing


simotek:
### What happened?

Window sizing is smaller when `use_xft = true` is set then when it is not, if a window is a forced size with min_width and max_width being the same then the window size will be slightly smaller then that.

This first image is with `use_xft=true` the colored temperature shape should be centered 
![shot-2023-05-09_18-54-42](https://user-images.githubusercontent.com/3827011/237056075-65368151-1bc5-43a1-82fd-a5ff2bcef135.jpg)

This second shot is showing everything working properly with it disabled
![shot-2023-05-09_18-55-23](https://user-images.githubusercontent.com/3827011/237056096-ca510c59-e5b0-44e4-af91-85863af497da.jpg)

Issue is seen running a few day old version of conky from git on openSUSE Tumbleweed

### Version

git

### Which OS/distro are you seeing the problem on?

Linux (other)

### Conky config

```Lua
https://github.com/philer/polycore  `/opt/conky/bin/conky -c  conkyrc.lua` I had to comment out line 45 of layout.lua to stop a crash as I don't have gpu support.
```

Caellian:
My guess was that font size (glyph width) gets used for window width calculation somehow, but that's not the case.

Closing for now as I can't reproduce (Openbox) with:
```
conky.config = {
  own_window = true,
  own_window_type = 'override',
  own_window_hints = 'undecorated,sticky,skip_taskbar,skip_pager',
  minimum_width = 500,
  maximum_width = 500,
  use_xft = true,
  font = 'Ubuntu:pixelsize=200',
}
conky.text = [[
  $conky_version
  $conky_version
  $conky_version
]]
```

Not sure if I'm missing something, but `use_xft` doesn't have any interaction with window size (besides making it larger if maximum_width not set).

Reopen the issue if you still have it and add which WM you're using.

Caellian:
I believe I found what causes this.

In configure event handler there's this:
```
int mw = surface->dpi_scale(maximum_width.get(*state));
if (text_width > mw && mw > 0) { text_width = mw; }
```
I don't think expected width in pixels should be affected by DPI. So even though it doesn't affect min_width, by affecting max_width it shrinks the window.

`surface->dpi_scale` is defined as:
```
int display_output_x11::dpi_scale(int value) {
#if defined(BUILD_XFT)
  if (use_xft.get(*state) && xft_dpi > 0) {
[...]

PR #1862, opened 2024-04-25
Introduce geometry primitives

Caellian:
# Changes

This PR introduces `point` and `rect` primitives and replaces previous value pairs with them. It also replaces output-wayland.cc `rectangle` as it's effectively doing the same.

This aims to make future changes more consistent and to reduce number of global variables by grouping them. As a side bonus, it allows us to optimize some common operations a little by relying on [Vc](https://vcdevel.github.io/Vc-1.4.4/) to make the compiler more aware of available opportunities to parallelize computation. Just grouping things results in significant reduction in number of generated move instructions:

image

## Details

- [x] Added `vec` and `rect` structures.
- [x] Added [Vc](https://vcdevel.github.io/Vc-1.4.4/) dependency for better SIMD optimization of common vector  operations.
  - This was previously done through Eigen, but I removed Eigen as 90% of its functionality wouldn't ever be used by Conky so it only added to maintenance burdain.


Caellian:
@brndnmtthws Is it okay to add Eigen as dependency? I can do without it, but it might optimize some instructions. Alternatively, I could make it optional and apply SIMD only if present.

src/display-x11.cc
```diff
@@ -648,26 +646,28 @@ bool handle_event<x_event_handler::CONFIGURE>(
...
+      text_width = new_text_size.x();
+      text_height = new_text_size.y();
 
-      text_width = window.width - 2 * border_total;
-      text_height = window.height - 2 * border_total;
-      int mw = surface->dpi_scale(maximum_width.get(*state));
-      if (text_width > mw && mw > 0) { text_width = mw; }
+      // don't apply dpi scaling to max pixel size
+      int mw = std::max(maximum_width.get(*state), 0);
+      if (text_width > mw) { text_width = mw; }
```
Caellian, line 661: Fixes #1528 hopefully

src/x11.cc
```diff
@@ -780,14 +781,14 @@ void x11_init_window(lua::state &l, bool own) {
...
+
   int64_t input_mask = ExposureMask | PropertyChangeMask;
 #ifdef OWN_WINDOW
-  if (own_window.get(l)) {
-    input_mask |= StructureNotifyMask;
 #if !defined(BUILD_XINPUT)
-    input_mask |= ButtonPressMask | ButtonReleaseMask;
+  if (own_window.get(l)) { input_mask |= ButtonPressMask | ButtonReleaseMask; }
 #endif
-  }
```
Caellian, line 790: Did this too soon. Anyway, this is supposed to add window change watching in one of future commits to update window bounds caching and fix #1852

Caellian (commented):
Last commit is expected be functionally equivalent to original

brndnmtthws:
> @brndnmtthws Is it okay to add Eigen as dependency? I can do without it, but it might optimize some instructions. Alternatively, I could make it optional and apply SIMD only if present.
> 
> EDIT: We do surprisingly many vector computations while drawing components. I don't use majority of Eigen though. It might be better to simply add a way to provide optimized assembly for certain architectures and data types. 🤷‍♂️ And then someone can easily add optimizations later if they want to.

I haven't used Eigen, but I suppose if we can include it as a 3rd party dependency in the source tree (like we do with tolua++ and catch2), that would be fine. Hopefully it doesn't increase compile times too much. I would put it alongside toluapp in the `3rdparty/` dir.

brndnmtthws:
I pushed a branch here to demonstrate what I meant by including it in the `3rdparty/` dir: https://github.com/brndnmtthws/conky/tree/add-eigen-3rdparty

You should be able to just cherry-pick e84f7b65635c7419742206b2928205759928605b from that branch, but remove the demo code from `main.cc`.

Caellian:
> I pushed a branch here to demonstrate what I meant by including it in the `3rdparty/` dir: https://github.com/brndnmtthws/conky/tree/add-eigen-3rdparty
> 
> You should be able to just cherry-pick [e84f7b6](https://github.com/brndnmtthws/conky/commit/e84f7b65635c7419742206b2928205759928605b) from that branch, but remove the demo code from `main.cc`.

I added it as a submodule and made CMake update submodules during configure. IMO this is better because:
- we're not tracking entire eigen sources in our history
- it's much easier to update it
- support for BLAS and LaPACK can be configured while building conky
- pulling all the sources would make this PR huge

 I got it to build correctly, just have to figure out how to include the target as dependency.

brndnmtthws:
> I added it as a submodule and made CMake update submodules during configure. IMO this is better because:
> 
>     * we're not tracking entire eigen sources in our history
> 
>     * it's much easier to update it
> 
>     * support for BLAS and LaPACK can be configured while building conky
> 
>     * pulling all the sources would make this PR huge
> 
>     * source authenticity doesn't have to be validated if someone's feeling paranoid 😄
> 
> 
> I got it to build correctly (takes ~1s to configure; build time is the same), just have to figure out how to include the target as dependency.
> 
> I still got little work to do on this PR anyway, so it will probably be done in a day or so.

I would prefer not to introduce a submodule. Submodules are usually more trouble than their worth. It would be much easier to just pick a specific release and include the sources we need. There's also subtrees, but I'm guessing those aren't worth the effort either.

Caellian:
> Submodules are usually more trouble than they're worth.

I think they're the most ergonomic solution in this case because we're not going to be modifying eigen sources, and build works even with shallow clone due to the [script](https://gist.github.com/Caellian/9f92506fa24df111729589d392e84553) I added. So no one would need to interact with the submodule unless it's a version bump, which would've been much simpler than reapplying modifications on updates. I personally see no discrete reason to not use a submodule in this case.

brndnmtthws:
> I think they're the most ergonomic solution in this case because we're not going to be modifying eigen sources, and build works even with shallow clone due to the [script](https://gist.github.com/Caellian/9f92506fa24df111729589d392e84553) I added. So no one would need to interact with the submodule unless it's a version bump, which would've been much simpler than reapplying modifications on updates. I personally see no discrete reason to not use a submodule in this case.

I understand your perspective, but it just isn't necessary to introduce all that complexity for something that doesn't require it. If we want to upgrade Eigen in the future, it's as simple as updating the source files, but for the most part we can treat 3rd party code as immutable, and layering everything on top of submodules just introduces unneeded complexity. I don't want downstream consumers or package maintainers to have to deal with all the quirks and edge cases that will inevitably be introduced.

brndnmtthws (approved):
👍 , seems like Vc is a better choice than Eigen for our needs.

This is a pretty big change, so it's probably going to have some unintended side effects. There are a few things I'd refactor, but probably better to do it in a separate PR later.

Caellian:
> There are a few things I'd refactor, but probably better to do it in a separate PR later.

Same here :laughing:, I just tried adding semantics to rectangle via tempates (absolute vs. sized) and realized there's a lot that could be improved. C++ doesn't have traits as Rust does so it's somewhat tricky to get things like swizzling to layer properly and still allow the compiler to optimize code.

I'll probably address those issues at some point, but it's better to have _something_ and I don't want to keep this PR waiting for too long as it will be painful to update.

PR #1863, opened 2024-04-25
Re-enable RSS in AppImage, also enable in Nix flake

### LINKED ISSUE #1861: [Bug]: conky: unknown variable '$rss'
### What happened?

I tried to use `rss` in an existing working conky configuration, but the output just shows `${rss}`. In the log, conky said `unknown variable '$rss'`. I noticed that on my Ubuntu 22.04 installation there is a rather outdated version of conky ("conky 1.12.2 compiled 2022-02-23 for Linux x86_64"), so I installed the newest AppImage but even with that, the message is the same.

### Version

1.12.2 and 1.20.2

Ubuntu

### Conky config

```Lua
conky -t '${rss https://www.heise.de/security/rss/news.rdf 60 item_titles 10}' -o -u 30 -x 10
```

### Stack trace

_No response_

### Relevant log output

```Shell
conky: desktop window (0x800015) is subwindow of root window (0x3cc)
conky: window type - desktop
conky: drawing to created window (0x1a00002)
conky: unknown variable '$rss'
conky: FOUND: console
conky: FOUND: ncurses
conky: FOUND: file
conky: FOUND: http
conky: FOUND: x11
conky: FOUND: wayland
conky: 'ubuntu' wayland session running 'ubuntu:GNOME' destop
conky: can't open /sys/class/power_supply/BAT0/uevent: Datei oder Verzeichnis nicht gefunden
conky: can't open /proc/acpi/battery/BAT0/state: Datei oder Verzeichnis nicht gefunden
conky: can't open /proc/apm: Datei oder Verzeichnis nicht gefunden
```

brndnmtthws:
The behaviour of the RSS flag changed in 4936c74, previously it would enable cURL support when it was switched on, but now you must pass both BUILD_RSS=ON and BUILD_CURL=ON to CMake at configure time.

This fixes #1861.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -56,6 +56,8 @@
               "-DBUILD_LUA_CAIRO=ON"
               "-DBUILD_LUA_IMLIB2=ON"
               "-DBUILD_LUA_RSVG=ON"
+              "-DBUILD_RSS=ON"
+              "-DBUILD_CURL=ON"
             ];
             nativeBuildInputs = [
               clang_16
@@ -69,10 +71,12 @@
             buildInputs =
               [
                 cairo
+                curl
                 freetype
                 gettext
                 imlib2
                 librsvg
+                libxml2
                 llvmPackages_16.libcxx
                 llvmPackages_16.libcxxabi
                 lua5_4
--- a/.github/workflows/build-and-test-macos.yaml
+++ b/.github/workflows/build-and-test-macos.yaml
@@ -33,20 +33,22 @@ jobs:
       - name: Install dependencies
         run: |
           brew install --force \
-            cmake \
-            freetype \
-            gettext \
-            imlib2 \
-            gperf \
-            lcov \
-            librsvg \
-            libxft \
-            libxinerama \
-            libxfixes \
-            libxi \
-            lua \
-            ninja \
-            pkg-config \
+            cmake              \
+            curl               \
+            freetype           \
+            gettext            \
+            gperf              \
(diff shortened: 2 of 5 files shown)


flake.nix
```diff
@@ -69,10 +71,12 @@
             buildInputs =
               [
                 cairo
+                curl
```
Caellian, line 74: This other [curl based thing](https://github.com/NixOS/nixpkgs/blob/dd37924974b9202f8226ed5d74a252a9785aedf8/pkgs/applications/networking/flent/http-getter.nix#L15) uses `curl` as well. So this seems fine.

flake.nix
```diff
@@ -69,10 +71,12 @@
...
                 freetype
                 gettext
                 imlib2
                 librsvg
+                libxml2
```
Caellian, line 79: Seems like appropriate include for [libxml2](https://search.nixos.org/packages?channel=23.11&show=libxml2&from=0&size=50&sort=relevance&type=packages&query=libxml2).

Caellian (commented):
This seems like it should be ok.

Notes:
- `build-and-test-macos.yaml` doesn't doesn't enable RSS, some people might be using as reference for building on MacOS.
- `codeql.yml` has duplicate `libcurl4-gnutls-dev` dependency.

Caellian (approved):
Oh, I didn't select approve. :smile:

Caellian (approved):
Good to merge.

PR #1877, opened 2024-04-30
Remove dpi scaling of maximum width

### LINKED ISSUE #1528: [Bug]: `use_xft = true` breaks window sizing

Caellian:
This PR removes `dpi_scale`ing of maximum width setting.

This seems incorrect because I expect maximum width to use absolute pixel value, otherwise meaning of `maximum_width` setting changes with DPI:

image
<sub>NOTE: <em>Height difference is caused by font height, conky height can't be limited.</em></sub>

This fixes #1528.

This PR also fixes a regression introduced in #1841, where I incorrectly assumed templated base class function would be overriden by display outputs so DPI scaling was completely disabled.

# Other changes

This PR also:
- sets `workspace` dimensions from WL `output_geometry` because workspace is used for some calculations in `conky.cc`.
  - this requires more work for proper support as noted in comments, but it should make Wayland less broken.
- adds conky xdg dir and script parent directory to Lua path so that scripts don't have to update it manually.

# Testing

- [x] Use mock Xft dpi value and see whether example from #1528 works better with/without this PR.

### DIFF
--- a/src/conky.cc
+++ b/src/conky.cc
@@ -864,7 +864,7 @@ void update_text_area() {
     if (text_height < dpi_scale(minimum_height.get(*state))) {
       text_height = dpi_scale(minimum_height.get(*state));
     }
-    int mw = dpi_scale(maximum_width.get(*state));
+    int mw = maximum_width.get(*state);
     if (text_width > mw && mw > 0) { text_width = mw; }
   }
 
@@ -987,7 +987,7 @@ static int text_size_updater(char *s, int special_index) {
   w += get_string_width(s);
 
   if (w > text_width) { text_width = w; }
-  int mw = dpi_scale(maximum_width.get(*state));
+  int mw = maximum_width.get(*state);
   if (text_width > mw && mw > 0) { text_width = mw; }
 
   text_height += last_font_height;
--- a/src/conky.cc
+++ b/src/conky.cc
@@ -1052,7 +1052,7 @@ static void draw_string(const char *s) {
   }
 #ifdef BUILD_GUI
   if (display_output() && display_output()->graphical()) {
-    int mw = display_output()->dpi_scale(maximum_width.get(*state));
+    int mw = maximum_width.get(*state);
     if (text_width == mw) {
       /* this means the text is probably pushing the limit,
        * so we'll chop it */
@@ -1122,7 +1122,7 @@ int draw_each_line_inner(char *s, int special_index, int last_special_applied) {
 
 #ifdef BUILD_GUI
   if (display_output() && display_output()->graphical()) {
-    mw = display_output()->dpi_scale(maximum_width.get(*state));
+    mw = maximum_width.get(*state);
     font_h = font_height();
     cur_y += font_ascent();
   }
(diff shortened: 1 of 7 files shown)


src/conky.cc
```diff
@@ -1961,6 +1962,40 @@ void load_config_file() {
...
+    l.getfield(-1, "path");
+
+    auto path = l.tostring(-1);
+    path.append(path_ext);
+    l.pop();
+    l.pushstring(path.c_str());
+
+    l.setfield(-2, "path");
+    l.pop();
+  }
```
Caellian, line 1997: Adding `$HOME/.config/conky` and config parent directory to lua path before loading the script.

src/display-x11.cc
```diff
@@ -945,18 +945,14 @@ void display_output_x11::move_win(int x, int y) {
...
 
-const size_t PIXELS_PER_INCH = 96;
-template <typename T, typename>
-T display_output_x11::dpi_scale(T value) {
-#if defined(BUILD_XFT)
+const float PIXELS_PER_INCH = 96.0;
+float display_output_x11::get_dpi_scale() {
+#ifdef BUILD_XFT
   if (use_xft.get(*state) && xft_dpi > 0) {
-    return (value * xft_dpi + (value > 0 ? 48 : -48)) / PIXELS_PER_INCH;
```
Caellian, line 953: I believe `(value > 0 ? 48 : -48)` was rounding the result for `int` values before, so I replaced it with `ceil` and `floor` in new code.

src/display-wayland.cc
```diff
@@ -300,7 +300,17 @@ static void output_geometry(void *data, struct wl_output *wl_output, int32_t x,
...
+                            int32_t transform) {
+  // TODO: Add support for proper output management through:
+  // - xdg-output-unstable-v1
+  // Maybe also support (if XDG protocol not reported):
+  // - kde-output-management(-v2)
+  // - wlr-output-management-unstable-v1
+  workarea[0] = x;  // TODO: use xdg_output.logical_position
+  workarea[1] = y;
+  workarea[2] = physical_width;
+  workarea[3] = physical_height;
```
Caellian, line 312: This is the workarea change.

src/conky.cc
```diff
@@ -275,7 +276,7 @@ int text_width = 1,
 struct information info;
 
 /* path to config file */
-std::string current_config;
+std::filesystem::path current_config;
```
brndnmtthws, line 279: Nice

brndnmtthws (approved):
Seems reasonable.

PR #1870, opened 2024-04-28
Require gperf at build-time

brndnmtthws:
Rather than allowing the build to continue without gperf, we should fail with an error so that the colour behaviour does not change in a backward incompatible manner. The old colour behaviour should continue to work going forward.

This resolves #1868.

PR #1880, opened 2024-05-01
Add BUILD_COLOUR_NAME_MAP build flag

brndnmtthws:
You can now specify `BUILD_COLOUR_NAME_MAP=OFF` to disable building the gperf-based colour map, which also removes the requirement for gperf at build time.

This adds back the colour stub feature removed in #1870, however this time it retains backward compatibility with the previous colour name handling behaviour by default.

### DIFF
--- a/src/colour-names-stub.hh
+++ b/src/colour-names-stub.hh
@@ -1,0 +1,28 @@
+/*
+ * To generate colour-names.hh, you must have gperf installed during build.
+ * This is a dummy implementation for builds without gperf.
+ * Color name matching will always return null (i.e. no match).
+ */
+
+#pragma once
+
+#include <cstdint>
+
+#include "logging.h"
+
+struct rgb {
+  const char *name;
+  uint8_t red;
+  uint8_t green;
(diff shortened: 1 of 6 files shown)


src/colour-names-stub.hh
```diff
@@ -0,0 +1,28 @@
...
+  uint8_t blue;
+};
+
+class color_name_hash {
+ public:
+  static const struct rgb *in_word_set(const char *str, size_t len);
+};
+
+const struct rgb *color_name_hash::in_word_set(const char *str, size_t len) {
+  DBGP("color parsing not supported");
```
Caellian, line 26: I guess the only thing that I'm not happy with in my initial impl is that this should've been `NORM_ERR` and placed last (after hex parser).
That would've avoided confusion in that one issue.

But given that the option is enabled by default, it's fine I guess. Maybe make it `DBGP2` because it will spam a lot.

Caellian (approved):

Caellian (approved):

PR #1903, opened 2024-05-08
Refactor the CMake 3rdparty/Vc-related bits

brndnmtthws:
* Include the 3rdparty/ dir from top-level
* Set includes/libs at top-level
* Remove the inner CMake version requirement
* Fix deprecated use of CMake's exec_program()

### DIFF
--- a/3rdparty/CMakeLists.txt
+++ b/3rdparty/CMakeLists.txt
@@ -1,7 +1,4 @@
-# Version 1.4.4
+# Vc version 1.4.4
 add_subdirectory(Vc)
-set(conky_libs ${conky_libs} Vc)
-set(conky_includes ${conky_includes} "${CMAKE_CURRENT_SOURCE_DIR}/Vc")
 
-set(conky_libs ${conky_libs} PARENT_SCOPE)
-set(conky_includes ${conky_includes} PARENT_SCOPE)
+add_subdirectory(toluapp)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -43,10 +43,11 @@ configure_file(${CMAKE_MODULE_PATH}/build.h.in ${CMAKE_BINARY_DIR}/build.h)
 
 set(conky_sources ${CMAKE_BINARY_DIR}/config.h ${CMAKE_BINARY_DIR}/build.h)
 
-# Include 3rdparty toluapp
+# Include 3rdparty code
 add_subdirectory(3rdparty)
-add_subdirectory(3rdparty/toluapp)
 set(conky_libs ${conky_libs} toluapp_lib_static)
+set(conky_libs ${conky_libs} Vc)
+set(conky_includes ${conky_includes} "${CMAKE_SOURCE_DIR}/3rdparty/Vc")
 
 # Finally, add some code
 add_subdirectory(lua)
--- a/3rdparty/Vc/CMakeLists.txt
+++ b/3rdparty/Vc/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.0)
+# cmake_minimum_required(VERSION 3.0)
 
 cmake_policy(SET CMP0028 NEW) # Double colon in target name means ALIAS or IMPORTED target.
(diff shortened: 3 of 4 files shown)


3rdparty/CMakeLists.txt
```diff
@@ -1,7 +1,4 @@
-# Version 1.4.4
+# Vc version 1.4.4
 add_subdirectory(Vc)
-set(conky_libs ${conky_libs} Vc)
-set(conky_includes ${conky_includes} "${CMAKE_CURRENT_SOURCE_DIR}/Vc")
 
-set(conky_libs ${conky_libs} PARENT_SCOPE)
-set(conky_includes ${conky_includes} PARENT_SCOPE)
```
Caellian, line 7: Why not keep 3rdparty related build configuration in 3rdparty/CMakeLists.txt?

CMakeLists.txt
```diff
@@ -43,10 +43,11 @@ configure_file(${CMAKE_MODULE_PATH}/build.h.in ${CMAKE_BINARY_DIR}/build.h)
 
 set(conky_sources ${CMAKE_BINARY_DIR}/config.h ${CMAKE_BINARY_DIR}/build.h)
 
-# Include 3rdparty toluapp
+# Include 3rdparty code
 add_subdirectory(3rdparty)
-add_subdirectory(3rdparty/toluapp)
 set(conky_libs ${conky_libs} toluapp_lib_static)
+set(conky_libs ${conky_libs} Vc)
+set(conky_includes ${conky_includes} "${CMAKE_SOURCE_DIR}/3rdparty/Vc")
```
Caellian, line 50: And move toluapp from here there?

It's simpler to have to modify a single file when adding/removing dependencies than doing so in 2 places.
So I'd change one of the following:
- Put all thirdparty configuration in thirdparty to avoid cluttering main CMake file.
  - Some `add_subdirectory` in the future might require 10 `set`s before and huge `mark_as_advanced` as Eigen did. I personally prefer [minimally invasive approach](https://github.com/pytorch/glow/tree/master/thirdparty) because its a lot less work to update sources that way.
- Completely remove `3rdparty/CMakeLists.txt`
- Something in-between those two options is to `set(vc_includes "${CMAKE_CURRENT_SOURCE_DIR}/Vc" PARENT_SCOPE)` in `3rdparty/CMakeLists.txt` and then do `set(conky_includes ${conky_includes} ${vc_includes})`, for both Vc and toluapp.
  - This separates details on where 3rdparty libraries keep their libs and includes from conky configuration.

3rdparty/CMakeLists.txt
brndnmtthws: I dislike this because it feels like the dependency is going in the wrong direction (from lower -> upper scope). The child shouldn't know about the parent (i.e., you should be able to take the code out of `3rdparty/` and move it elsewhere and it should work all the same).

CMakeLists.txt
brndnmtthws: I'd like to maintain the hierarchy. I'm okay with removing `3rdparty/CMakeLists.txt` entirely, but I prefer keeping the paths within `3rdparty/` isolated such that they don't depend paths higher up the chain.

brndnmtthws: Also this isn't a hill I'm willing to die on, more a matter of preference.

Caellian (approved):

CMakeLists.txt
Caellian: > I'd like to maintain the hierarchy. I'm okay with removing 3rdparty/CMakeLists.txt entirely, but I prefer keeping the paths within 3rdparty/ isolated such that they don't depend paths higher up the chain.

I get what you mean.

> Also this isn't a hill I'm willing to die on, more a matter of preference.

I die on every hill, even when wrong sometimes, so sorry if I'm being too nit-picky.

After looking at the comment again, I noticed how the tone might've come off wrong - the two comments are a single question but I posted them in a thread to reference multiple parts of the code, didn't mean it as "Why this? AND YOU DID THIS?", so I'm sorry if you received it that way, that wasn't my intention. I should've use `...see next...` at the end or something like that.

PR #1909, opened 2024-05-13
Disable Vc_ENABLE_INSTALL by default

Vincent-C:
# Checklist
- [] I have linked to any relevant GitHub issues, if applicable
- [] Documentation in `doc/` has been updated
- [] All new code is licensed under GPLv3

## Description

Debian maintainer here, forwarded from [Debian BTS #1071006](https://bugs.debian.org/1071006)

By default, the newly vendored Vc library will be installed alongside conky itself, e.g. [all of the headers and the cmake files that ship with Vc](https://web.archive.org/web/20240513084923/https://packages.debian.org/sid/amd64/conky-all/filelist). This is unnecessary and even undesired if the user already had Vc installed from other sources, e.g. `libvc-dev` in Debian/Ubuntu. I can turn this off with `-DVc_ENABLE_INSTALL=OFF` but I think this would be a sane default option.

### DIFF
--- a/3rdparty/Vc/CMakeLists.txt
+++ b/3rdparty/Vc/CMakeLists.txt
@@ -187,61 +187,61 @@ target_include_directories(Vc
    $<INSTALL_INTERFACE:include>
    )
 
-option(Vc_ENABLE_INSTALL "Whether to install the library." OFF)
-if (Vc_ENABLE_INSTALL)
-   install(TARGETS Vc EXPORT VcTargets DESTINATION lib${LIB_SUFFIX})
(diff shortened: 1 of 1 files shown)


Caellian:
I believe I should've commented out the installation part of Vc CMake while adding it:
- It's vendored so it might cause other programs to break in the future if we modify functionality and someone overwrites their package installation.
- Conky isn't a library so providing headers is not needed, we just need them at build time.
  - Vc isn't dynamic.

Caellian (approved):
Missed this in #1862.

Caellian:
@Vincent-C let's wait for @brndnmtthws to approve before merging, but I'm pretty sure removing the whole installation part is fine.

brndnmtthws (approved):
Makes sense, sorry we missed this, and thanks for the PR 🙏

PR #1910, opened 2024-05-13
Remove _member_access from geometry

### LINKED ISSUE #1908: [Bug]: segmentation fault upon modification of conkyrc. (Regression?)
### What happened?

I had Conky running (via the Bash script that I use always) and I modified my conkyrc. I saw this:

`108971 Segmentation fault      (core dumped) conky -c "${conk_path[conkyRc]}"`

I am on Linux Mint Cinnamon.

I compiled from source using the latest code from git.

Previously I was using a Conky that I had compiled from source, from git, approximately a week ago - and I think that then I did not have the problem. I believe also that there was a bug report about this or a similar problem - a closed report - but I cannot find that report at present.

### Version

1.21.0-pre-6bf383cf

Linux (other)

### Conky config

_No response_

### Stack trace

```GDB
# Er . .
$ gdb
(gdb) run -c <full-path-of-my-conkyRC>
[. . .]
No executable file specified
[. . .]
(gdb) run conky -c <full-path-of-my-conkyRC>
[. . .]
No executable file specified
```

### Relevant log output

_No response_

Caellian:
It was difficult/too costly ensuring temporary `this` was correct when `vec`s contained in `rect`s were accessed directly.

This PR also adds:
- semantics to `rect` for sized & absolute rectangles, and fixes another (unreported) bug with previous PR (#1862) where I used absolute end positions instead of width/height.
- some more utility functions and setters.

Fixes #1908.

### DIFF
--- a/cmake/ConkyBuildOptions.cmake
+++ b/cmake/ConkyBuildOptions.cmake
@@ -23,19 +23,19 @@ include(DependentOption)
 
 if(NOT CMAKE_BUILD_TYPE)
   if(MAINTAINER_MODE)
-    set(CMAKE_BUILD_TYPE
-        Debug
-        CACHE
-          STRING
-          "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel."
-          FORCE)
+    set(
+      CMAKE_BUILD_TYPE Debug
+      CACHE
+      STRING
+      "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel."
+      FORCE)
   else(MAINTAINER_MODE)
-    set(CMAKE_BUILD_TYPE
-        RelWithDebInfo
-        CACHE
-          STRING
-          "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel."
-          FORCE)
+    set(
+      CMAKE_BUILD_TYPE RelWithDebInfo
+      CACHE
+      STRING
+      "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel."
+      FORCE)
   endif(MAINTAINER_MODE)
 endif(NOT CMAKE_BUILD_TYPE)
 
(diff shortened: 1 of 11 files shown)


src/x11.cc
```diff
@@ -180,8 +180,12 @@ static int x11_error_handler(Display *d, XErrorEvent *err) {
       error_name, reinterpret_cast<uint64_t>(err->display),
       static_cast<int64_t>(err->resourceid), err->serial, code_description);
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wfree-nonheap-object"
+  // *_allocated takes care of avoiding freeing unallocated objects
   if (name_allocated) delete[] error_name;
   if (code_allocated) delete[] code_description;
+#pragma GCC diagnostic pop
```
Caellian, line 188: This produces a (new) warning in GCC 14.1.1

cmake/ConkyBuildOptions.cmake
```diff
@@ -51,7 +51,10 @@ if(MAINTAINER_MODE)
 endif(MAINTAINER_MODE)
 
 # Always use libc++ when compiling w/ clang
-add_compile_options($<$<COMPILE_LANG_AND_ID:CXX,Clang>:-stdlib=libc++>)
+add_compile_options(
+  $<$<COMPILE_LANG_AND_ID:CXX,Clang>:-stdlib=libc++>
+  $<$<COMPILE_LANG_AND_ID:CXX,Clang>:-Wno-unknown-warning-option>
+  $<$<COMPILE_LANG_AND_ID:CXX,GCC>:-Wno-unknown-warning>)
```
Caellian, line 57: Had to add these to avoid warning about new warnings with older compiler versions.

src/template.cc
```diff
@@ -226,7 +226,7 @@ char *find_and_replace_templates(const char *inbuf) {
       outlen += len;
       *o = '\0';
       outbuf = static_cast<char *>(realloc(outbuf, outlen * sizeof(char)));
-      strncat(outbuf, tmpl_out, len);
+      strcat(outbuf, tmpl_out);
```
Caellian, line 229: This was also a warning because the new compiler can tell that `len` is derived from `strlen(tmpl_out)` which is then the same as using `strcat` directly.

src/x11.cc
```diff
@@ -1139,44 +1141,46 @@ void set_struts(alignment align) {
...
-                workarea.height - window.geometry.y, 0, *workarea.height);
+                display_height - window.geometry.y(), 0, display_height);
             sizes[*x11_strut::BOTTOM_START_X] =
-                std::clamp(*window.geometry.x, 0, *workarea.width);
-            sizes[*x11_strut::BOTTOM_END_X] = std::clamp(
-                window.geometry.x + window.geometry.width, 0, *workarea.width);
+                std::clamp(window.geometry.x(), 0, display_width);
+            sizes[*x11_strut::BOTTOM_END_X] =
+                std::clamp(window.geometry.x() + window.geometry.width(), 0,
+                           display_width);
```
Caellian, line 1183: This was buggy. I didn't use `display_width` and `display_height` at all, and instead `workarea.width` which was in fact an absolute end x position and not width.

src/x11.cc
```diff
@@ -352,10 +356,8 @@ void update_x11_workarea() {
   }
 
   XineramaScreenInfo *ps = &si[i];
-  workarea[0] = ps->x_org;
-  workarea[1] = ps->y_org;
-  workarea[2] = workarea[0] + ps->width;
-  workarea[3] = workarea[1] + ps->height;
```
Caellian, line 358: As can be seen here.

brndnmtthws (approved):

Issue #1886, opened 2024-05-02
[Bug]: Conky freezes on mouse movement


Caellian:
> Unfortunately, another problem arose! Scripts stops updating on screen after a few seconds. .xsession-errors logs the following:
```
conky: FOUND: console
conky: FOUND: ncurses
conky: FOUND: file
conky: FOUND: x11

conky: 'openbox' x11 session running 'openbox' destop
```
>Then the scripts stops updating the screen.
>
>Not only does the scripts that needed " require 'cairo_xlib' " stop functioning, but it seems every conky process stops responding. Can't kill running processes either, they simply respawn before they all shut down.
>
> Don't know if this helps, but the moment you touch your mouse, conky stops responding.

_Originally posted by @belrus65 in https://github.com/brndnmtthws/conky/issues/1867#issuecomment-2088499835_

Caellian:
@belrus65 Can you send me your latest lua script as a comment attachment (might need to add `.txt` extension)?

I was debugging mouse events a lot in Openbox past few weeks and they seemed fine, so it might be a weird interaction with cairo.

I also have a few questions:
- You're using pure Openbox?
- Can you attach output of `xwininfo -root -tree` while running conky?

belrus65:
[lua-scripts.zip](https://github.com/brndnmtthws/conky/files/15183542/lua-scripts.zip)
![Screenshot-from-belrus65](https://github.com/brndnmtthws/conky/assets/55268627/e6e9b758-31b3-414f-b580-0678b0b27f92)
[xwininfo-results-conky-1.19.8.txt](https://github.com/brndnmtthws/conky/files/15183543/xwininfo-results-conky-1.19.8.txt)
[xwininfo-results-conky-1.20.2.txt](https://github.com/brndnmtthws/conky/files/15183544/xwininfo-results-conky-1.20.2.txt)

Here are the files. I have a pure openbox gentoo system with 3 monitors running several conky panels on each monitor (I've included a screenshot of my desktop to make it easier to explain). The problem starts as soon as I load up to 4 lua scripts (clock & memory bar on center monitor, and cpu usage dials on the other monitors). If I move the mouse or interact with the desktop (bring up menu with key bindings) the conky processes seem to stall for some time (the more I move the mouse to longer the stall exists). None of these problems existed in the previous (1.19.8)

lineage-of-roots:
@belrus65  

Can I see your conkyrc files as well please? (For invesitgation, and I like your layout :) )

[...]

PR #1920, opened 2024-05-17
Disable Xinput by default

Caellian:
Causes performance issues for some people (running multiple instances at once?).
See #1886.


brndnmtthws (approved):
Might be worth making it configurable at runtime eventually so people can experiment with it.

👍

Issue #1925, opened 2024-05-18
[Bug]: conky pixel unit isn't scaled by screen DPI


LinuxOnTheDesktop:
### What happened?

I built the latest release and installed it on two computers. On one computer, which has DPI in no sense, all was fine. On the other computer, which has a HiDPI screen and DPI scaling set to times-two, the right-hand half of my Conky is missing. In the previous GitHub release I did not have the problem.

Screenshot:

![image](https://github.com/brndnmtthws/conky/assets/47056543/c0705e44-eba1-4de0-aea2-7195a5a0874b)

I have Linux Mint 21.3 Cinnamon.

### Version

1.21.1-pre-256ffdb7

### Which OS/distro are you seeing the problem on?

Linux (other)

### Conky config

```Lua
home = os.getenv("HOME")
dofile ( home .. '/<redacted>/Conky/config/config_x1.lua' )
conky.config = configuration

-- Conky objects: http://conky.sourceforge.net/variables.html
-- Cannot without fuckery remove border from bars (as against from graphs).

-- To monitor cpu usage: top -p "$(pgrep conky)"

conky.text = [[
${offset 10}\
# -----------------------
# CPU usage and frequency
# -----------------------
# CPU
${color4}\
cpu\
# CPU usage
${offset 12}\
${color1}\
${cpubar cpu0 8,25}\
# Frequency
${offset 12}\
${color #6A786A}\
${freq_g}\
${offset 2}GHz\
#
# -----------------------------------------------------------------
# TEMPERATURE
# CPU (the mean of all cores).
# Seems to be no other useful sensor (only wifi & individual cores).
# Use monospace font.
# CPU critical temperature (according to Intel): 100 degrees.
# ------------------------------------------------------------------
#
# CPU TEMP
${offset 12}\
${if_match ${acpitemp}<66}\
# Pastel green
${color1}\
${else}\
${if_match ${acpitemp}<76}\
# Pastel yellow
${color2}\
${else}\
${if_match ${acpitemp}<86}\
# Orange
${color orange}\
${else}\
${if_match ${acpitemp}<99}\
# Pastel red
${color3}\
${else}\
# Red
${color red}\
${endif}\
${endif}\
${endif}\
${endif}\
${acpitemp}\
# Temperature symbol
${offset 1}${font Hack:italic:bold:size=10}${color8}°${font}\
# ------
# DRIVES
# ------
#
${goto 192}\
# ROOT
${font Hack:bold:size=9}${color4}\
[...]

PR #1926, opened 2024-05-18
Remove dpi scaling of minimum_width & maximum_width

Caellian:
This was left over from #1877.

### DIFF
--- a/src/conky.cc
+++ b/src/conky.cc
@@ -858,7 +858,7 @@ void update_text_area() {
   if (fixed_size == 0)
 #endif
   {
-    text_size = conky::vec2i(dpi_scale(minimum_width.get(*state)), 0);
+    text_size = conky::vec2i(minimum_width.get(*state), 0);
     last_font_height = font_height();
     for_each_line(text_buffer, text_size_updater);
     text_size += conky::vec2i::UnitX();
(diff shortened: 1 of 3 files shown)


src/display-wayland.cc
```diff
@@ -661,8 +661,8 @@ bool display_output_wayland::main_loop_wait(double t) {
          text_size.y() + 2 * border_total != height || scale_changed)) {
       /* clamp text_width to configured maximum */
       if (maximum_width.get(*state)) {
-        int mw = global_window->scale * maximum_width.get(*state);
```
Caellian, line 664: Didn't remove `maximum_width` scaling here.

src/conky.cc
```diff
@@ -858,7 +858,7 @@ void update_text_area() {
   if (fixed_size == 0)
 #endif
   {
-    text_size = conky::vec2i(dpi_scale(minimum_width.get(*state)), 0);
+    text_size = conky::vec2i(minimum_width.get(*state), 0);
```
Caellian, line 861: `minimum_width` shouldn't scale either.

Caellian:
Next step would be reverting these changes, introducing proper unit support :laughing:, and making raw numbers default to `pt` instead of `px` as most users expect (#1925).

brndnmtthws (approved):

PR #1936, opened 2024-05-25
Make dpi_scale match original behavior

### LINKED ISSUE #1930: [Bug]: Wrong hr line thickness
### What happened?

After updating to version 1.21.1-1, the display becomes very bold...
Bildschirmfoto_2024-05-22_13-45-48
Bildschirmfoto_2024-05-22_13-48-58

### Version

1.21.1-1

Arch Linux

### Conky config

_No response_

### Stack trace

_No response_

### Relevant log output

_No response_

Caellian:
I arrived at previous solution because the original code used premultiplied values to round the result so I (incorrectly) assumed it was doing something akin to truncation.

Also replaced SFINAE with static_assert - got really carried away with it previously because I believed it provided better type checking integration; but they're the same and errors look the same.

Fixes #1930.

### DIFF
--- a/src/display-output.hh
+++ b/src/display-output.hh
@@ -184,21 +184,13 @@ static inline conky::display_output_base *display_output() {
 }
 
 template <typename T>
-inline std::enable_if_t<std::is_arithmetic<T>::value, T> dpi_scale(T value) {
+inline T dpi_scale(T value) {
+  static_assert(std::is_arithmetic_v<T>,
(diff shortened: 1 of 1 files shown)


brndnmtthws (approved):

Caellian:
Merging without confirmation because (according to tests) the returned values match old behavior better, even if I don't know whether it fixes the issue.

PR #1937, opened 2024-05-25
Move color parse error message out of hex parser

Caellian:
Gives a better error message if no parsers (possibly more in future) can parse the color.

This avoids issues like getting a misleading error message when color name parsing didn't work.

I'm making this change until I'm done with logging improvements (a few tweaks left, but not sure when I'll be able to clean up the code).

### DIFF
--- a/tests/test-common.cc
+++ b/tests/test-common.cc
@@ -196,6 +196,7 @@ TEST_CASE("cpu_percentage and cpu_barval return correct values") {
     REQUIRE(cpu_barval(&obj1) == Approx(0.507));
 
     delete[] info.cpu_usage;
+    info.cpu_usage = nullptr;
   }
 }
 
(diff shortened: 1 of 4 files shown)


Caellian:
The test seems to be failing because it now catches an error in `scan_graph` argument parser. I'll fix that too.

brndnmtthws (approved):

tests/test-common.cc
```diff
@@ -196,6 +196,7 @@ TEST_CASE("cpu_percentage and cpu_barval return correct values") {
     REQUIRE(cpu_barval(&obj1) == Approx(0.507));
 
     delete[] info.cpu_usage;
+    info.cpu_usage = nullptr;
```
Caellian, line 199: This little thing is out of scope, but it causes any tests which attempt using `CRIT_ERR`, that happen after test-common (i.e. all of them) to crash. Tests shouldn't use `CRIT_ERR` because it interrupts their execution, but it's better to handle it explicitly than as a side effect of another test.

Issue #1942, opened 2024-05-29
[Bug]: Version 1.21.1 Images larger


ACR-Jeff:
### What happened?

I am on Void Linux, I updated to version 1.21.1, What I have here is a popup calendar, Images were perfect prior to the update, Images are now larger then their original size or specified given size. If I have an image as 1 pixel in diameter it is very noticeable the size difference, If I have an image as a custom Horizontal bar at 1 pixel height, It is always larger. Code sample: `${image ~/.config/conky/conky-i3/images/H-Line-green.png -p 29,55 -s 225x1}` This should result in the image being only 255x1 pixel in height, Yet it results in a larger image, Which causes the conky to be all out of line, I am only noticing the 1 pixel because it's so obvious.

![Conky-Issue](https://github.com/brndnmtthws/conky/assets/61212551/f7a5405c-8ba5-4019-a640-bd51db052c25)

### Version

1.21.1

### Which OS/distro are you seeing the problem on?

Linux (other)

### Conky config

```Lua
conky.config = {

-- -----------------------------------------
--             Windows Settings           --
-- -----------------------------------------
	own_window = true,
	own_window_type = 'normal',
	own_window_class = 'conky-calendar-i3',
	own_window_transparent = false,
	own_window_hints = 'undecorated,sticky,below,skip_taskbar,skip_pager',
	own_window_argb_visual = true,
	own_window_argb_value = 255,
	own_window_colour = '2e2e2e',

	double_buffer = true,
	--_buffers = true,
	background = true,
	format_human_readable = true,

-- -----------------------------------------
--             Possition & Size           --
-- -----------------------------------------
	alignment = 'bottom_left',

	minimum_width = 282, 
	maximum_width = 282,
	minimum_height = 275,
	--maximum_height = 180,
    	gap_x = 11,
	gap_y = -275,

-- -----------------------------------------
--                  Font                  --
-- -----------------------------------------
	use_xft = true,
	xftalpha = 0,
	override_utf8_locale = true,
	text_buffer_size = 2048,

-- -----------------------------------------
--             Border & Margin            --
-- -----------------------------------------
	draw_borders = false,
	draw_outline = false,
	draw_shades = false,
	draw_graph_borders = false,
	border_inner_margin = 0,
	border_outer_margin = 20,

-- -----------------------------------------
--                 Colors                 --
-- -----------------------------------------
	default_color = 'DCDCDC',
	color0 = '19ae38',
	color1 = 'DDDDDD',
[...]

PR #1949, opened 2024-05-30
Revert removal of dpi scaling

Caellian:
Reverts: #1926, #1877
Cause of: #1925 & maybe #1942
Reopens: #1528


Caellian:
I'll add a separate feature that adds unit handling, so real pixels can be used instead of _logical_ ones.

brndnmtthws (approved):

PR #1957, opened 2024-06-04
Fix for the graph scale text colour changing depending on graph position

randomcoder67:
# Checklist

## Description

In the current latest code of conky, when you have a graph scale enabled, the text will change colour whenever a "high point" on the graph reached the "end" of the graph. It's difficult to describe so I took some screenshots:

### Before
before-normal
before-error

The scale text changes colour to whatever the colour of the leftmost "bar" of the graph is.

### After
after-normal
after-error

No longer does this

I'm not sure if this was an intended feature, if it was, feel free to close this. imo it shouldn't change colour because it makes the text very hard to read when the whole graph is full of red. 
It's a simple fix, just adding one extra line to change the colour back to the "first" graph colour (i.e. the colour of the lowest values)

There shouldn't be any unintended side effects of the change, because looking at the code, this is the last if statement in a block, and right after the if statements, the colour gets changed anyway (`set_foreground_color(last_colour);`)

### DIFF
--- a/src/conky.cc
+++ b/src/conky.cc
@@ -1374,6 +1374,7 @@ int draw_each_line_inner(char *s, int special_index, int last_special_applied) {
             }
 #ifdef BUILD_MATH
             if (show_graph_scale.get(*state) && (current->show_scale == 1)) {
+              set_foreground_color(current->first_colour);
               int tmp_x = cur_x;
               int tmp_y = cur_y;
               cur_x += font_ascent() / 2;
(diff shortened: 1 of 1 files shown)


randomcoder67:
Not sure what the Docker issue is, I don't have any experience with Docker.

brndnmtthws (approved):
Looks okay to me, CCing @Caellian.

The docker build should be fixed in #1958 (hopefully for real this time).

Caellian (approved):
You have approval from me as the change should work and the `first_colour` is a better choice than the one that fills the rect. And using the first column for coloring is wrong anyway (should be current state?).

There are edge cases with readability still - in case someone uses a gradient that's got similar color stops (e.g. lightness gradient with 40% difference). But the colors might also be similar to `last_colour` (borders), so there's really no perfect solution that will work in all cases and make everyone happy with how it looks.

Thank you for the contribution, it's a good improvement. Feel free to merge it if you don't plan on making any further modifications.

PR #1963, opened 2024-06-09
Restore previous strut behavior

m42a:
# Checklist

## Description

This PR reverts conky to its pre-1.20.2 strut behavior. Before that version, conky would prefer to use vertical struts over horizontal struts (e.g. when placed in the top-left corner, conky would use the top strut instead of the left strut). PR #1849 changed this to prefer horizontal struts.

In the future it may be useful to allow the user to specify the strut direction, but for now I believe the best course of action is to restore the previous behavior ASAP.

### DIFF
--- a/src/geometry.h
+++ b/src/geometry.h
@@ -430,21 +430,13 @@ struct rect {
   /// @brief Rectangle width.
   /// @return width of this rectangle.
   inline T width() const {
-    if constexpr (Kind == rect_kind::SIZED) {
-      return this->m_other.x();
-    } else {
-      return this->m_other.x() + this->m_pos.x();
-    }
+    return size().x();
   }
 
   /// @brief Rectangle height.
   /// @return height of this rectangle.
   inline T height() const {
-    if constexpr (Kind == rect_kind::SIZED) {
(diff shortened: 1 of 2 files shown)


Caellian (approved):

Caellian:
It makes more sense IMO to consume horizontal space first as it's almost always greater than vertical. I also didn't intend on changing default behavior.

src/geometry.h
```diff
@@ -430,21 +430,13 @@ struct rect {
...
   }
 
   /// @brief Rectangle height.
   /// @return height of this rectangle.
   inline T height() const {
-    if constexpr (Kind == rect_kind::SIZED) {
-      return this->m_other.y();
-    } else {
-      return this->m_other.y() + this->m_pos.y();
-    }
```
Caellian, line 447: This change I don't agree with as much though, because it turns a scalar operation into a vector operation which can't be SIMDed for some types. Compiler _might_ optimize it away but that should be checked with godbolt.

Caellian (changes requested):

Caellian (approved):

PR #1972, opened 2024-06-13
Fix AC/battery detection logic on FreeBSD.

fernape:
The logic for the detection of the battery/AC line was not completely correct for FreeBSD. Launching conky on console shows the following:

Cannot read sysctl "hw.acpi.battery.time"
Cannot read sysctl "hw.acpi.battery.life"
Cannot read sysctl "hw.acpi.battery.state"
Cannot read sysctl "hw.acpi.acline"
Unknown battery state 8!

In a PC, the hw.acpi.battery MIB does not exist.
Also, the hw.acpi.acline is only present if supported by the hardware. In addition, some variables were used uninitialized and that causes strange behavior: in a PC it showed it worked on battery and the % of charge was an ridiculous big number.

bad

This patch addresses the issue. It fixes the problem in the PC. It has also being tested in a laptop running FreeBSD current plugin and unplugging the AC line and also snatching the battery mercilessly to see if something breaks.

good

# Checklist
- [x ] Documentation in `doc/` has been updated

## Description

* Describe how the changes will affect existing behaviour.
* Include any relevant screenshots/evidence demonstrating that the changes work and have been tested.

### DIFF
--- a/src/freebsd.cc
+++ b/src/freebsd.cc
@@ -455,17 +455,17 @@ static void get_battery_stats(int *battime, int *batcapacity, int *batstate,
 
   if (battery_present) {
     if (battime && GETSYSCTL("hw.acpi.battery.time", *battime)) {
-      fprintf(stderr, "Cannot read sysctl \"hw.acpi.battery.time\"\n");
+      NORM_ERR("Cannot read sysctl \"hw.acpi.battery.time\"");
     }
     if (batcapacity && GETSYSCTL("hw.acpi.battery.life", *batcapacity)) {
-      fprintf(stderr, "Cannot read sysctl \"hw.acpi.battery.life\"\n");
+      NORM_ERR("Cannot read sysctl \"hw.acpi.battery.life\"");
     }
     if (batstate && GETSYSCTL("hw.acpi.battery.state", *batstate)) {
-      fprintf(stderr, "Cannot read sysctl \"hw.acpi.battery.state\"\n");
+      NORM_ERR("Cannot read sysctl \"hw.acpi.battery.state\"");
     }
   }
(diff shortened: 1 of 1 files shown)


src/freebsd.cc
```diff
@@ -427,24 +427,50 @@ double get_acpi_temperature(int fd) {
...
+	  *ac = 1;
+	  *batstate = 7;
+	  return;
   }
-  if (batstate && GETSYSCTL("hw.acpi.battery.state", *batstate)) {
-    fprintf(stderr, "Cannot read sysctl \"hw.acpi.battery.state\"\n");
+
+  if (battery_present) {
+    if (battime && GETSYSCTL("hw.acpi.battery.time", *battime)) {
+      fprintf(stderr, "Cannot read sysctl \"hw.acpi.battery.time\"\n");
```
brndnmtthws, line 458:
```suggestion
      NORM_ERR("Cannot read sysctl \"hw.acpi.battery.time\"");
```

brndnmtthws (changes requested):
Overall it's good, but since you're making changes here we should replace those `fprintf()` calls with [`NORM_ERR`](https://github.com/brndnmtthws/conky/blob/33903e7d6c4681b4f842dc39755e3a4a72f34185/src/logging.h#L76-L81) instead, if you wouldn't mind.

fernape:
> Overall it's good, but since you're making changes here we should replace those `fprintf()` calls with [`NORM_ERR`](https://github.com/brndnmtthws/conky/blob/33903e7d6c4681b4f842dc39755e3a4a72f34185/src/logging.h#L76-L81) instead, if you wouldn't mind.

Thank you for the suggestion. I didn't know about `NORM_ERROR` and the `fprintf()` were already there. I'll take that into account.

brndnmtthws (approved):

PR #1980, opened 2024-07-02
Fix minimum_height not scaled with dpi_scale like minimum_width

agorgl:
Related to #1925, comment https://github.com/brndnmtthws/conky/issues/1925#issuecomment-2202903507


Caellian (approved):
Thank you.

PR #1991, opened 2024-07-20
Made upspeedgraph and downspeed graph scale on a common axis 

### LINKED ISSUE #1820: [Enhancement]: make downspeedgraph and upspeedgraph use same scaling
It might be better to clamp `downspeedgraph` and `upspeedgraph` according to maximum value across both graphs because people assume (#979, #1034) they're using the same vertical step.

donutAnees:
# Checklist

#### Before
In the current latest code of conky, downspeedgraph and upspeedgraph look the same because they're both individually scaled with respect to max value.

Screenshot 2024-07-20 at 10 59 35 PM

#### After
Screenshot 2024-07-21 at 3 15 48 AM
Screenshot 2024-07-21 at 3 15 59 AM

This has now been fixed and both of them are scaled on a common axis. I have tested the code, and it does not influence any other graphs.

The changes made include 
1. Adding ```char speedgraph``` flag to the ```struct graph``` and ```struct special_node```.
2. Declaring a global ```double maxval``` to keep track of the current max value across the two graphs.
3.  Adding parameter ```char speedgraph``` to scan_graph method.
4.  And finally checking if the current value getting added is bigger than maxval, if so we make our scale = maxval, in the graph_append method.

Fixes #1820

### DIFF
--- a/src/common.cc
+++ b/src/common.cc
@@ -352,7 +352,7 @@ void print_no_update(struct text_object *obj, char *p,
 
 #ifdef BUILD_GUI
 void scan_loadgraph_arg(struct text_object *obj, const char *arg) {
-  scan_graph(obj, arg, 0);
+  scan_graph(obj, arg, 0, FALSE);
 }
 
 double loadgraphval(struct text_object *obj) {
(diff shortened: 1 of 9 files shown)


brndnmtthws (approved):
This is great, thanks!

A bit weird that we mix `bool`, `char`, `true/false`, `TRUE/FALSE` but that's something for another PR.

PR #1993, opened 2024-07-22
Fixed graph parsing error

### LINKED ISSUE #1992: [Bug]: Graph configuration parsing error
### What happened?

Currently, when we try to use the following configuration in the conky.conf file: `downspeedgraph en0 50,280 10 -t`, it does not match the expected format `[height],[width] [scale]`. Instead, it matches the format `[height],[width] [color1] [color2]` due to the presence of the flag.

### Version

v1.21.4

Ubuntu

### Conky config

_No response_

### Stack trace

_No response_

### Relevant log output

_No response_

donutAnees:
# Checklist

### Before
In the current version of conky, ```scan_graph``` method, parsing doesn't consider the flags and can lead to undesirable changes. The parsing also had issues with resetting some variables after ```sscanf```.
### After
The changes made will resolve these issues. The parsing now correctly considers flags, preventing undesirable changes, and ensures that variables are properly reset.

Fixes #1992

### DIFF
--- a/src/specials.cc
+++ b/src/specials.cc
@@ -249,6 +249,7 @@ std::pair<char *, size_t> scan_command(const char *s) {
  * @param[out] obj  struct in which to save width, height and other options
  * @param[in]  args argument string to parse
  * @param[in]  defscale default scale if no scale argument given
+ * @param[in]  speedGraph if graph is network speed graph or not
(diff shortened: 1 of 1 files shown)


brndnmtthws (approved):
Thanks!

PR #2000, opened 2024-07-26
Implemented graph value clamping

### LINKED ISSUE #1060: [Enhancement]: clamp (ceil) graph values to first unit
please consider implementing an option to force rounding up of minimal values
which otherwise fall below the threshold of visibility in a given graph 
i.e. values between zero and the first visible unit

for example
network graphs can be troublesome as their values span such a large range
causing small values to go by without being displayed
(logarithmic scaling is not satisfactory for some cases)

this feature is especially useful for small graphs
such as those in a menu bar or status panel/strip
to be sure that even small values will be displayed

in consideration of high resolution displays
perhaps the option should specify a number of pixels
(as a single pixel may still be nearly invisible)

e.g. a option value of 4 would cause any graph value between zero
and the normal threshold to paint the first 4 pixels
(default option value: 1 pixel)

donutAnees:
# Checklist

### Before
Values that fall below the threshold of visibility in a given graph are not seen, leading to small values being omitted.

### After 
A new flag, '-m value', has been implemented, allowing the user to specify the threshold value between 0 and 5. This will enable values below the threshold to be displayed.

```downspeedgraph en0 50,280 ADFF2F 32CD32 -t -x -m 5```
Screenshot 2024-07-27 at 1 03 35 AM

Fixes #1060

### DIFF
--- a/doc/variables.yaml
+++ b/doc/variables.yaml
@@ -99,6 +99,7 @@ values:
       - (-l)
       - (-x)
       - (-y)
+      - (-m value)
   - name: apcupsd_model
     desc: Prints the model of the UPS.
   - name: apcupsd_name
@@ -331,7 +332,8 @@ values:
       numbers) when you use the -l switch. Takes the switch '-t' to use a
       temperature gradient, which makes the gradient values change depending
       on the amplitude of a particular graph value (try it and see). The flag
-      '-x' inverts the x axis and '-y' inverts the y axis of the graph.
+      '-x' inverts the x axis and '-y' inverts the y axis of the graph. The flag
+      '-m' clamps values below the specified value to the value (excluding 0) */
     args:
       - (cpuN)
       - (height),(width)
@@ -342,6 +344,7 @@ values:
       - (-l)
       - (-x)
       - (-y)
+      - (-m value)
   - name: curl
     desc: |-
       Download data from URI using Curl at the specified interval.
@@ -395,7 +398,8 @@ values:
       '-t' to use a temperature gradient, which makes the gradient values
       change depending on the amplitude of a particular graph value (try it
       and see). The flag '-x' inverts the x axis and '-y' inverts the y axis 
-      of the graph.
+      of the graph. The flag '-m' clamps values below the specified value 
+      to the value (excluding 0).
     args:
       - (device)
       - (height),(width)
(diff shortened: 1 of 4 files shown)


brndnmtthws:
Thanks for the PR! My only question is, are we sure "clamping" is the right term for this? When I think of clamping, I tend to think about having an upper bound limit or lower bound (i.e., setting a minimum and maximum possible range of values). This is a slightly different: we're essentially setting a lower bound limit, but only for nonzero values. It's different from having a floor or ceiling, as again–those apply whether you have a zero or not. I wonder if there is a better term to describe it?

Clamp is basically `min(max(x, min_val), max_val)`. But what we're doing here is more like `if x != 0 then max(min_val, x) else x end`.

Maybe even just calling this "nonzero minimum" or something is more descriptive.

Thoughts?

donutAnees:
Hey! Those are very valid points, traditional definition of clamping would not be a very applicable here. The reason I thought clamping made sense is that we are setting a lower bound with a condition, so it's more like 'conditional clamping' rather than the default definition. However, I believe the "nonzero minimum" would be a good term to describe this or something like "nonzero lowerbound" would be better. I’m really interested to hear your thoughts on this!

brndnmtthws:
> Hey! Those are very valid points, traditional definition of clamping would not be very applicable here. The reason I thought clamping made sense is that we are setting a lower bound with a condition, so it's more like 'conditional clamping' rather than the default definition. However, I believe the "nonzero minimum" would be a good term to describe this or something like "nonzero lowerbound" would be better. I’m really interested to hear your thoughts on this!

I think these are all good suggestions: conditional clamping, or nonzero minimum/lowerbound make sense. I think if you update the docs to use that language, then this PR should be good to go!

donutAnees:
> > Hey! Those are very valid points, traditional definition of clamping would not be very applicable here. The reason I thought clamping made sense is that we are setting a lower bound with a condition, so it's more like 'conditional clamping' rather than the default definition. However, I believe the "nonzero minimum" would be a good term to describe this or something like "nonzero lowerbound" would be better. I’m really interested to hear your thoughts on this!
> 
> I think these are all good suggestions: conditional clamping, or nonzero minimum/lowerbound make sense. I think if you update the docs to use that language (I defer to you on which to use, as the PR author), then this PR should be good to go!

"Thanks a lot for the freedom! I've made the changes to the doc. Cheers!"

brndnmtthws (approved):

PR #2049, opened 2024-10-02
Implement `-U`.

g0mb4:
If this argument is provided Conky won't start if another Conky process is already running.

Conky starts multiple times on my system at startup, and it is bothering me. I'm too lazy to check all of my config files, so I added this mechanism to prevent this behavior.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -36,8 +36,6 @@
 #include "display-output.hh"
 #include "lua-config.hh"
 
-#include <dirent.h>
-
 #ifdef BUILD_X11
 #include "x11.h"
 #endif /* BUILD_X11 */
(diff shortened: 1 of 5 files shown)


brndnmtthws:
A couple issues: `/proc` is generally Linux-only (see https://en.wikipedia.org/wiki/Procfs), so we can't assume it's there on BSDs and macOS.

Another problem, getting the process name from `/proc/[pid]/cmdline` isn't a great idea because the command could have a symlink or some other strangeness. Better to use either `/proc/[pid]/stat` or follow the `/proc[pid]/exe` symlink.

g0mb4:
Thanks! It's the typical case of "it works on my machine"...
I'll try to figure out a POSIX solution (maybe using `ps`).

brndnmtthws:
I know I'm a bit late on getting back to you on this, and I apologize for that.

I am not a big fan of shelling out to `ps` in this way, it also feels like something that is likely to break and probably not consistent across platforms. I'm more inclined to take the original patch you submitted, and just refactor it so that it's Linux-only, and if people want to port it to the BSDs/macOS then so be it.

g0mb4:
No problem at all.

I tested the `ps` on FreeBSD and on Linux and the output was the same. I read that the Mac implementation should work as expected (based on some man pages).

It feels a "good enough" soultion for me as well, but this was the most POSIX-way I can came up with.

g0mb4:
I can do the `/proc/[pid]/stat` method in a couple of days if it is okay.

Caellian:
I just added `ifdef` for Linux around everything and note in the manual, to avoid issue reports for it not working on non-Linux systems.

Caellian (approved):

src/main.cc
```diff
@@ -268,10 +270,51 @@ static void print_help(const char *prog_name) {
...
          "   -p, --pause=SECS          pause for SECS seconds at startup "
-         "before doing anything\n",
+         "before doing anything\n"
+         "   -U, --unique              only one conky process can be created\n",
          prog_name);
 }
 
+#if defined(__linux__)
+// NOTE: Only works on systems where there is a /proc/[pid]/stat file.
+static bool is_conky_already_running() {
```
brndnmtthws, line 280: Minor nit pick: can you relocate this function to `src/linux.cc`?

brndnmtthws (changes requested):
Overall looks good, I have one minor request (to move the Linux-only code to `src/linux.cc`.

src/main.cc
g0mb4: Sure.

brndnmtthws (approved):

LinuxOnTheDesktop:
This option seems like a useful addition. For, one can find a few reports on the web of people ending up with weird-looking conkies that turn out to be one conky superimposed upon another. I've encountered that myself. So there's even a case for making this option the default.

g0mb4:
Thanks! I can try to implement this feature on some BSDs if @brndnmtthws finds this as useful as you.

Caellian:
> So there's even a case for making this option the default.

After triaging issues I'd say changing the defaults will likely cause more confusion than having the option be opt-in. A lot of setups also spawn several conky instances to show information in different corners of the screen for example.

> I can try to implement this feature on some BSDs if [Brandon] finds it as useful as you.

It would be a nice feature-parity addition, but it's up to you. I opened #2072, so you can use it as a guide on which platforms need to be supported if you decide to add support any other. I will also redirect people there if they open new issues about the feature not being supported for one of those platforms.

LinuxOnTheDesktop:
@g0mb4 

Question: will `conky -U` run Conky if _another user_ is running a Conky? For, consider so-called fast-user switching.

g0mb4:
@LinuxOnTheDesktop 

It depends on how `/proc` was mounted. If it allows the current user to see the processes of the other user, then conky won't start.

PR #2057, opened 2024-10-10
Add automatic labeling for issues

Caellian:
This PR:
- replaces old labeler action with a [different one](https://github.com/srvaroa/labeler) that works on both PRs (as the previous one) _and_ issues.
- adds support for labeling of os related things.
- adds regex checking for issue content to automatically label issues based on keywords in (mostly) titles and body.
  - because users include configs in issues, a lot of keywords can't be checked for in the issue body as too many labels would be detected ([maybe in future](https://github.com/srvaroa/labeler/issues/156))

The only thing I don't like is that it re-applies labels to all issues when a new one is created. I made it pick labels very conservatively, but it could be annoying in some cases.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -164,12 +164,53 @@ lua:
           - src/lua-config.cc
           - src/lua-config.hh
 
-macos:
+'os: linux':
   - changed-files:
       - any-glob-to-any-file:
-          - src/darwin_sip.h
-          - src/darwin.h
-          - src/darwin.mm
+          - src/linux.cc
+          - src/linux.h
+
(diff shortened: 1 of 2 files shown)


Caellian:
I tested to ensure the [old labeling behavior is the same](https://github.com/Caellian/conky/pull/11) and that [new one works](https://github.com/Caellian/conky/issues).

Caellian:
I wanted to move `stale` labeling into this as well, but the changed labeler doesn't support auto-closing stale issues.

brndnmtthws (approved):

PR #2064, opened 2024-10-27
Fix if_match comparison for string operands containing operator symbols (#1988)

### LINKED ISSUE #1988: [Bug]: if_match parser doesn't handle comparison of strings containing `!`
### What happened?

The Wi-Fi ESSID is not escaped correctly, and it is confusing `if_match`. The execution continues, but the first code block is always matched regardless of the values (see log and config).

I tested both with the Debian version (1.18) and the latest version using AppImage (1.21.4), the problem is the same.

### Version

1.21.4

Debian

### Conky config

```Lua
conky.text = [[${color white}
WiFi:
${if_up wlp59s0}${if_match "${wireless_essid wlp59s0}" == "off/any"}${else}Wi-Fi    ↑ ${upspeed wlp59s0} - ↓ ${downspeed wlp59s0}
        ${wireless_essid wlp59s0} (${wireless_channel wlp59s0}) ${wireless_link_qual_perc wlp59s0}%
        ${addrs wlp59s0}
$endif$endif
]]
```

### Stack trace

_No response_

### Relevant log output

```Shell
conky: failed to parse compare string '"FRITZ!Box 7520 HI" == "off/any"'
conky: compare failed for expression '"FRITZ!Box 7520 HI" == "off/any"'
```

liestrela:
# Checklist

## Description
Maybe too naive, but skipping the first comparison operand if it is a string avoids the issue. Fixes #1988.

### DIFF
--- a/src/algebra.cc
+++ b/src/algebra.cc
@@ -176,6 +176,7 @@ long arg_to_long(const char *arg) {
   }
   return l;
 }
+
 int compare(const char *expr) {
   char *expr_dup;
   int idx, mtype;
--- a/src/algebra.cc
+++ b/src/algebra.cc
@@ -40,9 +40,15 @@
  * returns the index of the first op character or -1 on error
  */
 int find_match_op(const char *expr) {
-  unsigned int idx;
+  unsigned int idx = 0;
 
-  for (idx = 0; idx < strlen(expr); idx++) {
+  /* if first operand is a string, skip it */
+  if (expr[idx] == '"') {
+    for (idx=1; expr[idx] && expr[idx] != '"'; idx++);
+    idx++;
+  }
+
+  for (; idx < strlen(expr); idx++) {
     switch (expr[idx]) {
       case '=':
       case '!':
--- a/src/algebra.h
+++ b/src/algebra.h
@@ -47,5 +47,7 @@ enum arg_type {
 
 int compare(const char *);
 int check_if_match(struct text_object *);
+int get_match_type(const char *expr);
+int find_match_op(const char *expr);
 
 #endif /* _ALGEBRA_H */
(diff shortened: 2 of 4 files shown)


brndnmtthws:
This doesn't seem like the worst solution, at least it's a fairly non-invasive change.

Would you like to take a stab at writing a couple of simple unit tests for this? Ideally a test that breaks with the old code, and passes with the new code. I'm also a little unsure about how good the quote (`"`) parsing code works. You can probably use chatgpt or copilot to generate the tests trivially if you don't feel like doing it manually.

liestrela:
All of the following would break with the old code and now behave as expected:
```
1: ${if_match "${exec echo a!b}" == "ab"}A${else}B$endif want B
2: ${if_match "${exec echo a!/b}"=="a!/b"}A${else}B$endif want A
3: ${if_match "${exec echo a!=/a==b}"== "a!=/a==b"}A${else}B$endif want A
4: ${if_match "${exec echo a!=/b==b}"== "a!==/a==b"}A${else}B$endif want B
5: ${if_match  "${exec echo a!======b}"== "!==/=="}A${else}B$endif want B
5: ${IF_MATCH  " ${exec echo '!=<>=='}" == " !=<>=="}A${else}B$endif want A
6: ${IF_MATCH  "${exec echo 'a!=><=='}" < "b!=><=="}A${else}B$endif want A
7: ${IF_MATCH  "${exec echo '!=<>=='}" == "!=<>=="}A${else}B$endif want A
8: ${if_match "${exec echo '='}" == "="}A${else}B$endif want A
9: ${if_match "${exec echo FRITZ!Box 7520 HI}" == "off/any"}A${else}B$endif want B
```
9 is actually the string comparison the creator of the issue was having trouble with.
On the quote parsing, it is done in pretty much the same way in `get_arg_type`.

brndnmtthws (approved):
I added some tests, I think we're good now.

Issue #2072, opened 2024-11-07
[Feature]: Implement `-U` (single instance) option for BSD & MacOS


Caellian:
With #2049 merged, conky now supports `-U` option on Linux which makes it exit early if it detects another instance running. This feature requires a separate implementation for other platforms as they keep track of running processes differently.

I'm opening this tracking issue to track progress on support for other platforms Conky can currently run on.

Following platforms are currently supported:
- [x] Linux
- BSD
  - [ ] DragonFly
  - [ ] FreeBSD
  - [ ] NetBSD
  - [ ] OpenBSD
- [ ] MacOS
- [ ] Solaris

If you're interested, open a PR and mention this issue (just don't use "closes" phrasing) so it shows up here. Once support for all platforms is implemented this issue will be closed.

PR #2079, opened 2024-11-14
Implement `-U` for Haiku.

g0mb4:
Compiled with `cmake -DBUILD_X11=FALSE ..`

Tested on Haiku-r1beta5.
Takes part of #2072.


Caellian:
Changes to `common.cc` are already part of #2080. Let's get that one finished and merged first. Alternatively, given that Haiku doesn't compile at the moment (yes?), you can remove changes to `common.cc` to avoid difficulties with rebasing that one (given that it's larger).

g0mb4:
Sure, I'll revert `common.cc`.

g0mb4:
Works with the latest OpenBSD and NetBSD changes.

Issue #2052, opened 2024-10-03
Cleanup `text_object` construction in core.cc


Caellian:
Currently, in `core.cc`, [`construct_text_object`](https://github.com/brndnmtthws/conky/blob/b05810d439447666bb7a8c3823d94833cc282033/src/core.cc#L375) function uses a ~1600 lines long if-elseif chain.

All of those are wrapped in macros and really difficult to reason about.

I propose all those macros get replaced with `inline` functions that handle parameters and construction in a cleaner C++ way and gperf can be used to generate a hash based lookup for the appropriate function pointer.

A single function dereference ought to be faster than 450+ consecutive `strcmp`s. But performance isn't really as much the point as this code being very unreadable:

<details>

```c++
#define __OBJ_HEAD(a, n) \
  if (!strcmp(s, #a)) {  \
    obj->cb_handle = create_cb_handle(n);
#define __OBJ_IF obj_be_ifblock_if(ifblock_opaque, obj)
#define __OBJ_ARG(...)                              \
  if (!arg) {                                       \
    free(s);                                        \
    CRIT_ERR_FREE(obj, free_at_crash, __VA_ARGS__); \
  }

/* defines to be used below */
#define OBJ(a, n) __OBJ_HEAD(a, n) {
#define OBJ_ARG(a, n, ...) __OBJ_HEAD(a, n) __OBJ_ARG(__VA_ARGS__) {
#define OBJ_IF(a, n)         \
  __OBJ_HEAD(a, n) __OBJ_IF; \
  {
#define OBJ_IF_ARG(a, n, ...)                       \
  __OBJ_HEAD(a, n) __OBJ_ARG(__VA_ARGS__) __OBJ_IF; \
  {
#define END \
  }         \
  }         \
  else

#ifdef BUILD_GUI
  if (s[0] == '#') {
    obj->data.l = parse_color(s).to_argb32();
    obj->callbacks.print = &new_fg;
  } else
#endif /* BUILD_GUI */
#ifndef __OpenBSD__
    OBJ(acpitemp, nullptr)
  obj->data.i = open_acpi_temperature(arg);
  obj->callbacks.print = &print_acpitemp;
  obj->callbacks.free = &free_acpitemp;
  END OBJ(acpiacadapter, nullptr) if (arg != nullptr) {
#ifdef __linux__
    if (strpbrk(arg, "/.") != nullptr) {
      /*
       * a bit of paranoia. screen out funky paths
[...]

PR #2080, opened 2024-11-15
Fix OpenBSD build

### LINKED ISSUE #2073: [Bug]: unable to build due to Vc on OpenBSD 7.6
### What happened?

I cloned the latest from github  and after extracting it and adding a build directory and cd-ing to that directory, I type cmake ../ and there is scrolling on the erminal and I get the following

-- Build files have been written to: /home/sheldoni/Downloads/conky/build

and it stops.  So I see that there is a makefile now in build and I type make and I get the following error

In file included from /home/sheldoni/Downloads/conky/build/3rdparty/Vc/trigonometric_SSE2.cpp:31:
In file included from /home/sheldoni/Downloads/conky/3rdparty/Vc/Vc/vector.h:221:
/home/sheldoni/Downloads/conky/3rdparty/Vc/Vc/scalar/math.h:118:7: error: no member named 'sincos' in the global namespace
    ::sincos(x.data(), &sin->data(), &cos->data());
    ~~^
1 error generated.
*** Error 1 in . (3rdparty/Vc/CMakeFiles/Vc.dir/build.make:186 '3rdparty/Vc/CMakeFiles/Vc.dir/trigonometric_SSE2.cpp.o': cd /home/sheldoni/D...)
*** Error 2 in . (CMakeFiles/Makefile2:367 '3rdparty/Vc/CMakeFiles/Vc.dir/all': make -s -f 3rdparty/Vc/CMakeFiles/Vc.dir/build.make 3rdparty...)
*** Error 2 in /home/sheldoni/Downloads/conky/build (Makefile:156 'all': make -s -f CMakeFiles/Makefile2 all)

Am I doing something wrong?

### Version

latest cloned version

Linux (other)

### Conky config

_No response_

### Stack trace

_No response_

### Relevant log output

_No response_

g0mb4:
Using OpenBSD 7.6.

Try to close #2073, unfortunately #2075 is not enough.


Caellian:
Not all these changes are needed for BSD compatibility.

I'm assuming you're basing these off of patches in BSD package because they look similar. Those patches do a lot of seemingly unnecessary modifications.

For instance `void some_func(void)` is technically not needed anymore, esp. given that we use a C++ compiler, and is a C specific way of making a function explicitly 0-arg (cause errors when called with arguments). g++ and clang++ complain about this already I think, but more importantly, it doesn't help the BSD build and is a correctness change for C code.

Changing return types of functions from void to int is also out of scope of "Fixing BSD build" and is another change that should ideally be removed from this PR.

Some of these are good changes, but I strongly suggest you remove them from this PR to reduce its surface area. Ideally, fixing BSD build shouldn't introduce any quirks to non-BSD code, so removing them will help when bugs have to be tracked down.

Also, while we're at this, adding an optional BSD workflow would make this more sustainable. I've started with this so I'll add it here if you don't mind.

Caellian:
Referencing #2052 as cause of issues.

g0mb4:
I'm just running `make` and fixing the warnings/errors :)

g0mb4:
And yes, `void f(void)` is done by my autopiloted brain.

Caellian:
Not me telling you to reduce changes and then refactoring the whole file :sweat_smile: 

Uhm... got carried away, `bsdapm.cc` needed some love.

Caellian:
That should be it, I can build now with default options.

Default package dependencies:
```
- git (to clone the repo)
- cmake
- lua-5.3.6
- gperf

X11:
- imlib2

Wayland (not yet supported by OpenBSD)
```

Enabling `BUILD_XINPUT`, `BUILD_LUA_CAIRO`, `BUILD_LUA_CAIRO_XLIB` and `BUILD_LUA_IMLIB2` works without any additional package requirements.

@g0mb4 please test things and (optionally) fix any issues you find, I have a minimal build environment in a VM so it would be better if you try it out as I'm guessing you've got a graphical environment going. I achieved what I was aiming for in #2075.

Caellian:
I believe that pretty much covers my intent behind #2075. I'm building this from a minimal VM environment and would appreciate if you could test it @g0mb4. Feel free to implement any of the added stubs before requesting review.

For future reference (CI?), package names:

Minimal package dependencies:
- `git` (to clone)
- `cmake`
- `gperf`
- `lua-5.3`
- X11 (comes preinstalled)
  - `imlib2`

Optional pkg deps:
- `cairo` (if `BUILD_LUA_CAIRO`, `BUILD_LUA_CAIRO_XLIB`)
- `imlib2` (if `BUILD_LUA_IMLIB2`)
- `librsvg` (if `BUILD_LUA_RSVG`)

g0mb4:
Wow! Thank you for the help! I'll test is as soon as possible.

g0mb4:
It works on my setup.

g0mb4:
I think adding more features really should be part of a new PR.

Caellian:
Alright, then it's done.

I didn't test all possible features because built time was atrocious, but I think they're likely to work. Core, X11 stuff and LUA stuff is confirmed to build on different architectures.

Following `BUILD_*` features might warrant additional attention in separate PRs, but in general they should work if appropriate dependencies are installed:
|     `BUILD_*`     |          Works           |               Dependency               |
| :---------------: | :----------------------: | :------------------------------------: |
|    `AUDACIOUS`    |          should          |              `audacious`               |
|      `CURL`       |          should          |      `curl` (in default install)       |
|      `HTTP`       |          should          |            `libmicrohttpd`             |
|      `ICAL`       |          should          |               `libical`                |
| `INTEL_BACKLIGHT` |         unknown          | `/sys/class/backlight/intel_backlight` |
|       `IRC`       |          should          |             `libircclient`             |
|     `JOURNAL`     |         unlikely         |               `systemd`                |
|      `MYSQL`      |          should          |          `mysql.h` (mariadb?)          |
|     `NVIDIA`      |   nvidia support flaky   |            `NVCtrl` headers            |
|   `PULSEAUDIO`    |          should          |              `pulseaudio`              |
|       `RSS`       |          should          |    `curl` (default), `libxml-2.13`     |
|      `WLAN`       | not used; always enabled |        ioctl and native headers        |
|      `XMMS2`      |          should          |                `xmms2`                 |

Adding CI would be nice even if OpenBSD isn't a primary support target because it would improve code quality by failing on non standard language/stdlib use.

src/common.cc
```diff
@@ -130,20 +132,27 @@ double get_time() {
...
-    wordfree(&p);
-    return std::string(resolved_path);
+    std::string expanded = variable_substitute(input);
+    std::filesystem::path absolute = std::filesystem::absolute(expanded);
+    return std::filesystem::weakly_canonical(absolute);
+  } catch (const std::filesystem::filesystem_error &e) {
+    // file not found or permission issues
+    NORM_ERR("can't canonicalize path: %s", source.c_str());
+    return source;
+  }
```
Caellian, line 155: @brndnmtthws I requested review mostly for this change that affects non-BSD. I'd merge myself if it weren't for that (given that rest of changes won't "build less" haha). Because I hijacked g0mb4's PR, GitHub allows me to merge without a review, so I'll wait for you to merge it as intended.

Caellian: I see possible need for: `~USERNAME`, command substitution, maaaybe arithmetic expansion, limited wildcard expansion (producing a single-match path, fail on many)

Other are a side effect of how POSIX basically passes the content into shell. Arithmetic expansion is complicated because we need some way of recursively parsing it.

OpenBSD people say it should be replaced with `glob` if possible (makes sense for single-match thing described above). MacOS is deprecating it (it even used perl for implementation at some point). Windows has it's own thing (`CommandLineToArgvW`).

Whole glob thing also shows a bug with previous implementation - `to_real_path` could've returned a single string which contained multiple paths.

Should I make the current code a fallback for OpenBSD and use the old code or continue to implement a limited version of `wordexp`?

Caellian: Reverted old code for non-OpenBSD. Deferring this for a separate PR.

Caellian (approved):
From perspective of already supported platforms these are minor consistency tweaks.

From perspective of OpenBSD, it allows conky to at least build which is most important because official OpenBSD port is stuck on a very old version due to build issues.

I'm merging this so @g0mb4 can continue working on NetBSD support in #2089 without having to deal with rebasing.

@brndnmtthws feel free to revert this PR if you don't like something about this, but I'll merge it for now given that CI passes (builds) so none of the already supported platforms should be affected.

@g0mb4 don't delete the branch in case @brndnmtthws rolls it back and requests edits.

PR #1876, opened 2024-04-30
Use private mapping from provided shm_pool buffer

Caellian:
Using `MAP_SHARED` causes a crash on some compositors/systems as shown by #1824. According to WL docs:
> From version 7 onwards, the fd must be mapped with MAP_PRIVATE by the recipient, as MAP_SHARED may fail.

We don't need a `MAP_SHARED` as the provided buffer will be used exclusively by conky and not shared with other processes.

This PR addresses that and closes #1824.

# Testing
- Tested private mapping in another Wayland client and it worked with hyprland.

PR #2103, opened 2024-12-07
Fix out_to_wayland giving nothing

stacyharper:
This fix a problem caused by 6adf6b9dd4d368640bf7ef57f3e6199d83154d78:

The display outputs are sorted by priority on src/display-output.cc.

Before the cleanup commit, Wayland was choosed instead of X11. But no thanks to any explicit code. Just because outputs was inserted luckily in the correct order, before the sort.

Because out_to_x is true, and out_to_wayland is false by default, it means the users have to enable explicitely wayland on their config.

We should rather prefer Wayland because it need explicit configuration.

To do so, I think we should define that Wayland priority surpass the X11 one, instead of fixing the register_output order.

(There is yet another issue with Wayland output at the moment (caused by https://github.com/brndnmtthws/conky/pull/1876#issuecomment-2525044908). I'm investigating it. And hopefully, two patch are needed to fix most of the Wayland related tickets)

# Checklist

### DIFF
--- a/src/display-output.cc
+++ b/src/display-output.cc
@@ -80,22 +80,24 @@ std::vector<conky::display_output_base *> current_display_outputs;
 bool initialize_display_outputs() {
   std::vector<display_output_base *> outputs;
   outputs.reserve(static_cast<size_t>(output_t::OUTPUT_COUNT));
-  register_output<output_t::CONSOLE>(outputs);
-  register_output<output_t::NCURSES>(outputs);
-  register_output<output_t::FILE>(outputs);
-  register_output<output_t::HTTP>(outputs);
-  register_output<output_t::X11>(outputs);
+
+  // Order of registration is important!
+  // - Graphical outputs go before textual (e.g. X11 before NCurses).
+  // - Optional outputs go before non-optional (e.g. Wayland before X11).
+  // - Newer outputs go before older (e.g. NCurses before (hypothetical) Curses).
+  // - Fallbacks go last (in group)
   register_output<output_t::WAYLAND>(outputs);
+  register_output<output_t::X11>(outputs);
+  register_output<output_t::HTTP>(outputs);
+  register_output<output_t::FILE>(outputs);
+  register_output<output_t::NCURSES>(outputs);
+  register_output<output_t::CONSOLE>(outputs);  // global fallback - always works
 
   for (auto out : outputs) { NORM_ERR("FOUND: %s", out->name.c_str()); }
-
-  // Sort display outputs by descending priority, to try graphical ones first.
-  sort(outputs.begin(), outputs.end(), &display_output_base::priority_compare);
 
   int graphical_count = 0;
 
   for (auto output : outputs) {
-    if (output->priority < 0) continue;
     DBGP2("Testing display output '%s'... ", output->name.c_str());
     if (output->detect()) {
       DBGP2("Detected display output '%s'... ", output->name.c_str());
--- a/src/display-console.cc
+++ b/src/display-console.cc
@@ -50,10 +50,7 @@ void register_output<output_t::CONSOLE>(display_outputs_t &outputs) {
 }
 
 display_output_console::display_output_console(const std::string &name_)
-    : display_output_base(name_) {
-  // lowest priority, it's a fallback
-  priority = 0;
-}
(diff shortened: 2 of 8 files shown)


brndnmtthws:
Sounds reasonable to me, but I will defer to @Caellian on this who knows the Wayland details better than me.

src/display-output.cc
```diff
@@ -80,22 +80,24 @@ std::vector<conky::display_output_base *> current_display_outputs;
...
 
   for (auto out : outputs) { NORM_ERR("FOUND: %s", out->name.c_str()); }
 
-  // Sort display outputs by descending priority, to try graphical ones first.
-  sort(outputs.begin(), outputs.end(), &display_output_base::priority_compare);
-
   int graphical_count = 0;
 
   for (auto output : outputs) {
-    if (output->priority < 0) continue;
```
Caellian, line 98: I completely removed `priority` and made ordering depend on registration order. This achieves the same result, but makes the code a lot easier to read/navigate as opposed to having this number used by a single sort scattered across half a dozen files. It also makes the ordering much more apparent and explicit.

Caellian (approved):
Thank you for tracking this down, and sorry for inadvertently breaking things.

stacyharper:
@Caellian we still have to revert commit 940a365ada86cbf565381898e57f78c0f0a219ee to finally fix Wayland on layer shell output.

Caellian:
That _does_ make conky appear for me on hyprland.

stacyharper:
btw @Caellian: own_window seems to still give a blank window surface to me. Anything else is borked, or have I just miss-configured conky?

Caellian:
`own_window` doesn't do anything AFAICT, only enables an empty `if` statement. It will switch between layer shell and XDG shell at some point, but not now.

Send your config and WM so I can compare, my config is:

<details>

```lua
--
-- configurations.
--
--
--
-- For more on Lua, see:
--

conky.config = {
    alignment = 'top_left',
    background = false,
    border_width = 1,
    cpu_avg_samples = 2,
    default_color = 'white',
    double_buffer = true,
    draw_borders = false,
    draw_graph_borders = true,
    draw_outline = false,
    draw_shades = false,
    extra_newline = false,
    gap_x = 60,
    gap_y = 60,
    minimum_height = 5,
    minimum_width = 5,
    net_avg_samples = 2,
    no_buffers = true,
    out_to_wayland = true,
    own_window = false, -- true,
    own_window_class = 'Conky',
    own_window_type = 'normal',
    own_window_hints = 'undecorated,sticky,below,skip_taskbar,skip_pager',
    show_graph_range = false,
    show_graph_scale = false,
    stippled_borders = 0,
    update_interval = 1.0,
    uppercase = false,
    use_spacer = 'none',
    use_xft = true,
    color0 = "ff0000",
}

-- Variables: https://conky.cc/variables
conky.text = [[
${color grey}Info:$color ${scroll 32 Conky $conky_version - $sysname $nodename $kernel $machine}
$hr
${color grey}Uptime:$color $uptime
${color grey}Frequency (in MHz):$color $freq
${color grey}Frequency (in GHz):$color $freq_g
${color grey}RAM Usage:$color $mem/$memmax - $memperc% ${membar 4}
${color grey}Swap Usage:$color $swap/$swapmax - $swapperc% ${swapbar 4}
${color grey}CPU Usage:$color $cpu% $color0 ${cpubar 4} $color
${color grey}Processes:$color $processes  ${color grey}Running:$color $running_processes
$hr
${color grey}File systems:
 / $color${fs_used /}/${fs_size /} ${fs_bar 6 /}
${color grey}Networking:
Up:$color ${upspeed} ${color grey} - Down:$color ${downspeed}
$hr
${color grey}Name              PID     CPU%   MEM%
${color lightgrey} ${top name 1} ${top pid 1} ${top cpu 1} ${top mem 1}
${color lightgrey} ${top name 2} ${top pid 2} ${top cpu 2} ${top mem 2}
${color lightgrey} ${top name 3} ${top pid 3} ${top cpu 3} ${top mem 3}
${color lightgrey} ${top name 4} ${top pid 4} ${top cpu 4} ${top mem 4}
]]

```

</details>

stacyharper:
Right, that was I thought looking at the code. Just wanna confirm :) Thanks for all you help!

Issue #2106, opened 2024-12-07
[Feature]: Support specifying multiple `display`s in Wayland


abkein:
### What happened?

Here is the picture:
![swappy-20241208-015609](https://github.com/user-attachments/assets/c10c2a50-3ecf-49b8-af45-e3b50119940f)
As you can see, pdf document is present on both screens, but the conky is not. If I increase `gap_x` further (bigger than monitor's x-resolution), conky becomes fully invisible.
The same if open conky on the second monitor and set `gap_x` negative:
![swappy-20241208-020241](https://github.com/user-attachments/assets/933cc83f-fbe5-4737-ad12-0323d95963bb)

The main problem is that conky opens on the monitor the mouse was on, so usage of relative positioning becomes hard or impossible, if you want the conky to be at constant position.

Platform:
```
Hyprland 0.45.2 built from branch  at commit 12f9a0d0b93f691d4d9923716557154d74777b0a  ([gha] Nix: update inputs).
Date: Tue Nov 19 21:47:18 2024
Tag: v0.45.2, commits: 5451
built against aquamarine 0.5.0

OS: Arch Linux x86_64
Kernel: Linux 6.12.1-arch1-1
Display (Mi monitor): 3440x1440 @ 60 Hz in 35" [External]
Display (SA240Y): 1920x1080 @ 75 Hz in 24" [External]
Display (TMX2005): 3200x2000 @ 120 Hz (as 2003x1252) in 16" [Built-in]
```

### Version

Latest commit 28a39cb26c190e94ebfbb02ce9f15c85043943b1 built with [these flags](https://github.com/abkein/conky/blob/8840dec95e0a515fffb429f400befee0a2436b9d/CMakePresets.json)

### Which OS/distro are you seeing the problem on?

Arch Linux

### Conky config

```Lua
se,
    show_graph_scale = false,
    stippled_borders = 0,
    update_interval = 1.0,
    uppercase = false,
    use_spacer = 'none',
--    use_xft = true,
}

-- Variables: https://conky.cc/variables
conky.text = [[
${color grey}Info:$color ${scroll 32 Conky $conky_version - $sysname $nodename $kernel $machine}
$hr
${color grey}Uptime:$color $uptime
${color grey}Frequency (in MHz):$color $freq
${color grey}Frequency (in GHz):$color $freq_g
${color grey}RAM Usage:$color $mem/$memmax - $memperc% ${membar 4}
[...]

PR #2110, opened 2024-12-08
Simplify adding new wayland protocols

Caellian:
- Moved wl protocol files to separate directory in src, as src dir is already waaay to crowded.

Tested adding `xdg-output` - which will be needed for fixing stuff like #2106


src/CMakeLists.txt
```diff
@@ -273,34 +273,70 @@ if(BUILD_GUI)
...
+      OUTPUT ${CMAKE_CURRENT_BINARY_DIR}/${name}-protocol.c
+      COMMAND ${Wayland_SCANNER} private-code "${PROTOCOL_FILE}" "${name}-protocol.c"
+      DEPENDS "${name}-client-protocol.h")
+    list(APPEND wl_sources "${name}-protocol.c")
+    unset(PROTOCOL_FILE)
+    unset(WL_PROTOCOL_PATHS)
+  endmacro()
+  
+  add_protocol(xdg-shell)
+  add_protocol(wlr-layer-shell v1)
```
Caellian, line 337: This PR allows protocols to be added like this.

brndnmtthws (approved):
Very nice!

Issue #2112, opened 2024-12-08
[Bug]: dock alignment doesn't work after setting initial state to withdrawn


Bluey26:
### What happened?

For conky versions newer than `1.21.7` there seems to be a X11 regression in where the conky window is not showing itself anymore in the given position. `alignment = 'top_right'` seems to be ignored, and the window is shown in the top_left border, transparency is not detected anymore (`own_window_argb_value = 90` ignored) and there is also a "windowing" issue, since the `own_window_type = 'dock'` property is also ignored (the conky panel "reserves" the space and avoids other windows and the desktop itself to be on top of it.

I discovered this after compiling the lastest changes in conky, but it seems to be a problem arisen from prior changes. Downgrading my conky to `conky-git-1.21.7.r5.g71814776-1` seems to solve the issue, for some reason (x11 backend now has all the mentioned properties working).

This seems to be independent of the wayland_support argument. For some reason, the issue is not shown when using the **wayland backend** in a Wayland session (i.e. labwc)

**PS**: I don't think that my conky config is relevant in this case, because the .conf is the same for the wayland backend(1.21.9) and for the previous version (1.21.7), in both cases giving me a working conky.

**PS2:** Since there is a WIP wayland_backend that works now, i was thinking if it would be possible to have a conky property that lets the user to have a "preference" over the backend use, something like : `prefer_backend = wayland` or `prefer_wayland = 1`. This is an issue because if somebody uses x11 and wayland from time to time,  `out_to_wayland = true, out_to_x = true,` will give the x11 (xwayland) backend in wayland instead of the newer wayland backend.

### Version

Lastest (1.21.9)

### Which OS/distro are you seeing the problem on?

Arch Linux

Caellian:
Related PR in 1.21.8 is #2056.
But I believe that's not the cause of the issue and only unearths some other problem.
Can you check whether you're experiencing issue described in #2046 on 1.21.7?

Commits that touch `set_struts`:
- c5ee70177c7d9b586707cf1adefdc45894c5cdea
  - first change by me that could cause issues; set_struts rewrite
- 7b44b2a27bd0649c7c00a4f5f4ec521aa554777b
  - fixes missing variables from previous change
- 6bf383cf786a9615480b0268571cb2ad84e2fd24
  - introduces geometry primitives which get used in struts
- 2d50767216d2555b2d8d62765abaf41cb6d6979a (last in 1.21.7)
  - adds utilities to geometry primitives
[...]

Issue #1961, opened 2024-06-07
[Bug]: set_struts wrong with large DPI


Caellian:
Reported by @zygfryd

> Desktop reserved space (in panel mode) seems to be double the actual window size.

When `Xft.dpi: 192` (2x default).

After testing with [`test/workaread-adjustments`](https://github.com/brndnmtthws/conky/commit/cf177e11195d890efff4db60a82c2f9eb69d11e9) which applies DPI left and applies inverse on the right margin,
> Left alignment results in 3x reserved space (tested by maximizing a window), right alignment results in no reserved space.

Not sure whether right alignment can be fully trusted in the test because it could be affected by left alignment. It looks like `window.geometry` stores premultiplied values, and X11 expects logical position.

Related function: [x11.cc/set_struts](https://github.com/brndnmtthws/conky/blob/f88b50d21a86e37da6877c809fbabdc87ab30d4e/src/x11.cc#L1135)

Broken by [#1910](https://github.com/brndnmtthws/conky/pull/1910), [#1862](https://github.com/brndnmtthws/conky/pull/1862), or [#1849](https://github.com/brndnmtthws/conky/pull/1849).

zygfryd:
I just realized that right alignment wouldn't have worked because of having multiple monitors and testing on the left monitor. I edited inverse scaling on the left side, and that worked correctly.

Caellian:
A PR affected code this issue relates to: #1963

Issue #1917, opened 2024-05-15
[Bug]: 'panel' struts don't work with multi-monitor setup


brrrrrrrt:
### What happened?

i use two monitors.

conky window type is set to panel an positioned on the left side of the right monitor.

After Upgrading to 1.20 the window placement is "broken" when using the window type "panel".
once conky is started, all active windows move from the left to the right monitor and i am not able to move them all the way to the left on the left monitor, like some invisible barrier. the mouse can be moved freely on all monitors, but the windows cannot be moved all the way to the left.

downgrading to 1.19 solves the problem.

window type desktop works, this is how it should look like:
![Bildschirmfoto vom 2024-05-15 14-47-03](https://github.com/brndnmtthws/conky/assets/4489013/1a7b9612-63f1-429c-a05b-3fd3b6904a17)

after setting window type to "panel" it looks like this:
![Bildschirmfoto vom 2024-05-15 14-47-28](https://github.com/brndnmtthws/conky/assets/4489013/27fe11bd-b02b-448e-bb18-65fbe26e66b0)

windows cannot be moved all the way to the left, it stops at this position:
![Bildschirmfoto vom 2024-05-15 14-48-01](https://github.com/brndnmtthws/conky/assets/4489013/9af0dca2-337f-4f75-b95e-d7c31d62ed34)

### Version

1.20.2_pre

### Which OS/distro are you seeing the problem on?

Arch Linux

### Conky config

```Lua
-- vim: ts=4 sw=4 noet ai cindent syntax=lua

conky.config = {
    alignment = 'top_left',
    background = false,
    border_width = 1,
    cpu_avg_samples = 2, 
	default_color = '606060',
    default_outline_color = 'white',
    default_shade_color = 'black',
    double_buffer = true,
    draw_borders = false,
    draw_graph_borders = true,
    draw_outline = false,
    draw_shades = false,
    gap_x = 1921,
    gap_y = 20,
    minimum_height = 10,
	minimum_width = 330,
	maximum_width = 330,
    net_avg_samples = 2,
    no_buffers = true,
    out_to_console = false,
    out_to_stderr = false,
    extra_newline = false,
    own_window = true,
	own_window_transparent = true,
	own_window_argb_visual = true,
	own_window_type = 'panel',
	own_window_hints = 'below',
	short_units = true,
    stippled_borders = 2,
    update_interval = 1.5,
    uppercase = false,
    use_spacer = 'right',
    show_graph_scale = false,
[...]

PR #2115, opened 2024-12-09
Add support for mocking to tests & test X11 struts

Caellian:
This PR adds some very basic mocking capabilities to tests.

The goal of this PR is to test values set by `set_struts`.

Mocking was forced on me because `set_struts` calls `XChangeProperty` which obviously wouldn't work without X11 initialization code.

Mocking works by linking mocked methods before `conky_core`.

I added a basic system to allow retrieving mutations that mocked functions perform (e.g. defining atoms, `XChangeProperty` information, etc.).


src/gui.h
```diff
@@ -167,7 +167,7 @@ extern char window_created;
 
 void destroy_window(void);
 void create_gc(void);
-void set_struts(int);
+void set_struts(alignment);
```
Caellian, line 170: - [ ] I kept this line like this for a reason, it previously caused compilation to fail because `alignment` wasn't available to some file (wayland?) that uses it. Note for self to check before merge.

Caellian:
While mocking would be great, I'm putting this off for now because Lua options are _very_ difficult to mock the way they currently work. Constructor was the issue and a more data-oriented approach would greatly help.

However, the testing I did manage to run didn't indicate any issues so far and would only serve for regression testing.

The issues (#1917, #1961, #2112) likely come from how different WMs handle struts as there are some differences. To deal with that, conky would need to detect WM and use correct implementation at runtime.

<details>
<summary>New set struts documentation</summary>

```cpp
/// Reserve window manager space
///
/// Both `_NET_WM_STRUT` and `_NET_WM_STRUT_PARTIAL` work in coordinates
/// relative to root window (or sometimes Xinerama/Screen).
///
/// Values tell the WM which regions of the screen are invalidated. So, a
/// `bottom` value of `30` means that the window reserves bottom 30px of the
/// screen.
///
/// Because struts aren't handled the best by all WMs when multiple screens are
/// used, horizontal struts (top, bottom) should be preferred because that
/// works well for most multi-screen (horizontal monitor stacking) setups.
/// See: https://gitlab.gnome.org/GNOME/mutter/-/issues/452
///
/// Different WMs handle this differently, some adhere to the spec, some don't.
/// Spec compliant (relative to root window edges):
/// - mutter, metacity, openbox, marco, xfwm
/// Non-compliant (relative to edges of xinerama/single monitor):
/// - compiz, kwin, i3, fluxbox
///
/// Article why KWin doesn't follow the spec:
/// https://blog.martin-graesslin.com/blog/2016/08/panels-on-shared-screen-edges/
```

</details>

brndnmtthws:

Caellian:
Too broad and vague.

Besides being out-of-date, it's trying to implement a mocking framework for X11.

Besides a useful comment, there's nothing I find worth merging today.

PR #2119, opened 2024-12-16
[Refactor]: Move sources into subdirectories

Caellian:
A refactor that moves sources from `src/` directory into appropriate subdirectories.

This PR contains no changes to conky functionality.

I wanted to do this ever since my first commit :laughing:, but yeah, I feel like the `src/` directory is too crowded and very difficult to navigate. To the extent that it takes me a minute of scrolling sometimes before I can find the source file I'm looking for - and I'm familiar with this project.

When I first started I remember having to navigate though include statements.

## TODO
- [x] Rename files in a separate commit
- [ ] Fix build files and include statements


Caellian:
@brndnmtthws can you please confirm whether I should continue with this PR?

I made sure to separate commit with renames so this PR can be hidden in `.git-blame-ignore-revs` and rebasing should work out well for other active PRs?

Caellian:
Tested all include path with a [script](https://gist.github.com/Caellian/619d11b0b83cb047164ebcc11cee2ed0), it only yells out generated headers and library includes, so this should be good to go.

Caellian:
I lo-lo-love how GH was like "oh, a commit signature?! how... unmicrosofty of you" \*throws it in the trash\*

PR #2127, opened 2024-12-31
wayland: fix clean_text bug

bl4kraven:
The current wayland display's clean_text function is missing cairo_fill to actually clear the background, causing the text to overlap.

The cairo_rectangle just adds a sub-path rectangle and doesn't draw , you need to call cairo_fill to actually draw it.


Caellian:
Good change, not sure how I missed this. Sorry for late response.

PR #2133, opened 2025-01-05
Update blame ignore & labeler

Caellian:
This is a patch for #2119 PR, that adds "copied" commits to the `.git-blame-ignore-revs` file.

Also updates labeler.


brndnmtthws (approved):
Much cleaner, thanks!

Caellian:
> Much cleaner, thanks!

@brndnmtthws It didn't take me as long as I initially expected to fix all the paths in CMake. I got an idea of using a script to do it automatically and then generated it with an LLM. It was wrong, of course, but took me 95% of the way there. Then generated the other script to update import statements. Most difficult part was not getting carried away and touching up random files.

PR #1899, opened 2024-05-04
Improve logging and termination

Caellian:
This PR replaces conky's current logging (`NORM_ERR`, `DBGP`, `fprintf(stderr, ...)`) with a structured logging framework built on [spdlog](https://github.com/gabime/spdlog) (vendored v1.17.0).

## Motivation

- `CRIT_ERR` used `exit()` for everything (user errors, system errors, bugs) - no core dumps, no way to distinguish fault.
- Too few log levels: `NORM_ERR` was used for info, warnings, and errors alike.
- Error messages were often vague ("something wrong with /proc/i8k") and missing context.
- No structured context for debugging nested operations.
- Logs were only written to stderr with no journal support.

## Changes

- **spdlog framework** - Vendored spdlog 1.17.0 as a static library with exceptions disabled. Added `LOG_*` macros with source location tracking, runtime level control (`-D`/`-L`/`-q`), and `--concise` flag.

- **Termination handler** - `std::set_terminate` with `handle_terminate()` for graceful cleanup. `abort()` → `std::terminate()` in semaphore. Structured error macros (`USER_ERR`, `SYSTEM_ERR`, `COMMAND_ARG_ERR`, `CRIT_ERR`) that classify fault and log before throwing/terminating.

- **Error reclassification** - ~80 `CRIT_ERR` call sites reclassified by fault: user input errors throw (recoverable), system failures throw (recoverable), internal bugs terminate with core dump. Printf format strings converted to fmt `{}` syntax.

- **Exception cleanup** - Removed exception-as-control-flow patterns (`fork_throw`, `unknown_arg_throw`, `combine_needs_2_args_error`). Added try-catch around variable parsing for graceful error recovery (renders `${bad_var}` as literal instead of crashing).

- **Mass logging update** - All `NORM_ERR`/`DBGP`/`DBGP2`/`fprintf(stderr)`/`perror()` calls converted to `LOG_*` with proper level classification and improved messages. Error messages enriched with context values. Legacy macros removed.

- **Span context** - Scoped spans (`LOG_SCOPE`) with key-value attributes for structured diagnostic context. Zero-cost when disabled (no allocation, no formatting at info level). Per-message attributes via `LOG_*_WITH` macros with compile-time and runtime guards.

## Log output

<details>
<summary>Info level (default)</summary>
<pre>
[2026-04-28 09:32:31.289] [info ] [src/output/x11.cc:553] window type - override
[...]

PR #2138, opened 2025-01-10
[Refactor]: Cleanup CMake & separate libconky for all builds

Caellian:
This PR does a lot and is a pain to review, I'm making it to "save progress". I'll highlight important changes once I'm done and it's no longer a draft.

# Summary

- Separate libconky from conky for all builds to allow lua modules to access conky internals.
- Remove global variables that carried build sources/libs/dirs/options and replace them with targets.
- Use targets for platform testing to ensure consistent results when building.
- Replace build options stored in headers with compile definitions (stored in a target).
- Remove some old files that weren't being used.

# Rationale

## Targets instead of variables

Mutating global state with `include_directories` is irreversible and makes it hard to debug/maintain the build files. It also leaks into subprojects and then all projects end up having access to everything which allows including headers (like "logging.h") which shouldn't be accessible from wrong places.
Having these stored in targets allows prepending include/library paths to patch things - like `X11/X.h` defining `Always` and `None`, which breaks tests if catch2 is included after some conky headers.

## Removing config.h

Storing all these variables in `-DFLAG=VALUE` is convenient when targets are used, maintaining two separate lists of options (CMake and config.h.in) doesn't really provide any added benefit and is just annoying. There were already 2/3 options that were left behind and unused. Having the `ConkyBuildOptions.cmake` populate these automatically is super nice.

Also, as all options now use basically the same command:
```cmake
conky_option(
  <option> "<help_text>" [<initial_value>]
  [DEPENDS <conditions>]
  [FALLBACK <value>]
  [SHOW_IF <conditions>]
  [TYPE (BOOL|FILE|DIR|STRING|INT|FLOAT)]
  [WARN "<warning>"]
)
```
it's possible to create a `build_options.yaml` that can be used for both docs and actual compile-time configuration.

## Always using libconky

This comes up in #1899. Basically, `cairo_imlib2_helper` wants to access logging, but it can't because that would require bundling entire conky with it as #1899 adds .cc files (for global static LOGGER).
To deal with this, I think it's simplest to always separate libconky with PIC from main executable - this allows lua modules to access conky internals (like color parsing) without having to statically link to and include all conky sources. Alternatively, we could be binding a lot of the functions to Lua in order to make them accessible to modules, but that's a lot of work.

# Work left to do

- [ ] Fix lua module linking, currently static.
- [ ] `DEPENDENT_OPTION` function doesn't seem to set correct fallback value when `DEPENDS` isn't met.


Caellian:
Closing this in favour of smaller, targeted PRs. The ideas here are good but the diff is too large to review safely and there are (probable) bugs I located using Claude LLM that affect correctness on all platforms. It was _WIP_, but I can't continue 112 file PR one year later without introducing problems (the author was being a goof).

**What's worth carrying forward:**

<sup>split suggested by Claude Sonnet 4.6</sup>

**PR 1 — `conky-core` as a shared library + `conky-dependencies` interface target**
This is the actual payload of this branch and the fix for #1899. The change is: `add_library(conky-core SHARED ...)` in `src/CMakeLists.txt`, replacing the global `conky_libs`/`conky_includes` variable accumulation with `conky-dependencies INTERFACE`, and splitting `src/main/` out so the executable is just `main.cc` linking against `conky-core`. Everything else in this PR is separable from this. The X11/Catch2 header ordering fix (the `Always`/`None` macro collision) also belongs here since it's the immediate motivation for `conky-dependencies`.

**PR 2 — `DependentOption.cmake` rewrite**
The deferred dependency resolution via `run_dependency_checks()` is a genuine improvement over `cmake_dependent_option`. However the current implementation has two confirmed bugs that need fixing before it can be merged:

- `SHOW_IF` logic is inverted — `__${option}_VISIBLE` is initialised to `0` and set to `1` when the condition is *false*, making it behave as `HIDE_IF`. `BUILD_PORT_MONITORS SHOW_IF "OS_LINUX"` currently hides on Linux and shows on Darwin.
- `set(arg_FALLBACK INITIAL_VALUE)` stores the string literal `"INITIAL_VALUE"` rather than the variable's value (missing `"${...}"`). The warning message also references `${__DEPENDENT_OPTIONS_FALLBACK}` which is never set.

**PR 3 — `conky-definitions` replacing `config.h`**
The `source_define()` function and the `conky-definitions` interface target are a clean replacement for the dual-maintenance problem with `config.h.in`. Can be done independently once PR 1 lands.

Notes on bugs:
- `include(Conky)` must come after `include(OSDetection)` in the root `CMakeLists.txt` — `Conky.cmake` uses `OS_OPENBSD` to conditionally add `-stdlib=libc++` to `conky-options`, but currently evaluates it before OpemBSD detection sets the variable, causing all platforms to get the flag unconditionally.
- Changed CMake dynamic eval code is very buggy. It's needed to generate definitions correctly based on settled feature graph state. Needs work.

Caellian:
Don't delete the branch, few chunks of code might be useful.

PR #2153, opened 2025-02-02
Check whether first_colour is set before using it as foreground color

livanh:
# Checklist

## Description

For several months now, the first color of a graph's gradient is also being used as the foreground color, but without properly checking whether the gradient colors are actually set. If they are not, the scale text of the graph is set to black instead of the default color:
immagine

With this change, the scale text correctly follows the default color:
immagine

(I don't think the documentation needs an update, but I will do it if requested)


Caellian:
@livanh thank you for your contribution.

PR #2154, opened 2025-02-03
Use semicolon as file path separator for lua_load

Caellian:
`lua_load` previously expected space separated list of file names.

As spaces in file names do happen even to the best of us, I made `lua_load` handle `;` separators on top of space separators.

Basically, if string contains `;`, new logic will be used, if not, old logic will be used. Additionally, empty paths are skipped.
This allows for values like `./my file has spaces.lua;` which would've been impossible to load previously.

As the fallback is space separator, this shouldn't break existing configs that use multiple files already. I added a TODO for 2028 to remove the branch for old behavior.

Lastly, previous code copied every item into a separate string and then turned it back into a c style string, new code replaces appropriate delimiter with `\0` and uses the same chunk of memory.


brndnmtthws (approved):

LinuxOnTheDesktop:
&nbsp;&nbsp;&nbsp;&nbsp;Thanks for this. However, seemingly it does not improve readability. For, so far as I can tell, one needs still to have the list of lua scripts as one big, and single-line, quoted string. Indeed, if one replaces the spaces - acting as separators - with semi-colons, then the result is even more unreadable than before.

&nbsp;&nbsp;&nbsp;&nbsp;What is wanted, if I may say so, is the ability to do something like the following.

```
lua_load =
   <fullPathToFile1>;
   <fullPathToFile2>;
   <fullPathToFile3>;
   <fullPathToFile4>;
   <fullPathToFile5>
```

Or

```
lua_load = {
   [[fullPathToFile1]],
   [[fullPathToFile2]],
   [[fullPathToFile3]],
   [[fullPathToFile4]],
   [[fullPathToFile5]]
}
```

Caellian:
@LinuxOnTheDesktop yes, ideally it would just be a normal Lua table as far as data structure design goes, i.e.:

```lua
lua_load = {
  "path_1",
  "path_2",
  "path_3"
}
```

but, that's very annoying to do with current design - I'd have to introduce yet another class for handling a value that can be either a table or a string for this specific use case (like some other settings do).

The point of this PR was making it possible (straightforward) to specify file paths that contain spaces, not so much to make the config look better. Not sure why space character was used as an entry separator for file names in the first place, but a semicolon makes far more sense and is more common.

You can still use spaces until deprecation period of 3 years passes, if your file names don't contain semicolons that is.

PR #2167, opened 2025-02-20
Replace wordexp with custom implementation

Caellian:
This PR replaces wordexp with custom implementation for all platforms. Precursor of this implementation was used on platforms without `wordexp`, so this PR consolidates the two into one.

Linux version will loose ability to use some less commonly used expansions such as [field splitting](https://www.ibm.com/docs/en/i/7.3?topic=expansions-field-splitting). The important things will be reimplemented, the removed stuff is basically things that don't make sense i.e. only work well in terminal or shell scripts (e.g. field splitting depends on IFS env variable which isn't controlled by conky anyway; some other things that).

BSDs, Solaris, etc. will get much better handling of paths - they'll be able to use command expansion and math expansion.

## TODO

- [x] Add command expansion.
- [ ] Finish wildcard expansion.
- [ ] Implement math evaluation by expanding `algebra.cc` with code from [`tinyexpr`](https://github.com/codeplea/tinyexpr)

### DIFF
--- a/src/common.cc
+++ b/src/common.cc
@@ -39,9 +39,12 @@
 #include <unistd.h>
 #include <cctype>
 #include <cerrno>
+#include <cstdint>
 #include <cstdio>
 #include <cstdlib>
 #include <ctime>
+#include <filesystem>
+#include <string>
 #include <vector>
 
 #include "config.h"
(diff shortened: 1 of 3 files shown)


brndnmtthws:

PR #2168, opened 2025-02-26
Added multi pointer support

fluffyironboots:
Conky currently can't handle a second master pointer device. The second master only works while the main master is in the conky window. Reason for the change was a touchscreen as system monitor.

My changes are mainly a replacement of XQueryPointer with XIQueryPointer to be able to select the window with a device id.


Caellian (changes requested):
Thank you for your contribution!

Here's a few minor changes, mostly nitpicks, but `query_x11_window_at_pos` one is important.

Caellian (approved):
LGTM

Issue #2156, opened 2025-02-07
[Bug]: own_window_type = 'panel' broken since V1.21.1 and newer


Der-Schubi:
### What happened?

Hi!
I'm using Arch with Cinnamon on X11.
After updating Conky, own_window_type = 'panel' is not working anymore.
My original config file is much bigger, but I did a few tests with this minimal config.

It works fine with Conky V1.21.0 and starts behaving weird since V1.21.1:
Other windows can not go higher than the y-position of the Conky window, which is vertically in the center of the screen.
In the newest 1.22.1-pre-35c5a052 windows maximize right behind the Conky window.

### Version

1.22.1

### Which OS/distro are you seeing the problem on?

Arch Linux

### Conky config

```Lua
conky.config = {
  out_to_x = true,
  out_to_wayland = false,

  own_window = true,
  own_window_type = 'panel',

  alignment = 'right',

  xinerama_head = 1,

  use_xft = true,
};

conky.text = [[
${font Digital Readout Thick Upright:size=54}${alignc}${time %H:%M}${font}
${font Digital Readout Thick Upright:size=19}${alignc}${time %a %d.%m.%Y}${font}
]];
```

### Stack trace

```GDB

```

### Relevant log output

```Shell
conky: 'cinnamon' x11 session running 'X-Cinnamon' desktop
```

Caellian:
Can you give a try to [`fix/struts`](https://github.com/brndnmtthws/conky/tree/fix/struts) branch and let me know how the windows behave with new changes?

Der-Schubi:
I moved to Wayland in the meantime, but I'll try with a live CD. Please give me some time.

Issue #1085, opened 2021-04-12
Multiple screen panel type window bug


AyhamAl-Ali:
**Issue**

I am following [this idea](https://askubuntu.com/a/1257827) to make a custom window maximize size (a bit weird idea but yea this works unintentionally)

The problem is when I make a `middle-right` alignment panel type my second screen/display which is located on the right of the main screen will have an empty half bottom space (as seen in the screenshot below) which makes my window maximize for my second screen to half size.

**Information**

OS: **Pop!_OS 20.10 x86_64**
Kernel: **5.11.0-7612-generic**

<details><summary>Middle-Right panel config</summary>
<p>

```
conky.config = {
    alignment = 'middle_right',
    background = false,
    border_width = 0,
    cpu_avg_samples = 2,
    default_color = 'white',
    default_outline_color = 'white',
    default_shade_color = 'white',
    double_buffer = true,
    draw_borders = false,
    draw_graph_borders = true,
    draw_outline = false,
    draw_shades = false,
    xinerama_head = 0,
    extra_newline = false,
    font = 'DejaVu Sans Mono:size=12',
    gap_x = 15,
    gap_y = 0,
    minimum_height = 0,
    minimum_width = 0,
    net_avg_samples = 2,
    no_buffers = true,
    out_to_console = false,
    out_to_ncurses = false,
    out_to_stderr = false,
    out_to_x = true,
    own_window = true,
    own_window_class = 'Conky',
    own_window_type = 'panel',
    own_window_transparent = true,
    own_window_argb_visual = true,
    own_window_hints = 'undecorated,below,above,sticky,skip_taskbar,skip_pager',
    show_graph_range = false,
    show_graph_scale = false,
    stippled_borders = 0,
    update_interval = 1.0,
    uppercase = false,
    use_spacer = 'none',
    use_xft = true,
}

conky.text = [[ ]]

```

</p>
</details>
Other panels are the same config just different alignment

![image](https://user-images.githubusercontent.com/20037329/114354478-f7286d00-9b76-11eb-8293-45cb3cd9696e.png)

AyhamAl-Ali:
Have this been solved?

AyhamAl-Ali:
Damn, it has been 2 years now! Well, I am not sure if this was solved as I left the project and switched to Gnome's extensions that has full functionality for this.

Caellian:
Sorry you've been waiting so long, the reason this issue is "help wanted" is because it requires a second monitor to debug and fix.

[...]

PR #2190, opened 2025-04-26
Improve X11 strut handling

Caellian:
This PR reworks strut code to use compute which side of the screen struts should be placed on for conky, regardless of alignment settings.

This makes more sense because it will always pick the side where windows will have most space left. Most WMs have very simplistic approach to calculating workarea (which is spec compliant):
- [Openbox](https://github.com/Mikachu/openbox/blob/dac6e2f6f8f2e0c5586a9e19f18508a03db639cb/openbox/screen.c#L1428)
- [Fluxbox](https://github.com/fluxbox/fluxbox/blob/88bbf811dade299ca2dac66cb81e4b3d96cfe741/src/HeadArea.cc#L72)
- [Mutter](https://gitlab.gnome.org/GNOME/mutter/-/blob/ea8b65d0c92ebf84060ea0b5c61d02fa9004e1e9/src/core/boxes.c#L564)
- [xfwm](https://github.com/xfce-mirror/xfwm4/blob/37636f55bcbca064e62489d7fe183ef7b9371b4c/src/placement.c#L220)

So this will avoid conky eating up too much space for those.

The [spec](https://specifications.freedesktop.org/wm-spec/1.4/ar01s05.html#id-1.6.11) says:
> ... All coordinates are root window coordinates. ...

I previously used workarea size for computing strut bounds, which is wrong if there's multiple panels with struts on the same display axis (they will be subtracted and the right/bottom aligned struts will be incorrectly placed).

EWMH spec [will likely never be changed](https://gitlab.freedesktop.org/xdg/xdg-specs/-/merge_requests/22) to allow for use cases such as #1917. Some WMs like KWin (KDE; Plasma) and compiz have special handling here and cut out screens from struts in case some strut covers entirety of them (#1917 use case works for those). Some simply _ignore_ the struts if they cover more than certain amount of screen space. See [details](https://gitlab.gnome.org/GNOME/mutter/-/issues/452).

In order to support WMs that _do_ handle cutout behaviors, I added system information to `info` - so now we can account for edge cases like this at runtime. There's also a build option to turn that off, but it's enabled by default to ensure default build works best for most people.

Closes: #2156, #1085, maybe #1961 and #2112.

Tested on single screen and behaves as expected on:
- Cinnamon (mutter), GNOME (mutter)
- XFCE (xfwm)
- Openbox
- KDE (KWin)

Updated the docs to fix the wrong `dock` description stating that dock reserves screen are when that isn't the case. New, WM-dependent behavior is also added to documentation.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -11,6 +11,9 @@
 ### Reformatted all code (SVN@1007)
 3d26a4880e92df2c6004d85c1be458e35c6bfc3a
 
+# Spelling fixes
+b62e1158300f0644b304f2f8b7e8105ba09073b3
+
 ### outsource the whole template object machinery
 # Moves template variables code from core.c into template.c
 ff199355f66216600c4bdc6bec4743afe5b61470
--- a/cmake/ConkyBuildOptions.cmake
+++ b/cmake/ConkyBuildOptions.cmake
@@ -94,12 +94,12 @@ option(BUILD_EXTRAS "Build extras (includes syntax files for editors)" false)
 
 option(BUILD_I18N "Enable if you want internationalization support" true)
 
-option(BUILD_COLOUR_NAME_MAP "Include mappings of colour name -> RGB (e.g. red -> ff0000)" true)
-
 if(BUILD_I18N)
   set(LOCALE_DIR "${CMAKE_INSTALL_PREFIX}/share/locale"
     CACHE STRING "Directory containing the locales")
 endif(BUILD_I18N)
+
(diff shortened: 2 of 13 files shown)


brndnmtthws (approved):

Caellian:
TODO:
- [ ] Does it make sense to set zeroed struts on openbox for docks so we get consistent behavior?
- [ ] Openbox is very common, maybe try harder to detect it.
- [ ] What was initial intent behind having separate dock and panel modes, what are the differences supposed to even be?

Caellian:
Tested whether #1961 is fixed by setting `Xft.dpi` to double the value. Conky is correctly scaled to twice the size. So this issue can't be recreated anymore and I'll close it as well.

PR #2193, opened 2025-05-03
Add error for missing wayland-scanner

SFrijters:
# Checklist
- [ ] I have linked to any relevant GitHub issues, if applicable

## Description

I have fixed cross-compilation for conky in nixpkgs (to be reviewed/merged), but in the process I encountered a situation where wayland was found, but wayland-scanner was not. That allowed the configuration process to continue with just a warning, but then the build process failed when the wayland-scanner executable was not available. In my opinion it would be nicer to fail earlier, with a clear error message.

Tested locally as a patch applied to the nixpkgs derivation.


Caellian (approved):
LGTM

Issue #2211, opened 2025-06-25
[Bug]: postmarketOS identified as Alpine


JetpackJackson:
### What happened?

In the `${distribution}` variable, conky outputs "Alpine" instead of "postmarketOS".

contents of `/etc/os-release`:
```
PRETTY_NAME="postmarketOS v25.06"
NAME="postmarketOS"
VERSION_ID="v25.06"
VERSION="v25.06"
ID="postmarketos"
ID_LIKE="alpine"
HOME_URL="https://www.postmarketos.org/"
SUPPORT_URL="https://gitlab.postmarketos.org/postmarketOS"
BUG_REPORT_URL="https://gitlab.postmarketos.org/postmarketOS/pmaports/issues"
LOGO="postmarketos-logo"
ANSI_COLOR="0;32"
```

### Version

1.22.1

### Which OS/distro are you seeing the problem on?

Linux (other)

### Conky config

```Lua

```

### Stack trace

```GDB

```

### Relevant log output

```Shell

```

Caellian:
`print_distribution` function from `/data/os/linux.cc` reads this from `/proc/version` file.

`/etc/os-release` is much clearer for this use case and should be checked before `/proc/version`.

JetpackJackson:
Sure thing, I can test it in about 20-30 minutes or so.

JetpackJackson:
I'm having some issues getting cross-compilation to work (most likely PEBKAC on my end) but the code looks alright to me. I have to log off for today so I'm sorry I wasn't able to get the patch tested in time.

Caellian:
Merged into `main` so this issue is closed. In case of additional issues along these lines let me know and I'll reopen the issue.

PR #2221, opened 2025-07-08
Fix detection of postmaretOs in /etc/os-release parsing

Zawisza-Czarny:
# Checklist

## Description

This fixes the reported issue #2211 where the distro name was incorrectly displayed as AlpineOs) instead of postmarketOS
Output logs and console output confirm the correct display of the distro name.
image

### DIFF
--- a/src/data/os/linux.cc
+++ b/src/data/os/linux.cc
@@ -2793,10 +2793,26 @@ void print_distribution(struct text_object *obj, char *p,
   char *buf;
   struct stat sb;
 
+if ((buf = readfile("/etc/os-release", &bytes_read, 1))) {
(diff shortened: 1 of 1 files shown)


Caellian:
Thank you for the contribution!

I didn't like how you matched for `ID="postmarketos"` to set the variable so I switched it to reading the `NAME` instead. The goal of the function is to display the name of the distro, so there's no harm in returning that value directly instead of mapping names from IDs. While at it I also replaced C-style code that was already present in the function with newer C++ code because it's a bit more readable this way.

I was considering leaving a review at first, but I feel it's quicker for you to compare old and new code in the diff than it would've been for me to leave 4 comments with tweak requests.

Issue #2142, opened 2025-01-13
[IDEA] Make conky great again: add 100 conky themes


rubyFeedback:
First, this can be closed. It is an idea, a suggestion. I understand that some ideas or suggestions take time to implement. Nonetheless it would be nice if the main developers of conky could read this.

The title is just "catchy". A more realistic title would be:

"Consider adding ten curated conky themes by default to conky."

That is more realistic.

Now, we can grab conky themes from the world wide web, which is fine, but I would like to have
a "top ten" conky theme already distributed by conky, to then pick from.

This could also have commandline aliases such as:

    conky --theme1
    conky --theme2
    conky --theme3

and so forth.

Take a look at those four conky themes:

https://ubunlog.com/wp-content/uploads/2017/05/conky-harmatten.jpg.webp

I think it would be nice to have them in conky by default as-is.

Or, similar variants. Suggesting ten would be more realistic, but I propose 100,
just to satisfy every people's preference. ;)

These themes should be curated. Perhaps people, conky devs, can vote on which
ones to include if there are more than ten or hundred.

I would like to have that so I can just select through them via the commandline and
then pick one of those. Also an option such as "--copy-theme-1" should then be
added to dump the config into the current directory and allow users to modify it.
That way they can base their modifications on that modern conky theme.

I actually have an old conky but I haven't changed it in +10 years. I think I would
prefer to use someone else's conky but from a curated base; but also have a look
at different variants from within conky itself rather than external conky resources.

Anyway that is my suggestion. Please feel free to do with that suggestion whatever
you want to.

Caellian:
I like this idea and it would make it simpler for new users to start using conky. From what I can gather, a somewhat common experience is that people want to add conky, try using someone's dotfiles, and then sometimes fail to get them to work correctly for their system and stop using it, or find the process to tedious to repeat after a reinstall.

Problem with presets is that they have to be maintained. But this can also be used to our advantage by turning these presets into tests - people (myself included) often modify things without testing all the edge cases and combinations, and that can cause regressions in UI layouting which breaks ~stable~ existing config files. Turning these into tests would shift the maintenance burdain to feature implementers.

However, testing these raises a separate problem - there's no mocking support for most of conky features. I'm slowly working on this, but it requires some really complicated changes to the existing source code (i.e. Lua config handling).

[...]

PR #2278, opened 2025-11-16
feat: add theme presets

Cetttok:
created new module (themepresetmanager) and integrate it to conky in main.c.
implemented: 
- user can specify theme preset by alias in conifg option eg conky -c lean
- theme presets stored in git repo with simple struct (themealias.theme - folder name and everywords.conf - main config name)
- operations with git realise with libgit2 and I dont now there correct add includes ( in CMakeLists.txt files) 
- i dont know where i must to store my h and cpp files
- this code testes with /var/lib/conky/themes repo path and exapmle repo https://github.com/Cetttok/testRepoForConkyThemes
- this code cant clone repo with submodules and i dont know is it necessary
please view and say me what i must correct
i ready to perform yours requirements

Related to #2142

sorry for my English


CMakeLists.txt
```diff
@@ -53,13 +53,16 @@ set(conky_sources ${CMAKE_BINARY_DIR}/config.h ${CMAKE_BINARY_DIR}/build.h)
 add_subdirectory(3rdparty)
 set(conky_libs ${conky_libs} toluapp_lib_static)
 set(conky_libs ${conky_libs} Vc)
+set(conky_libs ${conky_libs} PkgConfig::LIBGIT2)
+
```
Caellian: Lua and Vc are added here because they're in thirdparty directory, but all the dependencies pulled from build system are added to `conky_libs` from `cmake/ConkyPlatformChecks.cmake` file. Search for `BUILD_IMLIB2` in CMake files for an example how to add libgit dependency.

CMakeLists.txt
```diff
@@ -74,6 +77,7 @@ if(BUILD_TESTING)
   include(Catch)
 endif()
 
+
```
Caellian: Go through the diff, either here on github in "Files changed" tab or using `git diff` and remove these unrelated changes. This space for instace, blank comments later, etc. make the PR harder to review than it needs to be.

src/conky.cc
```diff
@@ -160,6 +160,7 @@
 
 #ifdef BUILD_BUILTIN_CONFIG
 #include "defconfig.h"
+#include "themespresetmanager.h"
```
Caellian: Once you added a guard like `BUILD_IMLIB2`, the include needs to be guarded with `ifdef`s, as well as any code that's using that.

src/main.cc
```diff
@@ -300,7 +301,8 @@ inline void reset_optind() {
   optind = 0;
 #endif
 }
-
+const char * THEME_PRESETS_REPO_CLONING_URL = "https://github.com/Cetttok/testRepoForConkyThemes"; // it is example need to create norm repo
+const char * THEME_PRESETS_REPO_PATH = "/var/lib/conky/themes";
```
Caellian: Why are these in main.cc? You're declaring them here, just to pass them to the function in another file. Why not declare them above the function so they're easy to find and modify?

main.cc is already too large (much like conky.cc). Avoid placing more stuff in these two files.

src/main.cc
```diff
@@ -361,6 +367,10 @@ int main(int argc, char **argv) {
       case 'h':
         print_help(argv[0]);
         return 0;
+      // case 'S':
+      //   std::cout << "optarg " <<  optarg << std::endl;
+      //   std::cout << presets.getThemePath(std::string(optarg)) << std::endl;
+      //   break;
```
Caellian: Before merge, these comments need to be removed if the code isn't used. Very rarely it makes sense to leave code commented out - there's commit history for viewing old code if you need it. But let's keep that for final cleanup.

src/main.cc
```diff
@@ -389,10 +399,24 @@ int main(int argc, char **argv) {
...
+  ///
+  ///
+  ///
+  ///
+  /// ////
 
   try {
     set_current_config();
-
+    tryToReplaceAliasToPresetPath(THEME_PRESETS_REPO_PATH, THEME_PRESETS_REPO_CLONING_URL );
```
Caellian: `set_current_config` resolves paths to configs and _then_ you replace theme names with paths?

Shouldn't `tryToReplaceAliasToPresetPath` logic go into `set_current_config`, and before we do path resolution?

Ideally, there should be some easily recognizable disambiguation between a system path and a theme/preset name. Something like `theme:herpy_derpy_der`, so that you can easily and cheaply assume it's a theme, and not a directory relative to cwd (without asking the FS).

src/themespresetmanager.cpp
```diff
@@ -0,0 +1,288 @@
...
+ */
+#include "themespresetmanager.h"
+#include <filesystem>
+#include <iostream>
+#include <cstdlib>
+#include <sys/stat.h>
+#include <git2.h>
+#include <git2/common.h>
+const char * THEME_END_NAME = ".theme"; // can be changed at another one
+const char * THEME_CONFIG_END = ".conf"; // can be changed at another one
```
Caellian: No need for these comments. It's FOSS, of course it can be changed :)

src/themespresetmanager.cpp
```diff
@@ -0,0 +1,288 @@
...
+  if(git_remote_lookup(&remote, _repo, "origin")!=0){
+    std::cout << "SystemGitRepoSource::error while git remote lookup (origin): "<<git_error_last()->message << std::endl;
+    return false;
+  }
+  if (git_remote_fetch(remote,NULL,NULL,"fetch")!=0){
+    std::cout << "SystemGitRepoSource::error while git remote fetch (origin): "<<git_error_last()->message << std::endl;
+    return false;
+  }
+  git_reference * mainBranch = nullptr;
+  if (git_reference_lookup(&mainBranch, _repo, "refs/remotes/origin/main")!=0){
```
Caellian: Hardcoding "main" ref for repositories feels bad. It should be specified in the same place the repository URI is.

src/themespresetmanager.h
```diff
@@ -0,0 +1,95 @@
...
+  SystemGitRepoSource (std::string pathToGitAlreadyClonedRepo);
+  ~SystemGitRepoSource();
+  virtual bool loadThemesDb(); // if loded one or more theme returns true
+ protected:
+  bool pullChanges();
+  bool openRepo(std::string pathToRepo);
+  bool isGettedPathGitRepo(std::string pathToRepo);
+  git_repository * _repo = NULL;
+  // /std::string _pathToRepo = "";
+  std::string _repoUrl = "";
```
Caellian: Don't prefix with `_` in C++. All `_` prefixed names are reserved for compiler and standard library implementations. See this [SO answer](https://stackoverflow.com/questions/228783/what-are-the-rules-about-using-an-underscore-in-a-c-identifier).

Caellian (changes requested):
Here's a preliminary review.

I'm not 100% sure about adding this feature and would like @brndnmtthws input before accepting this PR. It's a neat idea and I see how people would find it useful and convenient, but it also makes it very easy to distribute malicious configs because people won't check the presets/themes before updating them, just like they don't for AUR lol.

We could author an official repo with reviewed configs, but that's also a lot of work or trust to put into someone. So this would mostly be useful to Linux distribution authors.

Cetttok:
Halo guys! I make most of all that  @Caellian told me.
That changed in last commits:
- repo not pulling of every path replacing request. only pulling one folder or subrepo of requested theme
- fixed memory leaks for libgit2 
- build options (bool BUILD_PRESETS string PRESETS_REPO_URL PRESETS_REPO_PATH)
- code used libgit masked in it option
- repo configuration file in xml (repo.conf) example in theme-presets-manager.h after advanced_repo_source declaration
- code in conky code style
- clang format on my module
- add loging about long operations (eg pulling ont theme) and theme presets repo that used
- adding support custom branches in non main repo (in submodules etc)
- git fetch git clone and another standart (and not standart) methods realised
thats not all.

Cetttok:
@Caellian i know that it is opensource (free of all)  but will be nice (for me) if you maybe say my Judgment for this

Caellian:
Thanks again for working on this. I want to give a clearer status update here, because I have intentionally not closed this PR.

I do think this is a useful direction for Conky. The best version of Conky definitely has a better story around discoverable, installable, and shareable configurations/themes. This PR is one of the few concrete attempts in that direction, so I do not want to discard it just because it is not immediately mergeable. It's also good work on your part: showcases good ability to navigate conky's large code-base, conky's CMake build system, and Linux environment well.

The reason I have not moved forward with it yet is that this is effectively a new subsystem, not a small feature. It introduces questions around dependency size, repository format, update/fetch behavior, default hosting, decentralization, trust, maintenance, testing, and security. Those decisions are _difficult to change later_ once users and distributions start relying on them.

There are a few things I would still want to rethink before merging something like this:

- I would like to minimize hard dependencies. I am not yet convinced Conky itself needs to depend on libgit2 for this. Static hosting, plain archives, or a simpler repository format may be enough.
- I would prefer the design to be decentralized, while still allowing us to point users toward a trustworthy default source (e.g. Ubuntu package repos).
  - I do not want users to be locked into one official repository in a way that requires rebuilding or patching the binary to use another source.
- I would like to avoid making XML a hard requirement.
- I think this should probably be tied to better visual/regression testing infrastructure, so that official/example configs can be tested in CI instead of silently breaking over time.

The largest concern, though, is security. Conky configs can execute Lua and shell commands, so a theme/config distribution mechanism can very easily become a malware distribution mechanism if we get the trust model wrong. This is especially risky for newer Linux users who may not understand that downloading and running a Conky theme is not equivalent to installing a harmless wallpaper.

Existing tools such as Conky Manager 2 handle this partly by having a much more controlled contribution path. That has downsides I don't want to see in a feature like this, but it also limits the risk of arbitrary third-party content being presented as safe or official. Conky has a very small maintainer/reviewer base (~2 people), so we need to be careful before adding an official-looking channel for downloading executable configurations. This adds a requirement to the infrastructure around it to amortize, at times, a slow review cycle.

So the current status is: I am keeping this open because I do not consider the idea rejected. An open PR is still there and visible. But I do not think now is the right time to merge it as-is. Before that, I think we need a clearer design for the repository model, trust boundaries, default source, dependency strategy, and testing story. I don't expect you to solve all of these on your own, and I'm letting the concept evolve in my head as well.

I may come back to this later after some of the surrounding infrastructure is in better shape. Even if this eventually needs to be reworked heavily, I still think the PR is valuable because it exposes the right set of problems and gives us something concrete to reason about. Once that happens, I will update your commit and keep you as author, ping you so you can chime in on it, and add any extras around it that might be needed.

tl;dr: this is valuable and aligned with where the project should go, but merging it now would commit us to infrastructure and policy we have not designed yet.

Note: I used an LLM to summarize my Ted talk long response.

PR #2285, opened 2025-11-24
feat(text): add sysfs monitoring bar variables

mfluhr:
## Description

I used "hwmon" for years now to monitor my CPU core temps, but needed to fallback to an execbar to have those values shown as bar nearby. Since doing so for all cores is quite heavy, I added some sysfs bar variables (hwmonbar, i2cbar and platformbar) that basically fill this gap. Doing so, spared me some CPU usage.

The code should not affect existing conky configurations, since it is new variables addition.

I tested this using the "hwmonbar" variable to monitor my AMD CPU core package temps, replacing this
    `${execbar sensors -u k10temp-pci-00c3 | grep temp1_input | awk '{print $2}' | cut -d '.' -f1}`
with this
    `${hwmonbar k10temp temp2 1 1.0 0.0}`

One important note: Due to the bar scaling when calling setup_bar(), the factor and offset should be adjusted properly so that the value is always inside a 0-100 range.

### DIFF
--- a/doc/variables.yaml
+++ b/doc/variables.yaml
@@ -823,6 +823,21 @@ values:
       - (factor offset)
     other:
       filename: null
+  - name: hwmonbar
+    desc: |-
+      Same as hwmon, except if the first value returned is a value
+      between 0-100, it will use that number to draw a horizontal bar (use the
+      factor and offset to transform the raw value). The height and width
+      parameters are optional, and default to the default_bar_height and
+      default_bar_width config settings, respectively.
+    args:
+      - (height),(width)
+      - (dev)
+      - type
+      - n
+      - (factor offset)
+    other:
+      filename: null
   - name: i2c
     desc: |-
       I2C sensor from sysfs (Linux 2.6). Parameter dev may be omitted if you have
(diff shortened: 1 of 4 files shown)


brndnmtthws (changes requested):
Thanks for the addition. I think there are a couple issues to fix before merge:

1) platformbar is wired to parse_i2c_bar, so it uses the i2c sysfs path and will not find platform devices. It should call parse_platform_bar. (src/core.cc)
2) sysfs_sensor_barval clamps negative values to 100.0 instead of 0.0, so a negative value shows a full bar. (src/data/os/linux.cc)

Please address these and I’m happy to re-review.

mfluhr:
Whoops.... That was 2 very dumb mistakes!
I just fixed that and amended the initial commit.

doc/variables.yaml
```diff
@@ -823,6 +823,20 @@ values:
       - (factor offset)
     other:
       filename: null
+  - name: hwmonbar
+    desc: |-
+      Same as hwmon, but displays the sensor reading as a horizontal bar. Use
+      factor and offset to scale the raw sysfs value into the 0-100 range
+      (e.g. millidegree temps need factor 0.001). Height and width default to
+      default_bar_height and default_bar_width respectively.
```
Caellian: I updated your docs because they were copy-pasted from somewhere and weren't describing the variables correctly.

src/data/os/linux.cc
```diff
@@ -1496,6 +1501,22 @@ void print_sysfs_sensor(struct text_object *obj, char *p,
...
+  if (!sf || sf->fd < 0) return 0.0;
+
+  r = get_sysfs_info(&sf->fd, sf->arg, sf->devtype, sf->type);
+
+  r = r * sf->factor + sf->offset;
+
+  if (r < 0.0 || r > 100.0) {
+    LOG_WARNING("sysfs bar value {} out of [0, 100]; adjust factor/offset", r);
+  }
+  return std::clamp(r, 0.0, 100.0);
```
Caellian, line 1535: Swapped a if-elseif with a `std::clamp` here. Switched to new logging (rebased).

Caellian (approved):
I rebased and fixed the docs. The PR is fine to merge now.

brndnmtthws (approved):

PR #2314, opened 2026-01-22
Added an optional argument to the 'memperc' variable

mfluhr:
## Description

Added an optional argument to the 'memperc' variable to specify which kind of memory to use.

As an Android OS developer, I am rebuilding the AOSP daily, and the build process is heavily multi-threaded, consuming almost all my RAM. Since I want to avoid the oomd killer kicking in, I am closely monitoring RAM usage, not only using `mem` and `memperc`, but also `memwithbuffers`.

Since the 'memperc' variable only uses the 'mem' value, I tought about adding an optional paramter to tailor which memory should be used when computing the percentage.

The code should not affect existing conky configurations, since it is a new optional argument addition.
It has been tested locally using the following configuration
```
[...]
${font DejaVu Sans:bold:size=14}MEMORY ${hr 2}${font}${voffset 8}
${font DejaVu Sans:bold:size=10}RAM${font}${alignc}$mem / $memmax${alignr}${font DejaVu Sans:bold:size=10}$memperc%${font}${voffset 2}
${membar 16}${voffset 4}
${font DejaVu Sans:bold:size=10}BUFFERED${font}${alignc}${memwithbuffers} / ${memmax}${alignr}${font DejaVu Sans:bold:size=10}${memperc -b}%${font}${voffset 2}
${memwithbuffersbar 16}${voffset 4}
${font DejaVu Sans:bold:size=10}SWAP${font}${alignc}${swap} / ${swapmax}${alignr}${font DejaVu Sans:bold:size=10}${swapperc}%${font}${voffset 2}
${swapbar 16}
[...]
```

### DIFF
--- a/doc/variables.yaml
+++ b/doc/variables.yaml
@@ -1368,6 +1368,10 @@ values:
       - (-x)
       - (-y)
       - (-m value)
+  - name: memwithbuffersperc
+    desc: |-
+      Percentage of memory in use (including memory used by system buffers
+      and caches).
   - name: mixer
     desc: |-
       Prints the mixer value as reported by the OS. On Linux, this
--- a/doc/variables.yaml
+++ b/doc/variables.yaml
@@ -1338,7 +1338,24 @@ values:
   - name: memmax
     desc: Total amount of memory.
   - name: memperc
-    desc: Percentage of memory in use.
+    desc: |-
+      Percentage of memory in use.
+
+      Can be fine-tuned using optional arguments:
+
+      * **-b**: includes system buffers and caches.
+      * **-e**: includes memory that is very easily freed.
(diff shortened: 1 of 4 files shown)


ETFM6T (approved):

mfluhr:
Ok, I can of course adjust the code in this direction...
...but the $memwithbuffers is already existing (hence my initial design....)

So, if I introduce now a '-b' argument, I think we should also remove the $memwithbuffers, and this would trigger errors in existing configurations. Are you OK with such change?

mfluhr:
Hi @Caellian 

I have refactored this PR, adding, as suggested, an optional argument to the 'memperc' variable.
Could you please review it and let me know what you think?

PR #2315, opened 2026-01-22
Added a new width argument to 'hr' and 'stippled_hr' variables

mfluhr:
## Description

When using an `hr` variable there is currently no real way to restrict the width of it, in order to draw something between the end of the line, and the right side of the window. I have therefore added a new optional parameter for both `hr` and `stippled_hr` variables to fix that.

There should be no impact on existing behavior, since we are adding a new (optional) parameter.

However, please note that the logic inside the `scan_stippled_hr()` function favors the height/thickness parameter instead of the space parameter when only one argument is given. In order to preserve this logic, I kept the space + height combo when 2 arguments are passed, so there is now way to have a `stippled_hr height width` combination.

I have tested both `hr` and `stippled_hr` in all possible combinations, checking that the line rendered on screen fits what is expected.

|Variable|Width|Height|
|:-|:-:|:-:|
|`${hr}`|_100%_| _1px_ |
|`${hr 10}` |_100%_| _10px_ |
|`${hr 10 150}` |150px| 10px |

<sup>_italics_ - default</sup>

### DIFF
--- a/src/conky.cc
+++ b/src/conky.cc
@@ -1181,7 +1181,11 @@ int draw_each_line_inner(char *s, int special_index, int last_special_applied) {
             int h = current->height;
             int mid = font_ascent() / 2;
 
-            w = text_start.x() + text_size.x() - cur_x;
+            /* if no width was specified, current->width is set to 0 */
+            w = current->width;
+            if (w <= 0) {
+              w = text_start.x() + text_size.x() - cur_x;
+            }
 
             if (display_output()) {
               display_output()->set_line_style(h, true);
@@ -1200,7 +1204,12 @@ int draw_each_line_inner(char *s, int special_index, int last_special_applied) {
             int mid = font_ascent() / 2;
             char ss[2] = {tmp_s, tmp_s};
 
-            w = text_start.x() + text_size.x() - cur_x - 1;
+            /* if no width was specified, current->width is set to 0 */
+            w = current->width;
+            if (w <= 0) {
+              w = text_start.x() + text_size.x() - cur_x - 1;
+            }
+
             if (display_output()) {
               display_output()->set_line_style(h, false);
               display_output()->set_dashes(ss);
--- a/src/conky.cc
+++ b/src/conky.cc
@@ -1187,8 +1187,13 @@ int draw_each_line_inner(char *s, int special_index, int last_special_applied) {
           if (display_output() && display_output()->graphical()) {
             int h = current->height;
             int mid = font_ascent() / 2;
+            int max_width = text_start.x() + text_size.x() - cur_x;
 
-            w = text_start.x() + text_size.x() - cur_x;
+            /* if no width was specified, current->width is set to 0 */
+            w = current->width;
+            if (w <= 0 || w > max_width) {
+              w = max_width;
+            }
 
             if (display_output()) {
               display_output()->set_line_style(h, true);
@@ -1205,9 +1210,15 @@ int draw_each_line_inner(char *s, int special_index, int last_special_applied) {
             int h = current->height;
             char tmp_s = current->arg;
             int mid = font_ascent() / 2;
+            int max_width = text_start.x() + text_size.x() - cur_x - 1;
             char ss[2] = {tmp_s, tmp_s};
 
-            w = text_start.x() + text_size.x() - cur_x - 1;
+            /* if no width was specified, current->width is set to 0 */
+            w = current->width;
+            if (w <= 0 || w > max_width) {
+              w = max_width;
+            }
+
             if (display_output()) {
               display_output()->set_line_style(h, false);
               display_output()->set_dashes(ss);
--- a/src/content/specials.cc
+++ b/src/content/specials.cc
@@ -116,8 +116,12 @@ struct graph {
   int minheight;    /* Clamp values below this threshold to this threshold */
 };
 
+struct classic_hr {
+  int width, height;
+};
+
 struct stippled_hr {
-  int height, arg;
+  int width, height, space;
 };
 
 struct tab {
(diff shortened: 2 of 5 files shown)


src/content/specials.cc
```diff
@@ -713,12 +717,36 @@ void new_graph(struct text_object *obj, char *buf, int buf_max_size,
...
+  hr->height = 1;
+
+  if (arg != nullptr) {
+    if (sscanf(arg, "%d %d", &hr->height, &hr->width) != 2) {
+      sscanf(arg, "%d", &hr->height);
+    }
+  }
+
+  if (hr->height <= 0) { hr->height = 1; }
+  obj->special_data = hr;
```
Caellian, line 736: Missing `obj->callbacks.free`.

src/content/specials.cc
```diff
@@ -713,12 +717,36 @@ void new_graph(struct text_object *obj, char *buf, int buf_max_size,
   if (out_to_stdout.get(*state)) { new_graph_in_shell(s, buf, buf_max_size); }
 }
 
+void scan_hr(struct text_object *obj, const char *arg) {
+  struct classic_hr *hr;
+
+  hr = static_cast<struct classic_hr *>(malloc(sizeof(struct classic_hr)));
+  memset(hr, 0, sizeof(struct classic_hr));
```
Caellian: C-style memory management. Use `new` or (better yet) a RAII struct.

src/content/specials.cc
```diff
@@ -713,12 +717,36 @@ void new_graph(struct text_object *obj, char *buf, int buf_max_size,
...
+  hr->width  = 0;
+  hr->height = 1;
+
+  if (arg != nullptr) {
+    if (sscanf(arg, "%d %d", &hr->height, &hr->width) != 2) {
+      sscanf(arg, "%d", &hr->height);
+    }
+  }
+
+  if (hr->height <= 0) { hr->height = 1; }
```
Caellian, line 735: Use `hr->height = std::max(1, hr->height)`.

src/content/specials.cc
```diff
@@ -727,15 +755,19 @@ void scan_stippled_hr(struct text_object *obj, const char *arg) {
...
-      sscanf(arg, "%d", &sh->height);
+    if (sscanf(arg, "%d %d %d", &sh->space, &sh->height, &sh->width) != 3) {
+      if (sscanf(arg, "%d %d", &sh->space, &sh->height) != 2) {
+        sscanf(arg, "%d", &sh->height);
+      }
     }
   }
-  if (sh->arg <= 0) { sh->arg = 1; }
+  if (sh->space <= 0) { sh->space = 1; }
+  if (sh->height <= 0) { sh->height = 1; }
```
Caellian, line 770: Also `std::max(1, ...)`

src/conky.cc
```diff
@@ -1181,7 +1181,11 @@ int draw_each_line_inner(char *s, int special_index, int last_special_applied) {
             int h = current->height;
             int mid = font_ascent() / 2;
 
-            w = text_start.x() + text_size.x() - cur_x;
+            /* if no width was specified, current->width is set to 0 */
+            w = current->width;
+            if (w <= 0) {
+              w = text_start.x() + text_size.x() - cur_x;
+            }
```
Caellian, line 1188: Missing upper bound.

src/conky.cc
```diff
@@ -1200,7 +1204,12 @@ int draw_each_line_inner(char *s, int special_index, int last_special_applied) {
             int mid = font_ascent() / 2;
             char ss[2] = {tmp_s, tmp_s};
 
-            w = text_start.x() + text_size.x() - cur_x - 1;
+            /* if no width was specified, current->width is set to 0 */
+            w = current->width;
+            if (w <= 0) {
+              w = text_start.x() + text_size.x() - cur_x - 1;
+            }
```
Caellian, line 1211: Also missing upper bound.

src/content/specials.cc
```diff
@@ -116,8 +116,12 @@ struct graph {
   int minheight;    /* Clamp values below this threshold to this threshold */
 };
 
+struct classic_hr {
+  int width, height;
+};
+
```
Caellian, line 122: Partial variant of `stippled_hr`. One struct would suffice to describe both. Waste is negligible.

src/content/specials.cc
Caellian, line 724: C-style constructor and pointless `memset`. Use `new` and add defaults to the struct.

Caellian (changes requested):

src/content/specials.cc
Caellian: The old code causes a leak too. Should probably be fixed while you're at it.

src/content/specials.cc
mfluhr: I agree, but the whole file uses this pattern.
Should I really change this to use 'new' with defaults? just for hr ? for all the file?

src/content/specials.cc
mfluhr: Done.

src/content/specials.cc
mfluhr: Done

src/content/specials.cc
mfluhr: Done.

src/content/specials.cc
mfluhr: Done

src/conky.cc
mfluhr: Done

src/conky.cc
mfluhr: Done

mfluhr:
Hi @Caellian,

I adjusted my commit to reflect all the comments that you added, and also re-based it against the actual state of the main branch (Sorry for the delay, but I was quite busy at work the last months, and had no time to concentrate on that sooner...).

There is only one comment where I am asking myself what to do, and this is the one where you suggested to move from the malloc/memset C-style to a more C++ compliant task. Why I took the C-style approach in the first place, is that all allocations are done this way inside the specials.cc file, and switching one (or 2) in this case to a more modern approach, leaving the other ones untouched would be quite inconsistent. I would see that more for a separate PR, focusing on modernizing the code.

brndnmtthws (approved):

mfluhr:
Hi @brndnmtthws 
I just checked your second commit, and, among cosmetic changes, I see that it actually deleted the "free" callbacks....
....that @Caellian asked me to add. Is that actually expected and will not cause any memory leak?

brndnmtthws:
> Hi @brndnmtthws I just checked your second commit, and, among cosmetic changes, I see that it actually deleted the "free" callbacks.... ....that @Caellian asked me to add. Is that actually expected and will not cause any memory leak?

Uh that's my bad. Let me double check that.

brndnmtthws:
I double-checked the teardown path, and removing these callbacks is intentional.

`scan_hr()` / `scan_stippled_hr()` store a plain `malloc` allocation in `obj->special_data`, and `free_text_objects()` already calls `free_and_zero(obj->special_data)` after any object-specific free callback runs. So without a custom callback, this still gets freed once.

The callbacks I removed were actually unsafe because they freed a local copy of `obj->special_data` but did not null `obj->special_data` itself. After the callback returned, the generic cleanup path would still call `free_and_zero(obj->special_data)`, causing a double-free.

So there should be no leak with the current code. If this allocation is later changed from `malloc` to `new`, then the cleanup path should be revisited because the generic cleanup uses `free()`.

PR #2320, opened 2026-02-09
Refactor: Replace magic numbers with named constant in linux.cc

debanjan-bhuinya:
### Description
I noticed a `FIXME` comment in `src/data/os/linux.cc` regarding an arbitrary buffer size.
I have replaced the hardcoded magic numbers (`256`, `255`) with a named constant `BUF_SIZE` to improve readability and maintainability.

### Changes
- Defined `const int BUF_SIZE = 256;`
- Updated `char buf[]` and `fgets()` to use `BUF_SIZE`.

### Verification
- [x] Ran `make -j$(nproc)` locally on Ubuntu.
- [x] Confirmed the build completes successfully (100%).

### DIFF
--- a/src/data/os/linux.cc
+++ b/src/data/os/linux.cc
@@ -504,8 +504,9 @@ void update_net_interfaces(FILE *net_dev_fp, bool is_first_update,
 
     /* quit only after all non-header lines from /proc/net/dev parsed */
     // FIXME: arbitrary size chosen to keep code simple.
-    char buf[256];
-    if (fgets(buf, 255, net_dev_fp) == nullptr) { break; }
+    const int BUF_SIZE = 256;
+    char buf[BUF_SIZE];
+    if (fgets(buf, BUF_SIZE - 1, net_dev_fp) == nullptr) { break; }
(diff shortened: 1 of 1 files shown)


Caellian:
FIXME is saying that 256 is a fixed buffer size that's chosen arbitrarily, and that ideally the code should dynamically allocate precisely the size of the buffer needed to store the contents of the file.

But it's a non issue - the buffer is large enough in all practical use cases, and dynamically allocating it correctly would be slower in practice while saving 100 or so bytes of memory. Fixing this minor inefficiency can only introduce bugs into otherwise working code, and this part of the code specifically is waiting for a rewrite so you'd be potentially causing merge conflicts without improving anything.

In general, PRs that are ~3 lines long only make sense if they're fixing a bug. If it's a tweak that improves code readability, you're just wasting the reviewer's time and contributing to global warming by spinning up CI.

I suggest you take a look at one of ~90 bugs in the issues, find something that resonates with you or affects your system, and fix that. That'd be more educational, better for your CV, and would help all users as well.

PR #2329, opened 2026-03-10
fix: replace magic buffer size 256 with named constant LINUX_BUF_LEN

riteshdhurwey:
## Summary
Replace hardcoded magic number `256` with a named constant `LINUX_BUF_LEN = 512`
for buffer sizes in linux.cc.

## Changes
- Added `static constexpr int LINUX_BUF_LEN = 512`
- Replaced all `char buf[256]` with `char buf[LINUX_BUF_LEN]` in linux.cc
- Resolves two FIXME comments about arbitrary buffer size

## Testing
Built and tested on Arch Linux. Zero compiler warnings.

### DIFF
--- a/src/data/os/linux.cc
+++ b/src/data/os/linux.cc
@@ -48,6 +48,7 @@
 // #include <assert.h>
 #include <time.h>
 #include <unordered_map>
+static constexpr int LINUX_BUF_LEN = 512;
 #include "../../lua/setting.hh"
 #include "../top.h"
 
@@ -180,7 +181,7 @@ int update_meminfo(void) {
   static int reported = 0;
 
   /* unsigned int a; */
-  char buf[256];
+  char buf[LINUX_BUF_LEN];
 
   /* With multi-threading, calculations that require
    * multiple steps to reach a final result can cause havok
@@ -504,7 +505,7 @@ void update_net_interfaces(FILE *net_dev_fp, bool is_first_update,
 
     /* quit only after all non-header lines from /proc/net/dev parsed */
     // FIXME: arbitrary size chosen to keep code simple.
-    char buf[256];
+    char buf[LINUX_BUF_LEN];
     if (fgets(buf, 255, net_dev_fp) == nullptr) { break; }
     p = buf;
     /* change char * p to first non-space character, which is the beginning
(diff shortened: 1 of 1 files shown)


Caellian:
It'd make sense if this PR actually modernized something in the code (e.g. turned C-style arrays into `std::array`).

I see you're a student, so I'll leave this PR open for a bit and let you repurpose it into a meaningful refactor. 

This file:
- Uses `while (!feof(fp))` a lot, this is an anti-pattern. Use an LLM to explain why this is bad. Fix it in one place on your own. Use an LLM to reproduce the pattern on all lines.
- A lot of the code still uses malloc/free instead of new/delete or RAII.
  - There's C-style casts with malloc that can be replaced with `std::vector`.
  - malloc constructed struct pointers that can use C++ `new` instead.
- Raw `FILE*` handles that can cause resource leaks if someone adds a `return` or `break`. Should use RAII.
- Too much global mutable state.
- Unchecked return values from system calls (glob for `ioctl`, `socket`, `sysinfo`)
- `goto` is used. can be replaced with `while(false) { ... break; ... }` to help the compiler reason about control flow better.
- `strncpy` used where `snprintf` would be safer for fixed buffers.
- OpenMP pragmas used on tiny loops - spawning OpenMP threads is slower than doing the actual work.
- Replace `pthread_mutex` with `std::mutex`.

Those are just some trivial problems an LLM can find by looking at the file. I filtered out low-value ones. Pick any number of these and address them **exhaustively** (not just one number).

Fixing any one of those will serve you better long term - both knowledge, building up a CV, potentially improving SysMon, etc. than just replacing a single number.

These are all meaningful improvements that could be done, it's just that fixing them one-by-one would spam 20+ PRs we'd need to review and we're not accepting changes like these unless they come in bulk because it's more trouble than it's worth to review each one.

Caellian:
It's been 3 weeks since my proposal for improvement of the PR, so I'll close it.

I welcome you attempt a larger refactor (with/without LLM tooling) given how accessible changes like these are these days, but it has to actually sufficiently improve the code and make sense in larger context of the application or at least a single file. As is, this PR is too simple to merge.

Refactors aren't allowed to touch functionality unless it's fixing a bug which would've made the refactor a lot less clean (buffer size change is bad unless you can rationalize it).

PR #2330, opened 2026-03-10
fix: percent_print respects use_spacer when pad_percents is unset

### LINKED ISSUE #2078: [Bug]: "cpu" is not affected by "use_spacer = 'left'"
### What happened?

Hi.
I think when [`use_spacer = 'left'`](https://conky.cc/config_settings#use_spacer) is used `cpu` should have leading spaces:
```
  1%
 10%
100%
```
Like `upspeed` or `downspeed` or `diskio_read` ...
This issue is not for the specified version. I know it for quite some time but now I've decided to report it. So far I've used if statements to fix it.

Actual behavior:

### Version

conky 1.21.5-1

Arch Linux

### Conky config

```Lua
conky.config = {
	alignment = 'bottom_left',
	background = false,
	border_width = 0,
	double_buffer = true,
	draw_outline = false,
	font = 'DejaVuSansMono Nerd Font:size=10.5',
	minimum_height = 18,
	minimum_width = 500,
	out_to_console = false,
	out_to_x = true,
	own_window = true,
	own_window_type = 'desktop',
	update_interval = 1.0,
	use_xft = true,
	max_text_width = 0,
	short_units = true,
	if_up_strictness = 'address',
	use_spacer = 'left',
};

conky.text = [[
  ${cpu cpu1}% ${cpu cpu2}% |   ${time %d.%m.%Y} |   ${time %H:%M}
]];
```

riteshdhurwey:
## Problem
`use_spacer = 'left'` was ignored for `${cpu}` and other percentage 
values when `pad_percents` was not explicitly set.

## Root Cause
`percent_print()` passed `pad_percents` (default: 0) as width to 
`spaced_print()`. A width of 0 means no padding, so spacer had no effect.

## Fix
Default width to 3 when `pad_percents` is unset, since percentage 
values are always 1-3 digits (0-100).

## Testing
Tested on Arch Linux with `use_spacer = 'left'`:
Before: `CPU: 5%` `CPU: 14%` (inconsistent width)
After:  `CPU:   5%` `CPU:  14%` (consistent width)

Closes #2078

### DIFF
--- a/src/conky.cc
+++ b/src/conky.cc
@@ -507,7 +507,11 @@ int spaced_print(char *buf, int size, const char *format, int width, ...) {
  * - i.e., unsigned values between 0 and 100
  * - respect the value of pad_percents */
 int percent_print(char *buf, int size, unsigned value) {
-  return spaced_print(buf, size, "%u", pad_percents.get(*state), value);
+  int width = pad_percents.get(*state);
+  if(width == 0){
+    width = 3;
+  }
+  return spaced_print(buf, size, "%u", width, value);
 }
 
 /* converts from bytes to human readable format (K, M, G, T)


Caellian:
See my comment in #2329 on tiny changes.

Caellian:
Sorry, my logic in the comment on that issue was a bit off. We shouldn't clamp. Update the variable default to 3 and documentation for `pad_percents`.

We should allow trimmed printing in case someone uses conky from shell and wants to split the output on spaces (fields), but that's far less common than UI where you almost never want your UI to jitter because of percentages.

Caellian:
Ok, given that it's been 3 weeks, I'll close this. If you regain interest in this PR let me know and I'll reopen it.

PR #2336, opened 2026-04-04
fix: Fixed rendering of gradients in graphs

### LINKED ISSUE #2169: [Bug]: Graph functions (downspeedgraph, upspeedgraph) no gradient
### What happened?

There should be a gradient from color 1 to color 2 while graphing downspeedgraph/upspeedgraph.
With current version conky uses only second value and draws graph with solid color from start to end.
There is also issue of scaling for small values. It may be connected with this issue.

Problem with gradient missing seems to start with  conky 1.22.0
Problem with wrong scaling seems to start with 1.21.5

What it looks like:
Image
What it should look like:
Image

### Version

1.22.0-1 

Arch Linux

### Conky config

```Lua

```

### Stack trace

```GDB

```

### Relevant log output

```Shell

```

reboot:
The incrementation of the colour_idx variable was no longer working after it was moved into a function and only incremented the parameter instead.

Closes #2169

# Checklist

## Description

Gradient rendering in graphs did not work.

Ran the old and patched version and verified that gradients only work in the patched version.

### DIFF
--- a/src/conky.cc
+++ b/src/conky.cc
@@ -1000,7 +1000,7 @@ static inline void set_foreground_color(Colour c) {
 
 static inline void draw_graph_bars(special_node *current, std::unique_ptr<Colour[]>& tmpcolour, 
                             conky::vec2i& text_offset, int i, int &j, int w, 
-                            int colour_idx, int cur_x, int by, int h) {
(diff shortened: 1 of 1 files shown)


Caellian (approved):
Perfect diff. Thank you for figuring it out.

PR #2337, opened 2026-04-06
fix(x11): mouse event handling - scroll direction, hit testing, XInput2 performance; always use XI2

### LINKED ISSUE #1886: [Bug]: Conky freezes on mouse movement

### LINKED ISSUE #2066: [Documentation]: add an example config for mouse events
Currently we have an explanation on the wiki how mouse events work and how to get
them to work. This article can be found here:

https://github.com/brndnmtthws/conky/wiki/Mouse-Events

Now while it is great to have such an example, I don't want to copy/paste, fiddle around
and then end up with things not working. I know ruby fairly well, but not lua, and I don't
fully understand what I really should do.

Would it be possible to provide a small example for conky that is **intrinsically complete**,
that is, works out of the box, where ideally these key-events are showcased? This can be
super-simple, like a hello world, but just written in a way for people to:

a) just drop it and start it, as-is, without any modification

and

b) have it work for the basic stuff

I believe adding such a minimal example would be super-convenient for other users too;
and ideally the wiki article could be expanded to point where this example could be found,
perhaps under examples/ or config/ or some such.

I'd love to try this out, but after reading the wiki article, I still am not sure what to do. (I
should also point out that it has been quite a long while since I last touched any config
for conky, so this adds to my confusion; that's why I'd love a tabula rasa state in this
way. Then I can understand how it should work, and adjust it, e. g. also answering
things such as "how can I call a function or method in a specific file". Actually, can we
even call executables via such actions? Would be nice if the example file could also
show this, again, just a minimal example so we can learn the monkey-see-monkey-do
way. Thanks for reading!

(Also, it would be nice if we could have generic MACROS so that we could avoid having to
use lua, e. g. EXECUTE_FILE /usr/bin/funny_cats.rb - while that may look ugly, I would 
ideally like to avoid having to use lua altogether.)

### LINKED ISSUE #2288: [Bug]: Mouse scroll event regression
### What happened?

In Lua event handler function:
```
function conky_event_handler(event)
	print(event.type)
	return false
end
```
I only receive events: `mouse_move` `mouse_enter` and `mouse_leave`. But I don't receive events: `button_down`, `button_up` nor `button_scroll`.

The problem also occurs on version 1.22.2. This work fine (in the same environment and with the same config) after downgrade Conky to 1.18.3 (and removed `conky_` prefix from event handler function name).

### Version

1.22.1 (from Debian package conky-all 1.22.1-1)

Debian

### Conky config

```Lua
conky.config = {
    alignment = 'bottom_right',
    gap_x = 0,
    gap_y = 0,
    border_width = 0,
    own_window = true,
    own_window_class = 'Conky',
    own_window_type = 'normal',
    own_window_hints = 'undecorated,sticky,skip_taskbar,skip_pager',
    
    out_to_console = false,
    out_to_ncurses = false,
    out_to_stderr = false,
    out_to_wayland = false,
    out_to_x = true,
    
    background = false,
    cpu_avg_samples = 2,
    default_color = 'white',
    double_buffer = true,
    draw_borders = false,
    draw_graph_borders = true,
    draw_outline = false,
    draw_shades = false,
    extra_newline = false,
    minimum_height = 5,
    minimum_width = 5,
    net_avg_samples = 2,
    no_buffers = true,
    show_graph_range = false,
    show_graph_scale = false,
    stippled_borders = 0,
    update_interval = 0.5,
    uppercase = false,
    use_spacer = 'none',
    use_xft = true,
    lua_load = '.config/conky/conky.lua',
    lua_mouse_hook = 'event_handler',
}
```

### Stack trace

```GDB

```

### Relevant log output

```Shell

```

Caellian:
## Scroll direction always reports DOWN (#2288)

Old code never read `XIScrollClassInfo`:

1. Scroll valuator indices hardcoded as enum ordinals (2, 3). Real indices come from `XIScrollClassInfo::number` and vary by device.
2. Direction sign depends on `XIScrollClassInfo::increment`. Code assumed positive delta = down/right, wrong for negative-increment drivers.

We now scan `XIScrollClass` entries in `init_xi_device` to discover actual indices and store `increment` in `conky_valuator_info`. Direction = `sign(delta * increment)`. Applied in both Lua hook and propagation paths.

## Clicks near window edges dropped (closes #2288)

`window.geometry.pos()` went stale after WM repositions (decoration frames, strut adjustments). Only updated on window creation and `move_win()`.

The `ConfigureNotify` handler originally had position-tracking code that compared `window.x/y` against `ev.xconfigure.x/y` to detect WM-forced repositions and set `fixed_pos`. @brndnmtthws disabled it in 7bdb0501 because `ev.xconfigure.x/y` is frame-relative for reparented windows, which made the comparison invalid and broke window alignment by setting `fixed_pos` spuriously.

It's now replaced with `XTranslateCoordinates(window.window, root, 0, 0)` on every `ConfigureNotify`, which gives correct screen coordinates regardless of reparenting. This avoids the original problem because we never read `ev.xconfigure.x/y` and don't touch `fixed_pos` at all. The updated position gets overwritten by `update_text_area()` next cycle; it only needs to be accurate for mouse hit testing between events.

## Additional correctness fixes

- `x11_mouse_button_code` returned uninitialized `mouse_button_t` for unmapped buttons. Now returns `std::optional`, callers skip on nullopt.
- `valuator_relative_value` returned `std::optional<double>` that was never empty (indexed a zero-initialized fixed array). Now checks the valuator map first so callers can distinguish "no data" from "zero".
- `generate_events` defaulted `scroll_direction` to `Button4`, producing a phantom scroll-up on zero-delta events. Now defaults to `0` and skips generation when no direction is determined.
- `process_surface_events` used `free()` on `new`-allocated `xi_event_data` (UB, skips destructors). Changed to `delete`.
- `propagate_xinput_event` used `free()` on `new`-allocated XEvents. Changed to `delete`.

## Mouse input performance with multiple instances (closes #1886)

Every conky instance receives all root mouse events via XInput2. Each event triggered 3+ sync X11 round-trips (`XIQueryPointer`, `XQueryTree` x2) just to check if the cursor was over this instance. With many (20+) instances, we overwhelmed the X server.

We now check `window.geometry.contains(pos_absolute)` first (pure arithmetic, no X11 calls). If outside bounds, skip everything. Also cache `query_x11_top_parent(window.window)` since it only changes on reparent (invalidated in `ReparentNotify` handler). Common case goes from 3+ round-trips to zero.

**Depends on:** `ConfigureNotify` tracking position again.

## Example configuration for mouse events (closes #2066)

Added `data/conky_mouse_events.conf`. Self-contained, handler defined inline (no `lua_load`), logs all fields for all event types. Based on the exhaustive handler from PR #955.

## BUILD_XINPUT option removed, legacy fallback dropped

`BUILD_XINPUT` was an experimental flag to disable buggy XInput2 code and provide a way for users to keep using the stable base.

With the above fixes, there's no reason to keep it separate. `libXi` is a core X11 component available in all modern environments. XInput2 (or X11) code is now guarded by `BUILD_X11` directly where needed.

The legacy non-XInput2 mouse input fallback (`#else` path in MOUSE_INPUT handler) and the BORDER_CROSSING handler are removed. BORDER_CROSSING was the non-XInput fallback for EnterNotify/LeaveNotify; crossing events are already synthesized from XI_Motion in the XInput2 path.

This **removes the basic X11 mouse input fallback**, which was in principle obsolete since XInput2 shipped in 2009 - though it would've just been bloat at that point since conky had no mouse handling. My 2023 implementation could've used XInput2 exclusively, but I was cautious (and rightly so) not to introduce code I couldn't verify worked well for all users. With more users adopting mouse events and the bug reports that followed, we now have a much better understanding of the edge cases. This removal simplifies the X11 mouse handling and keeps the implementation compact for users that haven't transitioned to Wayland yet.

### DIFF
--- a/cmake/ConkyBuildOptions.cmake
+++ b/cmake/ConkyBuildOptions.cmake
@@ -223,10 +223,6 @@ dependent_option(BUILD_IMLIB2 "Enable Imlib2 support" true
 dependent_option(BUILD_XSHAPE "Enable Xshape support" true
   "BUILD_X11" false
   "Xshape support requires X11")
-dependent_option(BUILD_XINPUT "Build Xinput 2 support (slow)" false
-  "BUILD_X11" false
-  "Xinput 2 support requires X11")
-
 # if we build with any GUI support
 if(BUILD_X11)
   set(BUILD_GUI true)
--- a/cmake/ConkyPlatformChecks.cmake
+++ b/cmake/ConkyPlatformChecks.cmake
@@ -641,14 +641,11 @@ if(BUILD_X11)
       set(conky_libs ${conky_libs} ${X11_Xfixes_LIB})
     endif(BUILD_XFIXES)
 
-    # check for Xinput
-    if(BUILD_XINPUT)
-      if(NOT X11_Xinput_FOUND)
-        message(FATAL_ERROR "Unable to find Xinput library")
-      endif(NOT X11_Xinput_FOUND)
-
-      set(conky_libs ${conky_libs} ${X11_Xinput_LIB})
-    endif(BUILD_XINPUT)
+    # XInput2 (always enabled with X11)
+    if(NOT X11_Xinput_FOUND)
+      message(FATAL_ERROR "Unable to find Xinput library")
+    endif(NOT X11_Xinput_FOUND)
+    set(conky_libs ${conky_libs} ${X11_Xinput_LIB})
 
     if(X11_xcb_FOUND)
       set(HAVE_XCB true)
--- a/cmake/config.h.in
+++ b/cmake/config.h.in
@@ -64,8 +64,6 @@
 
 #cmakedefine BUILD_XFIXES 1
 
-#cmakedefine BUILD_XINPUT 1
-
 #cmakedefine BUILD_ARGB 1
 
 #cmakedefine BUILD_XDBE 1
--- a/data/conky_mouse_events.conf
+++ b/data/conky_mouse_events.conf
@@ -1,0 +1,55 @@
+-- Example: Lua mouse event handling
+--
+-- Demonstrates the lua_mouse_hook API. Run with:
+--   conky -c conky_mouse_events.conf
+--
+-- The hook receives a table with the following fields:
+--
+--   All events:
+--     type      string  "button_down", "button_up", "mouse_scroll",
+--                       "mouse_move", "mouse_enter", "mouse_leave"
+--     time      number  milliseconds since epoch
+--     x, y      number  position relative to conky window
+--     x_abs, y_abs  number  position relative to root/screen
+--
+--   button_down / button_up:
+--     button       string  "left", "right", "middle", "back", "forward"
+--     button_code  number  raw evdev button code
+--     mods         table   { shift, control, alt, super, caps_lock, num_lock }
+--
+--   mouse_scroll:
+--     direction  string  "up", "down", "left", "right"
+--     mods       table   (same as above)
+--
+--   mouse_move:
+--     mods  table  (same as above)
+--
+-- Return true from the hook to consume the event (prevent propagation
+-- to windows below conky). Return false to let it pass through.
+
+conky.config = {
+    alignment = 'top_left',
+    background = false,
+    double_buffer = true,
+    font = 'DejaVu Sans Mono:size=10',
+    gap_x = 60,
+    gap_y = 60,
+    minimum_width = 300,
+    minimum_height = 200,
+    own_window = true,
+    own_window_class = 'Conky',
+    own_window_type = 'normal',
+    own_window_hints = 'undecorated,sticky,below,skip_taskbar,skip_pager',
+    out_to_x = true,
+    update_interval = 1.0,
+    use_xft = true,
+    lua_load = './mouse_events.lua',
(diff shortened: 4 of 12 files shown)


brndnmtthws (approved):

Caellian:
I should mention that this was tested using openbox via xephyr.

I'm merging as changes are improvements in behavior and making XInput mandatory simplifies the build.

ZupixUI confirmed improvement, scroll bug he reported remains, just behaves differently - always returns "up" instead of always returning "down". This is probably a driver issue but can be addressed in a separate PR.

PR #1844, opened 2024-04-19
Separate cairo-xlib.h Lua bindings from cairo.h

Caellian:
This allows cairo.h to be used from Lua on Wayland without X11 as requirement.

A way of creating surfaces in Lua that uses Shm allocated buffers should be added in future. This PR only cleans up build options to avoid weird errors.

Closes #1804 - `BUILD_LUA_CAIRO` didn't depend on X11 and `lua/CMakeLists.txt` doesn't provide warnings like `dependent_option`s do, so it wasn't clear why the build was failing with `BUILD_LUA_CAIRO && !BUILD_X11`.

PR #2341, opened 2026-04-18
feat(wayland): expose cairo surface and window size to Lua

PlaidPiper:
## Description

**What this changes.** Completes the Wayland + Lua-cairo path that
PR #1844 set up at the build-system level. With `BUILD_LUA_CAIRO`
enabled on a non-X11 build, `cairo.h` could already be `require`d
from Lua, but `conky_window.cairo_surface` was never populated —
leaving native-Wayland Lua configs unable to obtain a cairo context.
This PR wires the live Wayland cairo surface and window dimensions
into the `conky_window` table so Wayland Lua configs can use cairo
the same way X11 Lua configs do today.

Closes the TODO Caellian flagged in #1844's description:

> A way of creating surfaces in Lua that uses Shm allocated buffers
> should be added in future.

The approach here is deliberately small — hand Lua the native
surface rather than introduce a new Lua-side surface-creation API.
Happy to take direction on whether a first-class Lua function
(e.g. `conky.get_wayland_surface()`) would be preferred; the C-side
accessor pattern matches how X11's Drawable/Visual/Display are
exposed today.

References:
- Build-system predecessor: #1844
- Original gap report: #1804

### Why the changes were necessary

Existing X11 users with `BUILD_LUA_CAIRO` can do the standard dance:

```lua
local s = cairo_xlib_surface_create(
    conky_window.display, conky_window.drawable,
    conky_window.visual, conky_window.width, conky_window.height)
local cr = cairo_create(s)
```

Wayland users have no equivalent path. Writing a working Wayland
cairo config is impossible today unless they drop back to X11 or
ship a sibling buffer allocator. This PR closes that gap by giving
Lua a ready-to-use `cairo_surface_t` on Wayland:

```lua
local cr = cairo_create(conky_window.cairo_surface)
```

### Upgrade impact on existing users

**None.** This patch is strictly additive on a feature slice no existing
user could have been using, so an existing config that works on current
`main` works byte-for-byte identical after this merge.

- **conky.conf config keys:** zero new settings, zero renamed settings,
  zero defaults flipped. Verified by grepping all touched files for
  `conky::simple_config_setting` / `config_setting` — the only matches
  are pre-existing declarations.
- **Lua `conky_window` table:** no existing key renamed, re-typed, or
  removed. One new key (`cairo_surface`) is populated on Wayland when
  `BUILD_LUA_CAIRO` is on, absent otherwise. X11 users see the exact
  same table they saw before.
- **Build matrix:** zero CMake option changes, zero new dependencies.
  Packagers building with `BUILD_WAYLAND=OFF` or `BUILD_LUA_CAIRO=OFF`
  get zero new code in their binary — every new block is gated by one
  or both of those feature macros.
- **Runtime switch:** `out_to_wayland` (pre-existing) gates every new
  code path. If the user has not set `out_to_wayland = true` in their
  config, the Wayland push is unreachable.

No deprecation warnings, no migration notes, no `conky.conf` changes
required. Users who currently depend on `conky_window.cairo_surface`
being `nil` on Wayland as a compositor sniff would see that value flip
to non-nil — but such code cannot exist in practice because there has
been no prior way to obtain a usable Wayland cairo surface through the
Lua table; any such sniff would have been dead-code detection.

### Effect on existing behavior

Strictly additive. No behavior change for:

- Pure X11 builds — unchanged; only the X11 registration and X11 push
  path runs at init and at `llua_setup_window_table` time.
- Wayland builds without `BUILD_LUA_CAIRO` — unchanged; both new code
  blocks are gated on `BUILD_LUA_CAIRO`.
- Dual X11+Wayland builds — unchanged at runtime; `out_to_wayland.get()`
  gates the actual push, so X11-only users see the X11 shape they
  already had. Registering the same tolua type twice is safe (tolua's
  hash upsert is idempotent), but the code is structured so each
  registration leg guards its own feature.
- ncurses / HTTP / file-only builds (`BUILD_GUI=OFF`) — unchanged; the
  whole conky_window table construction sits inside `#ifdef BUILD_GUI`.
- macOS CI (`BUILD_WAYLAND=OFF`) — unchanged; the Wayland leg doesn't
  compile. Verified locally with the exact CI invocation.

### How the changes were tested and validated

Three-commit branch (feat → fix → test):

- `feat(wayland): expose cairo surface and window size to Lua` — the
  core implementation, reviewed against a self-review checklist (see
  below).
- `fix(wayland): register cairo_surface_t tolua usertype before
  llua_init returns` — addresses a v1 defect discovered during
  downstream smoke-testing on a real Wayfire session; see "Downstream
  smoke-test findings" below.
- `test(wayland): add registration guard for cairo_surface_t tolua
  usertype` — Catch2 regression guard so the fix can't silently
  regress.

**Local validation on Linux x86_64 (Gentoo, clang 21):**

```bash
cmake . -B build -G Ninja \
  -DMAINTAINER_MODE=ON \
  -DBUILD_WAYLAND=ON \
  -DBUILD_LUA_CAIRO=ON \
  -DBUILD_X11=OFF \
  -DCHECK_CODE_QUALITY=ON
ninja -C build
ctest --test-dir build --output-on-failure   # 39/39 pass
clang-format --Werror --dry-run <touched files>  # 0 violations on added lines
```

**Matrix verification** (config + llua.cc compile smoke):

| Matrix | BUILD_X11 | BUILD_WAYLAND | BUILD_LUA_CAIRO | Result |
| --- | --- | --- | --- | --- |
| Primary (this PR's target) | OFF | ON | ON | ✅ builds, 39/39 tests |
| macOS CI shape | OFF | OFF | — | ✅ llua.cc compiles |
| X11-only Linux | ON | OFF | ON | ✅ llua.cc compiles |
| Wayland, no Cairo | OFF | ON | OFF | ✅ llua.cc compiles |

**End-to-end smoke test on a live compositor:** built conky with the
patch applied under Gentoo Portage, running against Wayfire 0.10 with
a user Lua config that does
`cairo_create(conky_window.cairo_surface)` in a `lua_draw_hook_pre`
callback and renders gradient time-series graphs for CPU/GPU/RAM/
network/cooling. Screenshot below; conky has been running uninterrupted
with no `traps: conky` entries in dmesg since the fix landed.

### Downstream smoke-test findings (why there are three commits, not one)

The initial commit on this branch (`e33b0af`) addressed seven
self-review issues before it was submitted upstream, but the feature
was **not** end-to-end smoke-tested on a real compositor until the
branch was about to be submitted. Doing that smoke test exposed a
defect the offline build and unit tests could not:
`tolua_pushusertype(lua_L, surface, "cairo_surface_t")` was being
called without a matching
`tolua_usertype(lua_L, "cairo_surface_t")` registration earlier in
the Lua state's lifetime. Without that registration, tolua stamps
userdata with a NULL metatable; the first Lua-side cairo call then
dereferences junk in liblua's metatable-lookup path and crashes with
a general protection fault. Symptom: `traps: conky[<pid>] general
protection fault ... in liblua5.4.so.0.0.0[2e7d0,...]` once per
`update_interval` tick.

X11 never hits this because the `#if defined(BUILD_X11)` block at the
end of `llua_init()` explicitly registers `Drawable`/`Visual`/`Display`
before `llua_setup_window_table()` ever runs. The fix mirrors that
pattern for `cairo_surface_t` under `BUILD_WAYLAND && BUILD_LUA_CAIRO`,
and widens the `tolua_open()` guard to cover the new feature
combination when X11 is off.

Test coverage for regressions:

- The original null-guard tests in `tests/test-display-wayland.cc`
  remained valid but did not (and could not) exercise the registration
  path, because they rely on the accessors fail-closed path where the
  push is skipped entirely.
- The new `test(wayland)` commit adds a third Catch2 section,
  `llua_init registers cairo_surface_t tolua usertype`, which pushes a
  sentinel pointer with the tag and asks `tolua_isusertype` whether it
  recognises the stack value. A zero result means the crash shape has
  returned; non-zero means the registration fired. This test fails
  closed if the fix is ever reverted or the guard drifts.

Flagging this sequence explicitly because reviewers often ask whether
late-discovered regressions indicate insufficient review discipline —
in this case the gap was a missing environment, not missing rigor: the
unit tests run in a test binary that never executes the C→Lua handoff
path the real runtime uses. Open to suggestions on how to surface that
boundary in CI (e.g. a headless wlroots session spinning up under
`ctest`), though I'd rather keep this PR narrowly scoped to the gap
it's fixing.

### Authorship and process disclosure

**This PR is human + AI pair work.** Flagging it up front so reviewers
know what to bring.

- **Operator:** @PlaidPiper — runs a Gentoo workstation
  on Wayfire/Wayland, has been carrying this patch as a downstream
  Portage user-patch for weeks to drive a Wayland-native Lua conky
  config. Owns the engineering judgment, the test environment, the
  smoke-test discovery of the tolua registration defect, and the
  review/accept of every change on this branch.
- **Agent:** Claude Code (Anthropic's CLI agent, Opus 4.7). Drafted the
  C++ edits, commit messages, tests, documentation updates, this PR
  body, and the three-matrix compile verification. Writes under direct
  operator supervision; does not merge, push, or open PRs without
  explicit operator approval at each step.

How the collaboration actually worked here, in case it helps calibrate
review attention:

1. Operator brought a local patch that "worked" but had been flagged as
   not upstream-ready.
2. Agent audited it against `CLAUDE.md` / `AGENTS.md` /
   `.github/instructions/*` / `pull_request_template.md` and identified
   seven issues to fix before submission (extern "C" scope, missing null
   checks, double width/height writes, raw-pointer lifetime on resize,
   tolua type-tag naming, accessor shape, comment accuracy).
3. Operator installed the polished patch on a live Wayfire session —
   which immediately uncovered the missing `tolua_usertype` registration
   (the offline build passed; the live compositor didn't). That
   discovery is the reason for the second commit on this branch.
4. Operator asked the agent to add a regression guard so the defect
   cannot silently return. That's the third commit.
5. Every commit was reviewed by the operator before moving on, including
   a re-read of the maintainer rubrics at each stage.

Each commit carries a `Co-Authored-By: Claude Opus 4.7 (1M context)
<noreply@anthropic.com>` trailer, consistent with the operator's
downstream workflow for AI-assisted contributions.

The agent and operator would rather flag this openly than surprise the
project. If the project's position is that AI-assisted PRs need a
different labeling, review cadence, or extra human-in-the-loop step
before merge, let us know and we'll adapt. If the review turns up
anything that feels like "this looks like it was written by an agent
and missed a human-sense check" — please say so; that's exactly the
feedback loop we're trying to create.

### Notes for review

- **Lifetime/ownership.** `conky::get_wayland_cairo_surface()` calls
  `cairo_surface_reference()` before returning, so the caller (the
  tolua stack) holds a reference independent of conky's own surface
  recreation on resize. The Lua-side `cairo_surface_t` userdata is
  released via cairo's existing `__gc` metamethod from `cairo.pkg`,
  which calls `cairo_surface_destroy()`.
- **Per-frame refresh in `llua_update_window_table`.** Intentional: the
  Wayland buffer/surface is destroyed and recreated on resize, so we
  refresh the Lua-visible pointer each frame. The reference-count
  accounting above makes this safe — each refresh adds a ref that is
  balanced by the previous userdata's eventual `__gc`. Removed a
  premature refactor to a diff-and-unref shape during review because
  the refresh was earlier identified as load-bearing for high-frequency
  Lua configs.
- **`tolua_open` guard widening.** Before this PR, `tolua_open(lua_L)`
  ran only under `BUILD_X11`. It now also runs under
  `BUILD_WAYLAND && BUILD_LUA_CAIRO`. Calling `tolua_open` is idempotent
  (it installs a known-position state table and bails if present), so
  dual builds are unaffected.
- **`doc/lua.yaml`** updated to add a `cairo_surface` row to the
  `conky_window` table and rewrite the X11-only NOTE to describe both
  X11 and Wayland paths.
- **`.github/instructions/cpp` rubric focus areas** that this PR
  crosses: lifetime/ownership, null dereferences, platform-guarded
  logic with matching test guards. Reviewed against each.
- **Open question for the maintainer:** would a first-class Lua
  function (e.g. `conky.get_wayland_surface()` or a method on a
  conky-specific surface object) be preferred over extending the
  `conky_window` table? The extension matches X11's existing shape,
  but Wayland is a clean slate and could be designed afresh. Happy to
  rework.

## Checklist

- [x] I have linked to any relevant GitHub issues (#1844, #1804)
- [x] Documentation in `doc/` has been updated (`doc/lua.yaml`)

image

### DIFF
--- a/doc/lua.yaml
+++ b/doc/lua.yaml
@@ -175,23 +175,27 @@ values:
       This table contains some information about Conky's window.
       The following table describes the values contained:
 
-      | Key                 | Value                                                                           |
-      |---------------------|---------------------------------------------------------------------------------|
-      | drawable            | Window's drawable (Xlib Drawable), requires Lua extras enabled at compile time. |
-      | visual              | Window's visual (Xlib Visual), requires Lua extras enabled at compile time.     |
-      | display             | Window's display (Xlib Display), requires Lua extras enabled at compile time.   |
-      | width               | Window width (in pixels).                                                       |
-      | height              | Window height (in pixels).                                                      |
-      | border_inner_margin | Window's inner border margin (in pixels).                                       |
-      | border_outer_margin | Window's outer border margin (in pixels).                                       |
-      | border_width        | Window's border width (in pixels).                                              |
-      | text_start_x        | The x component of the starting coordinate of text drawing.                     |
-      | text_start_y        | The y component of the starting coordinate of text drawing.                     |
-      | text_width          | The width of the text drawing region.                                           |
-      | text_height         | The height of the text drawing region.                                          |
+      | Key                 | Value                                                                                           |
+      |---------------------|-------------------------------------------------------------------------------------------------|
+      | drawable            | Window's drawable (Xlib Drawable). X11 only; requires Lua extras enabled at compile time.       |
+      | visual              | Window's visual (Xlib Visual). X11 only; requires Lua extras enabled at compile time.           |
+      | display             | Window's display (Xlib Display). X11 only; requires Lua extras enabled at compile time.         |
+      | cairo_surface       | The live `cairo_surface_t` backing the window. Wayland only; populated when `out_to_wayland` is |
+      |                     | enabled and `BUILD_LUA_CAIRO` was set at compile time. Pass directly to `cairo_create()`.       |
+      | width               | Window width (in pixels).                                                                       |
+      | height              | Window height (in pixels).                                                                      |
+      | border_inner_margin | Window's inner border margin (in pixels).                                                       |
+      | border_outer_margin | Window's outer border margin (in pixels).                                                       |
+      | border_width        | Window's border width (in pixels).                                                              |
+      | text_start_x        | The x component of the starting coordinate of text drawing.                                     |
+      | text_start_y        | The y component of the starting coordinate of text drawing.                                     |
+      | text_width          | The width of the text drawing region.                                                           |
+      | text_height         | The height of the text drawing region.                                                          |
 
-      NOTE: This table is only defined when X support is
-      enabled.
+      NOTE: This table is defined when either X11 or Wayland
+      graphical output is enabled. On X11, use `drawable`/`visual`/
+      `display` to construct a cairo surface; on Wayland, the live
+      surface is provided directly as `cairo_surface`.
   - name: ret_scale_x,ret_scale_y:cairo_draw_image(file, cs, x, y, scale_x, scale_y)
     desc: |-
       Renders an image onto a cairo_surface_t, using imlib2. Returns the amount the image was scaled by
--- a/src/lua/llua.cc
+++ b/src/lua/llua.cc
@@ -40,6 +40,13 @@
 #include "x11-settings.h"
 #endif /* BUILD_X11 */
 
+#ifdef BUILD_WAYLAND
+// Brings in the out_to_wayland runtime setting (via wl.h) and the
+// get_wayland_cairo_surface() accessor used to hand the live Wayland
+// cairo surface to Lua, mirroring the X11 Drawable/Visual/Display path.
+#include "../output/display-wayland.hh"
+#endif /* BUILD_WAYLAND */
+
 #ifdef BUILD_MOUSE_EVENTS
 #include "../mouse-events.h"
 #endif /* BUILD_MOUSE_EVENTS */
@@ -680,12 +687,43 @@ void llua_setup_window_table(conky::rect<int> text_rect) {
   }
 #endif /*BUILD_X11*/
 
+#ifdef BUILD_WAYLAND
+  // Mirror the X11 branch above: publish a handle Lua scripts can pass to
+  // cairo_create(). The Wayland surface is populated asynchronously, so
+  // get_wayland_cairo_surface() may return nullptr here at startup; we
+  // push only when the surface is ready. update_window_table() refreshes
+  // this each frame so late-arriving surfaces are picked up.
+  if (out_to_wayland.get(*state)) {
+    cairo_surface_t *wl_surface = conky::get_wayland_cairo_surface();
+    if (wl_surface != nullptr) {
+      // The tolua type tag is the typedef name from cairo.pkg
+      // (`cairo_surface_t`), not the underlying struct tag. Using the
+      // struct tag leaves the userdata unmatched by lua's cairo_create
+      // metatable check.
+      llua_set_userdata("cairo_surface", "cairo_surface_t", wl_surface);
+    }
+  }
+#endif /*BUILD_WAYLAND*/
+
 #ifdef BUILD_GUI
   if (out_to_gui(*state)) {
 #ifdef BUILD_X11
     llua_set_number("width", window.geometry.width());
     llua_set_number("height", window.geometry.height());
 #endif /*BUILD_X11*/
+#ifdef BUILD_WAYLAND
+    // On Wayland the window dimensions live on the file-static
+    // conky::global_window tracked by display-wayland.cc. We expose them
+    // via the existing conky::window_get_width_height(struct window*,
+    // int*, int*) helper, routed through a thin wrapper that resolves
+    // global_window internally.
+    if (out_to_wayland.get(*state)) {
+      int wl_w = 0, wl_h = 0;
+      conky::get_wayland_window_size(&wl_w, &wl_h);
+      llua_set_number("width", wl_w);
+      llua_set_number("height", wl_h);
+    }
+#endif /*BUILD_WAYLAND*/
     llua_set_number("border_inner_margin", border_inner_margin.get(*state));
     llua_set_number("border_outer_margin", border_outer_margin.get(*state));
     llua_set_number("border_width", border_width.get(*state));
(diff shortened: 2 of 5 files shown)


Caellian:
Dude... too much text. I gave the PR description and diff to Claude (lol), as well as context around X11 and the plan I had after #1844, and the summary is "wrong approach".

You're mirroring X11 design which wasn't ideal to begin with. The right solution here is a unified function that works identically on both backends without having to specify surface/display/etc. internals. A deprecation path is important too. I thought about building on top of your work in your fork, but I'd keep ~1 line of it, so I'll just close this.

I appreciate your effort, time and tokens, but this shouldn't be done by a first-time contributor. I don't appreciate a wall of text that describes this little code though, your prompt should end with "Keep the description brief and technical."

PlaidPiper:
Thank you for the feedback I'll keep trying to not produce slop! And thank you for maintaining conky!

Caellian:
I can't tell how much time you put into it (beyond 2 hours between commits), but regardless, it's just not a good direction to take this code.

As I said - LLM or not, this change is something that takes some familiarity with the project to get right. If this was my first PR, I'd probably go a similar route as you did. If LLM guided you in the right way, I'd likely just push to your remote, but there was too little overlap for that to make sense.

Caellian:
See #2342

PR #2342, opened 2026-04-18
feat(lua): add `conky_surface()` for creating Cairo surface on Wayland; update `conky_window.width/height` from Wayland

Caellian:
## Summary

- Added `conky_surface()` to Lua. It exposes the Cairo surface of the current window.
  - Scripts using it should work seamlessly across display output backends.
- Replaced `cairo_xlib.pkg` with `cairo_xlib.cc` to emit  deprecation warnings on use.
  - Warnings point to wiki (will update after merge), and new `conky_surface()` function.
- Fixed Cairo include paths in CMake for X11 builds.
  - Wasn't used previously, but now `conky_surface()` needing Cairo in X11 backend surfaced include errors.
    - Technically, it should be a different feature flag, but intent is the same and meaning will shift, so no point in making the build more complicated.
- Added `conky_window` fields:
  - `conky_window.pixel_size.x`
  - `conky_window.pixel_size.y
  - `conky_window.scale`
- Removed pointless call to `llua_update_window_table` in `conky.cc` - backends call it anyway once they've adapted the window to changes made by `update_text_area`.

## Testing

- Verified `conky_surface()` with `BUILD_LUA_CAIRO_XLIB` on Wayland and Openbox (via Xephyr).
- Validated the replacement `libcairo_xlib.cc` works the same as before.

## LLM Use

Used Claude Opus 4.6 for:

- Reducing boilerplate in `tolua++` generated `libcairo_xlib.cc` bindings.
- Identifying state update paths and inserting new arguments.

Initial commit had substantial amount of code written by an LLM (Lua stack stuff is easier to test than write), I removed Co-Authored-By field later after splitting LLM written code out of this PR.


brndnmtthws (approved):

simotek:
Thankyou @Caellian if we ever meet at a conference I owe you a Beer or Coffee, I started looking at implementing this the other week then got busy with some other projects. I will test it in the next few days and then i'm glad I have some extra free time for other projects.

Caellian:
Thanks, I hold you to that then :smiley:

PR #2344, opened 2026-04-19
fix(settings): simplify background alpha settings (#2109)

### LINKED ISSUE #2015: [Bug]: Conky fake transparency doesn't work with xfdesktop (XFCE)
### What happened?

Almost 2 months ago, I made a fresh debian unstable installation on a laptop. I installed all my apps there, including conky, and I transfered my .conkyrc (the one I use for more than a decade now) to it. After that, I edited the conky.text part to correct stuff that is related to the hardware changes from the one system to the other (the sensors basically),

When conky runs on xfce though, its background is completely black. For that reason, I also installed openbox there (the wm i use on my main pc since 2009 when xfce was light years behind gnome and kde) and conky is transparent there.
As suggested by other online users online with similar issues, also tried setting own_window_argb_visual to true, but nothing changed. As a sideeffect, this variable made conky have a black background on openboox too. 
Setting own_window_type to normal did not change anything too.

Compositing on xfce is and will remain disabled, I can post a screenshot of the whole desktop if needed, and I have omitted the conky.text part of my config because it is irrelevant with the issue.

### Version

1.21.6

Debian

### Conky config

```Lua
conky.config = {
    alignment = 'bottom_right',
    background = true,
    border_width = 0,
    cpu_avg_samples = 2,
    default_color = 'ghost white',
    default_outline_color = 'black',
    default_shade_color = 'black',
    double_buffer = true,
    draw_borders = false,
    draw_graph_borders = true,
    draw_outline = false,
    draw_shades = false,
    font = 'ITC Avant Garde Gothic Pro:size=16',
    font1 = 'ITC Avant Garde Gothic Pro:size=14',
    gap_x = 0,
    gap_y = 40,
    no_buffers = true,
    out_to_x = true,
    own_window = true,
    own_window_class = 'Conky',
    own_window_type = 'desktop',
    show_graph_range = false,
    show_graph_scale = false,
    update_interval = 1.0,
    use_xft = true,
}
```

### Stack trace

_No response_

### Relevant log output

_No response_

### LINKED ISSUE #2109: [Enhancement]: Cleanup color management
@bi4k8 mentioned this in #1504 and I agree:

> I think we need to decide once and for all how to handle alpha in Conky: is it separate from colors, or included with them? We conflate ends and means in the current configuration scheme: `own_window_transparent` overrides `own_window_argb_value`, but should really be equivalent to `own_window_argb_value = 0`, except that it actually triggers a different implementation that can work in more X11 configurations. Meanwhile, color parsing does support reading alpha values, except for places where the parsed alpha is overridden by `xftalpha` or `own_window_argb_value`.
> 
> _Originally posted by @bi4k8 in https://github.com/brndnmtthws/conky/issues/1504#issuecomment-1512159257_

This is very confusing to deal with by users. As a user I'd expect a single `window_background = "color"` property that also accepts alpha, and then for conky to deal with all the implementation details.

There are however 3 different config options for this "1 concept":
- `own_window_argb_value`
- `own_window_colour`
- `own_window_transparent`

And additionally `own_window_argb_visual` that, when I forget about implementation details of X11, doesn't tell me much.

I suggest the first three to be replaced with simply `window_background_colo(u)r`, and for the `own_window_argb_visual` to be renamed to `enable_x11_true_transparency` because that communicates so much more details at a first glance:
- I know there's "fake" transparency
- I know it's only X11
  - because Wayland makes it hard to do the other (worse) thing
- I know it has (almost) nothing to do with actual "rgb" color value.

### LINKED ISSUE #2301: [Bug]: Wayland: Conky fails to draw background, always transparent regardless of own_window_argb_value
### What happened?

Description:

Conky on Wayland has a critical bug where it does not draw a window background at all. This results in the conky window being completely transparent, making it invisible, regardless of the own_window_argb_value setting (even when set to 255 for full opacity).

This is not a pseudo-transparency issue; it is a complete failure to render the background. The own_window_argb_visual setting appears to be non-functional.

Environment:

    OS: NixOS 25.11
    Window Manager: Sway
    Display Server: Wayland
    Conky Version:
      conky 1.22.2 compiled for Linux x86_64
      
      Compiled in features:
      
      System config file: /etc/conky/conky.conf
      Package library path: /nix/store/wyqs9p3c7cxwidgi54y3d4jpd7c5q6r1-conky-1.22.2/lib/conky
      
      
       General:
        * math
        * hddtemp
        * portmon
        * IPv6
        * Curl
        * RSS
        * wireless
        * support for IBM/Lenovo notebooks
        * builtin default configuration
        * old configuration syntax
        * Imlib2
        * OSS mixer support
        * apcupsd
        * iostats
        * ncurses
        * Internationalization support
      
       Lua bindings:
        * Cairo
        * Imlib2
       X11:
        * Xdamage extension
        * Xinerama extension (virtual display)
        * Xshape extension (click through)
        * XDBE (double buffer extension)
        * Xft
        * ARGB visual
        * Own window
        * Mouse events
       Wayland:
        * ARGB visual
        * Mouse events
      
       Music detection:
        * CMUS
        * MPD
        * MOC
      
       Default values:
        * Netdevice: eno1
        * Local configfile: $HOME/.conkyrc
        * Localedir: /nix/store/wyqs9p3c7cxwidgi54y3d4jpd7c5q6r1-conky-1.22.2/share/locale
        * Maximum netdevices: 256
        * Maximum text size: 16384
        * Size text buffer: 256
    Conky Package: Nixpkgs (built with Wayland support)

Expected Behavior:
Setting own_window_argb_value = 255 should create a fully opaque background.

Actual Behavior:
The background is never drawn, resulting in complete transparency.

Additional Context:
This appears to be related to issue #1723 regarding buffer flushing with own_window_argb_visual on Wayland. This bug makes Conky unusable for any use case requiring a visible background on Wayland compositors.

--

### Version

1.22.2

Linux (other)

### Conky config

```Lua
--
-- configurations.
--
--
--
-- For more on Lua, see:
--

conky.config = {
    alignment = 'top_left',
    background = false,
    border_width = 1,
    cpu_avg_samples = 2,
    default_color = 'red',
    default_outline_color = 'green',
    default_shade_color = 'green',
    double_buffer = true,
    draw_borders = false,
    draw_graph_borders = true,
    draw_outline = false,
    draw_shades = false,
    extra_newline = false,
    gap_x = 10,
    gap_y = 10,
    minimum_height = 5,
    net_avg_samples = 2,
    no_buffers = true,
    out_to_x = false,
    out_to_wayland = true,
    own_window = true,
    own_window_colour = '#767676',
    own_window_argb_value = 100,
    own_window_type = 'override',
    own_window_class = 'Conky',
    own_window_hints = 'undecorated,sticky,below,skip_taskbar,skip_pager',
    show_graph_range = false,
    show_graph_scale = false,
    stippled_borders = 0,
    update_interval = 1.0,
    uppercase = false,
    use_spacer = 'none',
    use_xft = true,
    minimum_width = 450,
    maximum_width = 450,
}
```

### Stack trace

```GDB

```

### Relevant log output

```Shell

```

Caellian:
This PR deprecates:
- `own_window_argb_visual` - just set the color to opaque
- `own_window_argb_value` - just set the alpha on the color
- `own_window_transparent` - just set the alpha=0 on the color

It removes `BUILD_ARGB` flag - it's been supported by X11 for quite some time.

It's backwards compatible: using old settings will result in same behavior as before (ignoring the fixed bugs), only difference is that `own_window_argb_visual = false` won't make the window opaque (ignoring other alpha settings), but this could also confuse some users.

16 value combinations are now resolved by specificity, which should be the most intuitive option for users.

## Testing
- [x] Closes #2301: overwrote bg alpha unconditionally with `own_window_argb_value`.
- [x] Closes #2015: fallback (without compositor) in X11 code always set the transparent color when built with `BUILD_ARGB`, without compositor running this means black.
- [x] Closes #2109: asks for this fix directly.

**LLM use:** rubber ducky and looking through ancient git history to figure out rationale behind each option.

### DIFF
--- a/cmake/ConkyBuildOptions.cmake
+++ b/cmake/ConkyBuildOptions.cmake
@@ -205,9 +205,6 @@ else()
     "Xfixes support requires X11")
 endif(OS_DARWIN)
 
-dependent_option(BUILD_ARGB "Build ARGB (real transparency) support" true
-  "BUILD_X11;OWN_WINDOW" false
-  "ARGB support requires X11 and OWN_WINDOW enabled, not needed on Wayland")
 dependent_option(BUILD_XINERAMA "Build Xinerama support" true
   "BUILD_X11" false
   "Xinerama support requires X11")
--- a/cmake/config.h.in
+++ b/cmake/config.h.in
@@ -64,8 +64,6 @@
 
 #cmakedefine BUILD_XFIXES 1
 
-#cmakedefine BUILD_ARGB 1
-
 #cmakedefine BUILD_XDBE 1
 
 #cmakedefine BUILD_PORT_MONITORS 1
--- a/doc/config_settings.yaml
+++ b/doc/config_settings.yaml
@@ -410,17 +410,6 @@ values:
     desc: Overwrite the file given as argument.
   - name: own_window
     desc: Boolean, draw conky in own window instead of drawing on root window.
-  - name: own_window_argb_value
-    desc: |-
-      **Deprecated.** Use [`own_window_colour`](#own_window_colour) with an alpha
-      channel instead (e.g. `'#8000'` for 50% opacity).
-
-      Sets the background opacity. Valid range is 0-255, where 0 is fully
-      transparent and 255 is fully opaque. Overrides the alpha channel of
-      `own_window_colour` when set to a value below 255.
-    default: 255
-    args:
-      - integer_number
   - name: own_window_class
     desc: Manually set the WM_CLASS name.
     default: Conky
@@ -463,13 +452,6 @@ values:
     desc: |-
       Allows overriding conky window name.
     default: 'conky (<hostname>)'
-  - name: own_window_transparent
-    desc: |-
-      **Deprecated.** Use [`own_window_colour`](#own_window_colour) with alpha 0
-      instead (e.g. `'#0000'`).
-
-      Sets background opacity to 0%, making the background fully transparent.
-    default: 'false'
   - name: own_window_type
     desc: |-
       If `own_window` is set, under X11 you can specify type of window conky
--- a/doc/config_settings.yaml
+++ b/doc/config_settings.yaml
@@ -412,27 +412,27 @@ values:
     desc: Boolean, draw conky in own window instead of drawing on root window.
   - name: own_window_argb_value
     desc: |-
-      When ARGB visuals are enabled, this use this to modify the
-      alpha value used. Valid range is 0-255, where 0 is 0% opacity, and 255
-      is 100% opacity.
+      **Deprecated.** Use [`own_window_colour`](#own_window_colour) with an alpha
+      channel instead (e.g. `'#8000'` for 50% opacity).
+
+      Sets the background opacity. Valid range is 0-255, where 0 is fully
+      transparent and 255 is fully opaque. Overrides the alpha channel of
+      `own_window_colour` when set to a value below 255.
     default: 255
     args:
       - integer_number
-  - name: own_window_argb_visual
-    desc: |-
-      Boolean, use ARGB visual? ARGB can be used for real transparency, note
-      that a composite manager is required on X11.
-      This option will not work as desired (in most cases) in conjunction with
-      ['own_window_type override'](#own_window_type).
   - name: own_window_class
     desc: Manually set the WM_CLASS name.
     default: Conky
   - name: own_window_colour
     desc: |-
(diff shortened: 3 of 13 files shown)


brndnmtthws (approved):

Caellian:
An issue raised on linked PR warrants testing before fast-forwarding. I didn't really test this much yet. Thanks for approval. Will merge once I test old vs. new against the issues -have to install X11 first because some X11 WMs crash in Xephyr :frowning_face:.

Issue #2310, opened 2026-01-17
[Feature]: support variable definitions in Lua


rubyFeedback:
I would like to suggest to add functions to conky.

With this I do not mean lua; I mean functions definitions and function calls in conkyrc.

Let me explain this with an example next.

Right now I have this to show the top 4 processes:

    ${color grey}Name              PID     CPU%   MEM%
    ${color #eee} ${top name 1} ${top pid 1} ${top cpu 1} ${top mem 1}
    ${color #ccc} ${top name 2} ${top pid 2} ${top cpu 2} ${top mem 2}
    ${color #aaa} ${top name 3} ${top pid 3} ${top cpu 3} ${top mem 3}
    ${color #888} ${top name 4} ${top pid 4} ${top cpu 4} ${top mem 4}

I would like to actually bundle this in a function such as:

    show_the_top_four_processes

For instance:

    def show_the_top_four_processes()
      ${color grey}Name              PID     CPU%   MEM%
      ${color #eee} ${top name 1} ${top pid 1} ${top cpu 1} ${top mem 1}
      ${color #ccc} ${top name 2} ${top pid 2} ${top cpu 2} ${top mem 2}
      ${color #aaa} ${top name 3} ${top pid 3} ${top cpu 3} ${top mem 3}
      ${color #888} ${top name 4} ${top pid 4} ${top cpu 4} ${top mem 4}
    end

This could be treated as simple macro-expansion, so not really a "full"
function. Though, a full function including passing arguments, may be
nice to have - but my idea here is solely for cleaning up the conkyrc
file, so that I have fewer # and $ variables to look at.

I don't care if a () is mandatory or not; if it is mandatory we can then call
it like this:

    show_the_top_four_processes()

My idea here is to have my conkyrc file very clean. Rather than a gazillion
$ variables, I would like to group this more logically, so that I can simply
re-arrange the function calls that are more descriptive to me.

The benefit would be that I could simplify my conkyrc a LOT.

Also, I would like to be able to store these functions elsewhere on my
filesystem. Right now I think we must store everything in a single
(or mostly one) conkyrc file. I'd like to be able to call any local file
actually - at the least for such a conkyrc file.

Anyway - I don't know how other people see this, so perhaps please
comment, even if this proposal is closed. I'd like to know what other
people think about this - or perhaps my ideas are very off, right now
I don't know.

Grim-GH:
Already possible using Lua?
[...]

Issue #2240, opened 2025-09-02
[Bug]: Inconsistency of spelling in variable own_window_colour


slaszt:
## Problem

The variable `own_window_colour` uses the British/Canadian/etc. spelling, whereas other colour variables such as `default_color` and `colorN` do not. This inconsistency makes troubleshooting more difficult when the window colour does not display as expected.

## Expected behaviour

It is expected that variables use the same spelling of common words to avoid ambiguity and errors when editing configuration files.

## Potential solution

Perhaps `own_window_colour` could be changed to `own_window_color` to harmonize the naming of colour variables. However, this would have the effect of breaking existing configurations that use the current variable.

### Version

1.18.3

### Which OS/distro are you seeing the problem on?

Debian

### Conky config

```Lua

```

### Stack trace

```GDB

```

### Relevant log output

```Shell

```

PR #2347, opened 2026-04-25
reafactor(variables): Add variable registry

### LINKED ISSUE #2052: Cleanup `text_object` construction in core.cc

Caellian:
This is a large refactor that untangles most of the code from `core.cc`. It removes the huge if-elseif-elseif chain in `core.cc` and replaces it with a registry pattern and `std::unordered_map` lookup.

Improvements:
- Removes a lot of headers that were exporting only printing functions previously.
- Opens up possibility of automating documentation coverage.
- Simplifies variable aliases (#2240 cleaner).
- Allows dynamic insertion of variables (a must for #2310).
- Speeds up compilation because `core.cc` no longer needs blocks on everything else.
- Variables are declared next to related code. 100x more readable.
- Parsing variables is `O(1)` amortized, instead of `O(n)`. (Though this wasn't really a huge cost)

Changes that affect functionality (fix bugs) are separated into their own commits so they can be reverted separately if needed; none of `refactor` commits affect behavior.

<details>
<summary>This PR requires bumping GCC minimal supported version to 13 (expand for details)</summary>

<hr/>

GCC 13.1 was released: April 26, 2023

GCC 13.1 is needed because a [macro added in `variables.hh`](https://github.com/brndnmtthws/conky/pull/2347/changes#diff-dea3766914e9c4e91da66cc9644cdce00fa2935e0d8073586e27cfd423700149R235-R241) relies on [P1907R1](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1907r1.html) ("Inconsistencies with non-type template parameters") being applied as a defect resolution to C++17 mode. The `print_variable` macro bridges C++17 lambdas into NTTP function pointers via a local struct's static member, whose address has no linkage under too-broad [C++17 [basic.link]/8 rule](https://timsong-cpp.github.io/cppwp/n4659/basic.link#8). GCC 13+ and Clang treat P1907R1 as a DR and accept this in C++17 mode; GCC 11/12 do not. In C++20 this workaround becomes unnecessary because stateless lambdas are valid NTTPs directly.

Intuitively, this makes sense to be a DR because a non-capturing lambda is fully static, and a struct declared in it should be statically linkable too.

Not having P1907R1 means that we can't use non-capturing lambdas for variable parsing, which avoidable but ugly because it reverts to having ~100 functions that are 1-3 lines of "read and print variable" code. DR makes this much cleaner.

Ubuntu 24.04 does have GCC 13/14, so this change doesn't affect compatibility matrix for official distro packaging (buildd), 22.04 will only backport security vulnerabilities.

GCC issue: [gcc#96645](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=96645)

<hr/>

</details>

Closes #2052.

This is WIP until I move all variable parsing out of `core.cc`. But each commit compiles and runs in isolation and can be tested as-is.


brndnmtthws:

brndnmtthws:
As a heads-up: I am going to change the branch policy to disallow force pushes to any branch (not just `main`), just in case anyone with collaborator access gets compromised again.

Caellian:
> As a heads-up: I am going to change the branch policy to disallow force pushes to any branch (not just main), just in case anyone with collaborator access gets compromised again.

@brndnmtthws That breaks rebasing, makes PRs extremely annoying to use, and doesn't really improve security. 

Without force-pushes, PR branches effectively become append-only:
- rebasing onto updated main is impossible
- fixup commits and squash are broken on working branches
- long-running branches accumulate merge commits or stale history
  - reviewers can't review commit diff anymore because - you're forcing them to review the final diff because it's the only non-noisy one
  - these PRs now must be squash merged, even if parts of them were reversible before
    - disables revert commits for larger changes (e.g. refactor + one tiny change)

A collaborator with previous contribution can just push a new commit on top of someone else's PR, approve and merge the branch. Someone with alt accounts that both contributed can also still make a PR and verify it on their own.

The only solution is to limit approvals, and make them auto-retract when branch is mutated. At the moment, 185 accounts (besides yours) have mutable access to main, if there's an alt account among those, a single person can push malicious code to main on their own, regardless of branch/PR mutability.

Signing commits with GPG is what's supposed to address this problem, but that introduces friction for new developers.

What I suggest you do instead:
- auto-retract approvals on PR branch change,
- an additional tier of trust,
  - limit approvals to lasers, bi4k8, su8. g0mba, mmuman, ..., me
  - manage membership via a CI cron job: something like `gh pr list --repo brndnmtthws/conky --state merged --limit 1000 --json author --jq '[.[].author.login] | group_by(.) | map({author: .[0], count: length}) | sort_by(-.count)'`, but time-amortized.

With that in place, you don't disable majority of git functionality and it solves the problem you're pointing to without the solution being a nuisance or enforcing bad practices GitHub already makes all too easy.

brndnmtthws:
The main issue here was that someone rewrote the git history, which is annoying but not the worst thing. Let me think a bit more about it, I don't want to add too much friction and disabling force pushes just means you need to use merges instead of rebasing in your branch, but the final merge into master is always a rebase. I'll remove the rule for now but I might change my mind. I don't love the idea of requiring signed commits because most people don't sign their commits properly and invalidating approvals can be annoying when you want to approve the PR even if there are still some more follow-up changes needed. There's a balance here between avoiding bad things (such as likely credential compromise in this case) by making sure history can't be rewritten and making it super inconvenient for people.

Caellian:
Ok, thank you for staying ahead of this. In this instance, my PRs were safe because I don't really ever pull from PR branches so their changes would've been overwritten once I rebased my local worktrees.

I don't merge using GitHub Web UI because it's incapable of preserving even commits that are directly on main.

I see it could be an issue for other contributors though and something that could erode trust towards people who are trying to enter the project.

Not promising anything, but I might make a PR for the proposed solution, and rethink approval retraction (maybe make it conditional and trigger under certain circumstances (a new account appearing in commits)). GH triggers CI events even on force pushes (commit overwrites), so that shouldn't be much of an issue.

PR #2348, opened 2026-04-25
Fix #2333: implement hash-based caching for execp to persist scroll state

### LINKED ISSUE #2333: [Bug]: ${scroll} does not work from within Bash file
### What happened?

When i execute a bash file withing conky with `${execp bashfile.sh}` the scroll function doesn't work! It shows the text but is not scrolling.

```
echo "\${scroll wait 10 1 6 This text needs to be scrolled, but does not!}"
```

### Version

1.19.8

Ubuntu

### Conky config

```Lua
conky.config = {
    background = false,
    alignment = 'top_middle',
    gap_x = -900,
    gap_y = -428,
    minimum_width = 400,.
    minimum_height = 100,
    no_buffers = true,
    double_buffer = true,
    text_buffer_size = 2048,
    draw_borders = false,
    draw_outline = false,
    draw_shades = false,
    draw_graph_borders = false,
    uppercase = false,
    out_to_stderr = false,
    own_window = true,
    own_window_title = 'conky-mediaplayer',
    own_window_type = 'normal',
    own_window_hints = 'undecorated,below,skip_taskbar,skip_pager',
    use_xft = true,
    font = 'Liberation Sans:size=7',
    xftalpha = 1,
    override_utf8_locale = true,
    update_interval = 0.3,
    default_color = 'white',
    imlib_cache_size = 10000,
    imlib_cache_flush_interval = 60,
    lua_load = './image.lua ./time-bar.lua',
    lua_draw_hook_pre = 'main_bars',
};

conky.text = [[
${execp ./player.sh}
${execi 1 ./cover.sh}
#${exec bash ./click_event.sh} # > /dev/null 2>&1}
]];
```

### Stack trace

```GDB

```

### Relevant log output

```Shell

```

harshal24-chavan:
# Checklist

## Description
This PR implements hash-based caching to persist the scroll state.
Fixes #2333

1. Added a `last_execp_hash` member to the text_object. 
2. In src/data/exec.cc, I modified fill_p to compare the current output hash with the `last_execp_hash` hash.
3. If the hash matches, the existing object tree (obj->sub) is preserved. We call `extract_variable_text_internal` and `generate_text_internal` on the existing tree instead of re-parsing.

* Describe how the changes will affect existing behaviour.
1. For ${scroll}: It now persists its animation state and "circles/loops" correctly even when inside an execp call.
2. For other variables: Dynamic variables (like ${cpu}) still update correctly because extract_variable_text_internal is called every cycle

1. Environment: Tested on Arch Linux (Kernel 6.x).
2. Verified that a ${scroll} inside a bash script loops matches the timing of a standard static scroll.
3. Dynamic Update: Verified that changing the bash script output triggers a re-parse (due to hash mismatch) and updates the display immediately
4. Stability: Monitored via bpytop; memory remained stable at ~15MB with no leaks detected over an extended period. CPU usage remained consistent.
5. Style: Code was formatted using clang-format according to the project's .clang-format specification.

* Include any relevant screenshots/evidence demonstrating that the changes work and have been tested.

### DIFF
--- a/src/content/text_object.h
+++ b/src/content/text_object.h
@@ -121,6 +121,8 @@ struct text_object {
   } data;
 
   void *special_data;
+
+  size_t last_execp_hash; /* caching-hash for execp content*/
   long line;
   bool parse;  /* if true then data.s should still be parsed */
   bool thread; /* if true then data.s should be set by a separate thread */
--- a/src/data/exec.cc
+++ b/src/data/exec.cc
@@ -43,10 +43,11 @@
 #include "../content/text_object.h"
 #include "../update-cb.hh"
 
-struct execi_data {
+struct exec_data {
   float interval{0};
   char *cmd{nullptr};
-  execi_data() = default;
+  size_t last_execp_hash{0};
+  exec_data() = default;
 };
 
 static const int cmd_len = 256;
@@ -228,7 +229,9 @@ void fill_p(const char *buffer, struct text_object *obj, char *p,
   if (obj->parse) {
     size_t current_hash = std::hash<std::string>{}(buffer);
 
-    if (obj->last_execp_hash == 0 || current_hash != obj->last_execp_hash) {
+    auto *ed = static_cast<struct exec_data *>(obj->data.opaque);
+
+    if (ed->last_execp_hash == 0 || current_hash != ed->last_execp_hash) {
       if (obj->sub) {
         free_text_objects(obj->sub);
         free(obj->sub);
@@ -260,15 +263,13 @@ void fill_p(const char *buffer, struct text_object *obj, char *p,
  * @param[in] execflag bitwise flag used to specify the exec variant we need to
  * process
  */
-void scan_exec_arg(struct text_object *obj, const char *arg,
-                   unsigned int execflag) {
+void scan_exec_arg(struct text_object *obj, const char *arg, exec_flag flags) {
   const char *cmd = arg;
   char *orig_cmd = nullptr;
-  struct execi_data *ed;
+  auto ed = new exec_data;
 
   /* in case we have an execi object, we need to parse out the interval */
-  if ((execflag & EF_EXECI) != 0u) {
-    ed = new execi_data;
+  if (flags & exec_flag::interval) {
     int n;
 
     /* store the interval in ed->interval */
@@ -285,12 +286,12 @@ void scan_exec_arg(struct text_object *obj, const char *arg,
   }
 
   /* parse any special options for the graphical exec types */
-  if ((execflag & EF_BAR) != 0u) {
+  if (flags & exec_flag::bar) {
     cmd = scan_bar(obj, cmd, 100);
 #ifdef BUILD_GUI
-  } else if ((execflag & EF_GAUGE) != 0u) {
+  } else if (flags & exec_flag::gauge) {
     cmd = scan_gauge(obj, cmd, 100);
-  } else if ((execflag & EF_GRAPH) != 0u) {
+  } else if (flags & exec_flag::graph) {
     auto [buf, skip] = scan_command(cmd);
     scan_graph(obj, cmd + skip, 100, FALSE);
     cmd = buf;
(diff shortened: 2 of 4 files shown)


src/data/exec.cc
```diff
@@ -225,11 +226,24 @@ static inline double get_barnum(const char *buf) {
 void fill_p(const char *buffer, struct text_object *obj, char *p,
             unsigned int p_max_size) {
   if (obj->parse) {
-    evaluate(buffer, p, p_max_size);
+    size_t current_hash = std::hash<std::string>{}(buffer);
+
+    if (obj->last_execp_hash == std::numeric_limits<size_t>::max() ||
```
Caellian: This check is wrong.

You initialize with `size_t last_execp_hash = std::numeric_limits<size_t>::max()`, but you don't actually - the code that constructs `text_object` uses `malloc` and `memset(0)`.

There's no difference between 0 and max sentinel (hash is equally likely to be either), so you should just simplify this check to check for zero.

src/content/text_object.h
```diff
@@ -121,6 +121,9 @@ struct text_object {
   } data;
 
   void *special_data;
+
+  size_t last_execp_hash =
+      std::numeric_limits<size_t>::max(); /* caching-hash for execp content*/
```
Caellian: and remove initializer here (doesn't get used so it's confusing).

src/data/exec.cc
```diff
@@ -225,11 +226,24 @@ static inline double get_barnum(const char *buf) {
...
+        current_hash != obj->last_execp_hash) {
+      if (obj->sub) {
+        free_text_objects(obj->sub);
+        delete obj->sub;
+      }
+      obj->sub = new text_object{};
```
Caellian: Don't use `new` here. It's good C++ code, but the old code isn't, so you're mixing allocation strategies - older code uses `malloc` and `free`. If some other code calls free instead of delete on this it might blow up.

src/content/text_object.h
```diff
@@ -121,6 +121,9 @@ struct text_object {
   } data;
 
   void *special_data;
+
+  size_t last_execp_hash =
```
Caellian: Also, this shouldn't live on every `text_object` when only `execp` needs it. Move it into `execi_data` which is unique per-exec variable and won't bloat all text objects..

Caellian (changes requested):

src/data/exec.cc
harshal24-chavan: Thanks for your reviews, I'll get to work on these and update the PR.

Lowrida:
So is this problem now solved or.. ?

Caellian (dismissed):
I told you:

> Also, this shouldn't live on every `text_object` when only `execp` needs it. Move it into `execi_data` which is unique per-exec variable and won't bloat all text objects..

I was wrong. This caused UB because `execp` without interval actually allocated a `char *` instead of `execi_data`. So I fixed it on your branch.

I just need your ok on merging this because I manufactured history and marked you as the author of the squashed and modified commit. Squash was preparation for fast-forward, and I just wanted to streamline the process and reduce back-and-forth. Review the diff on your commit and let me know whether it's what you would've written (`text_object` works and is fine).

My commit on top is just reducing technical debt that modifying `text_object` would introduce and some 10 year old code that prevented you from using `execi_data` in the first place (which is a slightly better choice).

You can revert your commits from a [bundle](https://github.com/user-attachments/files/27110542/original-commits.zip) or locally with `git reset --hard 4fbddbe8fa56767b2664dc47a5c5737eea6b5dda`, if you want them back for any reason.

harshal24-chavan:
@Caellian 

OK from my side. The changes look good.

Thank you for the refactor and for cleaning everything up!

Caellian:
Np, I planned on fast-forwarding these two commits instead of squashing because they're separate changes and either can introduce bugs on its own. Had I done that without modifying your commit they would've been coupled due to the UB.

This seems good to fast-forward, but I need @brndnmtthws to approve on my commit now :sweat_smile:.

brndnmtthws (approved):

PR #2369, opened 2026-05-23
fix(x11): correct XDBE swap presentation on Xwayland (font thickening + frame repeat)

tmineno:
Two related fixes for XDBE double-buffering under Xwayland (and drivers where `XdbeBackground` is unreliable). Found while running conky as an X11 client on GNOME/Mutter + NVIDIA.

## 1. Back buffer not cleared -> fonts thicken (`ec9bc11`)

In the XDBE path `clear_text()` is a no-op and the back buffer is only ever reset via the `XdbeBackground` swap action. Some drivers don't honor it (notably under Xwayland and certain NVIDIA setups); the back buffer then keeps the previous frame, so antialiased glyphs blend on top of themselves and fonts appear to **thicken over time** -- most visibly when other windows overlap conky and trigger extra redraws.

Fix: clear the back buffer explicitly with `XFillRectangle` after the swap, mirroring the existing XPMDB path.

## 2. Deferred swap -> momentary frame repeat (`b3e0409`)

The swap was left in the X output buffer and flushed lazily. Under Xwayland the compositor can coalesce/reorder it with later requests and momentarily **re-present the previous frame** -- a visible "frame repeat" (e.g. a clock's seconds bouncing back, 40 -> 39 -> 40, at the boundary).

Fix: `XSync` after the swap so the server processes it before conky proceeds, giving a clean, ordered present. A plain `XFlush` only queues the request and is **not** enough (verified empirically: XFlush still bounces, XSync does not).

## Blast radius
- X11 output with `double_buffer = true` on XDBE builds only. XPMDB builds, Wayland output, and `double_buffer = false` are unchanged.
- On drivers where `XdbeBackground` already works, the explicit clear is a redundant reset (identical output).

## Testing
Verified on Xwayland + Mutter (GNOME) + NVIDIA with an ARGB (semi-transparent) `own_window_colour` and `double_buffer = true`: no font thickening on overlap, no frame-repeat at the second boundary.


Caellian:
How does XPMDB work for you?

I'm asking because XPMDB path is much closer to how Wayland side works and also how most GUI toolkits handled double buffering on X11 (not relying on Xdbe ext - even though now it's always provided). I was hoping to remove one or the other and make it so that X11 just uses Cairo if it's included the same way Wayland does (pipeline unification), after Cairo rendering backend is up-to-speed of course.

tmineno:
@Caellian Tested it — built with `-DBUILD_XDBE=OFF` to exercise `use_xpmdb` on the same box (Xwayland + Mutter + NVIDIA, ARGB `own_window_colour`, `double_buffer = true`).

- **Font thickening:** none. XPMDB already clears the back buffer every frame (the path fix #1 mirrors), so nothing accumulates.
- **Flicker:** XPMDB *isn't* immune — I had to (1) `XSync` instead of `XFlush` after the present, and (2) drop the dirty-region clip so the **whole** window is copied each frame (XDBE's swap already presents the whole window; a partial copy leaves the rest of Xwayland's reallocated buffer blank). With both, it's solid.

XPMDB's copy-to-window model maps better onto the Wayland/Cairo path, so +1 on consolidating there. I've added both fixes to the XPMDB path in this PR.

tmineno:
Correction to "solid": longer soak shows a rare residual blank-frame flicker, and it's on the XDBE path too (the shipped build with both fixes) — so not XPMDB-specific. Whole-window present and `XSync` cut the rate; dropping the server-side background didn't help further, and none of it fully eliminated the flicker; two different present paths (Xdbe swap vs pixmap copy) showing the same symptom puts the cause below X11, in Xwayland's async surface-commit → compositor-present. (Couldn't cross-check via native Wayland output — `out_to_wayland` hard-requires `wlr-layer-shell`, which Mutter doesn't implement.) So XPMDB ≈ XDBE here; fully eliminating it likely needs the Wayland/Cairo present path.

Caellian:
[Here's a patch](https://raw.githubusercontent.com/Caellian/mutter-layer-shell/main/.github/patches/add-layer-shell.patch) for Mutter that adds `wlr-layer-shell`. I don't use it, so I never tested it - only that it compiles and looks somewhat reasonable (Claude output).

If you go with it, match the version you already have on your system and apply the patch on top. It might yield better or worse results depending on how extensively you're using conky (cairo backend is not complete).

tmineno:
Follow-up on the cross-check I said I couldn't run. I backported @Caellian's patch to Mutter 46.2 (small delta — Mtk migration and the transaction-based role API are already done in 46), built it, and ran conky `out_to_wayland` inside, then repeated with nested `sway` as a control. Three runs, host Mutter on NVIDIA fixed at the screen end:

| Run | conky's neighbour compositor | renderer | result |
|---|---|---|---|
| 1 | host Mutter via Xwayland | NVIDIA GL | rare flicker (~1/many min) — baseline |
| 2 | backported nested Mutter 46.2 | GL on NVIDIA EGL | rare flicker (~1/many min) |
| 3 | nested sway (`--unsupported-gpu`) | pixman (CPU) | rock-solid |

Runs 2 and 3 used a synthetic seconds-clock config; run 3 was also verified with my real `~/.conkyrc` (only `out_to_wayland` flipped) — same rock-solid result. Conky renders shm via Cairo (CPU) in every run. The variable across 2/3 is the nested compositor's renderer. GL-on-NVIDIA flickers; pixman doesn't.

My earlier "Xwayland-inherent" framing was wrong. The data suggests the flicker correlates with **NVIDIA's GL/EGL surface chain**, not with Xwayland, conky, or layer-shell.

For PR #2369: the XDBE/XPMDB fixes already merged stand. The residual flicker is downstream of conky's present and shouldn't be attributed to Xwayland.

Caveat: host Mutter is still in all three chains; TTY-sway (no host Mutter) would further isolate. Not yet run.

Caellian:
Did you run the patched Mutter with XWayland or Wayland backend on Mutter? I don't think `wlr-layer-shell` would be used anyway if you used XWayland. I was personally mostly interested in how native Wayland with the patch works, but also sent it as an alternative you might prefer over XWayland compatibility layer once native Cairo catches up.

NVIDIA EGL made me recall some Brodie's YT videos: NVIDIA was being incessant on EGLStream and didn't support GBM, this caused painful ecosystem fragmentation on Wayland, with Mutter supporting both paths, but all that + XWayland were extremely flaky and it caused exactly the issues you're solving with your PRs (flickering, stuttering, frame reordering, buffer timing bugs, ...). NVIDIA ultimately caved and added support for GBM which resolved the buffer management issues, and most Wayland compositors now use explicit sync.

However, this was happening 2024-2025; version of Mutter you're daily driving (46.2; mid-2014), and Ubuntu LTS 20.04 in general are version locked right at the start of resolution of these sync issues. This might possibly be resolved on newer Ubuntu (e.g. 25.04). If you have two USB drives, try installing it on one to see. I wouldn't recommend installing newer GNOME/Mutter on older Ubuntu, it could break unless you build your whole session manually (which is more work than just installing a newer Ubuntu).

tmineno:
**Native Wayland**, to answer your question directly. The PR work (XDBE/XPMDB) is all on the XWayland path, but your patch let me add the second leg you were curious about: applied on top of my 46.2 (a few API deltas — the 50.1 patch needed Mtk/logical-monitor adjustments), conky connects to the host Mutter as a real `zwlr_layer_shell_v1` client via `out_to_wayland`, no XWayland in that loop — it comes up, anchors, and renders through the cairo backend. The result that matters: the rare flicker shows on both legs.

Then I chased the explicit-sync angle, and you were right that 46.2 sits right at the start of the NVIDIA sync mess — worse than just being old: Canonical builds `linux-drm-syncobj` straight out of their libmutter (zero syncobj symbols in the packaged binary), even though the driver and DRM caps on this box advertise timeline syncobj + eventfd + `EGL_ANDROID_native_fence_sync` fine. (And it's GBM, not EGLStream — open kernel module.) So I built 46.2 from source with the protocol left in, swapped it onto the host, and the session does start advertising `wp_linux_drm_syncobj_manager_v1`. Short look seemed clean, but a longer soak still flickers — so explicit sync reduces it at best, it isn't the cure here.

To isolate the cause I took the host compositor out of the chain entirely: conky as a layer-shell client under sway on a bare TTY, GLES2 renderer (NVIDIA GL/EGL), scanning out directly via KMS. Clean run, drove the monitor fine, and the rare blank-frame flicker still showed. To be sure it wasn't client-side timing, I also recorded the composited output with `wf-recorder`, and hashed every frame: over a 25-min soak the composited buffer is strictly monotonic — no blank, no stale/reverted frame, even at the exact timestamps I saw a flicker on the glass. So a correct buffer reaches KMS every frame and a bad one still hits the panel.

Taken together, the residual flicker survives implicit and explicit sync, XWayland and native Wayland, Mutter and sway, and only goes away when NVIDIA GL/EGL isn't producing the scanned-out frame (a CPU/pixman compositor is the only fully-clean setup I've found). That lands it in the **NVIDIA GL/EGL → KMS scanout path, below the composite** — underneath both conky and the sync protocol, nothing conky can reach from the client side.

I'll still give 25.04 a go later, but since it reproduces down to bare KMS on the 595 driver I doubt a newer Mutter/Ubuntu moves it much.

One thing I wanted to ask you about: the `out_to_wayland` path looks single-buffered to me — `struct window` keeps one shm cairo surface, and every frame it clears + redraws that same buffer and re-attaches it with no `wl_buffer.release` tracking, so it races the compositor reading the on-screen buffer (on this box that's a *constant* flicker, separate from the rare one — would fit your "cairo backend is not complete"). Is that intentional / something you already know about? I've got a buffer-pool fix prototyped that clears it up, happy to send it as its own PR if that's wanted.

Net for this PR: the XDBE/XPMDB swap is still correct on its own, it just isn't what removes the flicker on this hardware. Thanks for the layer-shell patch and the implicit-sync pointer — that's what got conky onto native Wayland and let me isolate where this actually lives.

Caellian:
@tmineno any followup on this? I see it's been marked as draft for past month with no new changes. Can I close the PR or are you planing on finishing it in future (and don't have time atm)?

PR #2378, opened 2026-05-25
ci: consolidate required checks

brndnmtthws:
## Summary

Consolidates the PR-facing CI jobs into one `CI` workflow and adds a final `All checks passed` job that depends on the individual build, Docker, web, Nix, AppImage, and release jobs. This lets branch protection require one stable status check instead of a list of matrix and workflow-specific checks.

## User Impact

Maintainers can configure one required check for pull requests while still preserving optional job behavior for web/doc-only changes and release tags. Contributors get the same underlying validation, but the merge requirement is simpler and less fragile when CI job names or matrices change.

## Root Cause

CI was spread across separate workflow files, so GitHub branch protection had to track each required check individually. Some workflows were path-filtered or release-only, which made required-check configuration easy to drift from the actual CI surface.

## Fix

Adds a unified `.github/workflows/ci.yaml` workflow with job-level path filtering via `dorny/paths-filter`, keeps skipped optional jobs acceptable in the final gate, and moves final status aggregation into `.github/scripts/check-ci-results.py`. The change also bumps the pinned sccache version, adds `actionlint` and `shellcheck` to mise, and wires actionlint into the Lefthook pre-commit hook.

## Validation

- `mise run actionlint`
- `uv run --with ruff ruff check .github/scripts/check-ci-results.py`
- `uv run --with ty ty check .github/scripts/check-ci-results.py`
- `git diff --check HEAD^ HEAD`
- `lefthook validate`


Caellian (approved):
It's good as is. Given that you asked for a review besides an LLM, I took your review request to mean you want human input. So changes I made are mostly of pedantic nature:
- `detect-ci-jobs.py`: file matchers stored in functions were lifted into data layer (.gitignore-like format), allowing loading from yml files later - `labeler.yml` and this CI code both have "lists of files", labeler action is a bit limited (e.g. don't apply 20+ labels on `refactor` PRs), so there's potential for incremental improvement and unification there later.
- `check-ci-results.py` wasn't needed. GH has `contains` which can check the result lists. Downside is that it's not printing an annotation for failing job, but the graph above the annotations shows same info in Web UI (and is easier to navigate to).

Regarding direction, I prefer python being used for scheduling jobs than having convoluted declarative CI files. In GH ecosystem, ideal solution would probably by a JS action using Node, in a separate repo - but that's a lot more work and introduces JS.

So overall, this is a really good change and will reduce visual noise in PRs.

PR #2382, opened 2026-05-31
fix: wayland double-buffer management and release listener

### LINKED ISSUE #2381: [Feature]: throttle updates by frame output
- [x] Wayland [shm buffer](https://wayland.app/protocols/wayland#wl_shm) management logic is partially implemented. Conky is currently only allocating a single buffer from [`shm_pool`](https://wayland.app/protocols/wayland#wl_shm_pool) and mutating it instead of managing two separate buffers and just swapping them after Wayland commit call.
  
  This shouldn't be guarded by a compile flag - it's a requirement from Wayland. It's a bug currently because after commit the compositors expect they can use the shared memory freely as a static image and conky instead just clears the buffer and starts drawing the new one.

- [x] [`wl_buffer`](https://wayland.app/protocols/wayland#wl_buffer) release listener is missing. Wayland compositors send a [`wl_buffer.release`](https://wayland.app/protocols/wayland#wl_buffer:event:release) event when they're done reading a buffer. This is the signal that it's safe to continue drawing on the buffer that was previously committed.

- [ ] We're not using frame callback ([`wl_surface.frame`](https://wayland.app/protocols/wayland#wl_surface:request:frame)). Without a frame callback, there's no synchronization with compositor vsync. Rendering happens independently of the compositor's frame cycle, meaning we can be drawing frames far more frequently than compositor is able to present. This part is less of an issue given that conky doesn't draw eagerly and usually the update cycle is far slower than what compositor is able to present, but ideal code should wait for `frame` callback.
  - A good example where this might matter is multiple desktops: if conky window is hidden on a separate desktop, compositor will stop issuing `frame` callbacks, and then we can save resources by stopping all system monitoring until it's actually needed. This part requires some larger changes though and addressing just the first two parts of this issue would be a large overall improvement.

Brought up by @tmineno in #2369.

satyvm:
# Checklist
- [] Documentation in `doc/` has been updated

## Description

Partially fixes #2381

### Problem

The Wayland SHM buffer management had two protocol violations:
- **Single buffer mutation after commit** — Only one `wl_buffer` was allocated from the `wl_shm_pool`. After `wl_surface_commit()`, the compositor expects to read the buffer undisturbed, but conky was
  immediately clearing it and drawing the next frame into the same memory. This caused visual glitches and undefined behavior depending on compositor timing.
- **Missing `wl_buffer.release` usage** — A `buffer_release` callback existed but nothing tracked or checked whether the compositor was still reading a buffer. Without honoring `wl_buffer.release`, there
  was no safe synchronization between conky and the compositor.

### Changes
- **Double buffering**: The `window` struct now holds two sets of buffers (`cairo_surface[2]`, `cr[2]`, `layout[2]`, `pango_context[2]`) with a `current_buffer` index. The SHM pool is allocated at 2x size.
  After commit, the active buffer swaps so drawing always targets the buffer the compositor isn't reading.
- **`wl_buffer.release` listener**: Each `wl_buffer` has a registered listener that clears a `busy` flag when the compositor releases it. After committing, `window_commit_buffer()` waits for the next
  buffer to be released before returning.

### What this does NOT address
Frame callback synchronization (`wl_surface.frame`) is not implemented in this PR. As noted in the issue, this is a separate optimization that would require larger architectural changes to the update loop.

### Testing
- Verified the code compiles on macOS (Wayland codepath not exercised at runtime).
- Will test later on a VM, when i find time, will update here.

### DIFF
--- a/src/output/display-wayland.cc
+++ b/src/output/display-wayland.cc
@@ -26,6 +26,7 @@
 
 #include "display-wayland.hh"
 
+#include <spdlog/details/console_globals.h>
 #include <wayland-client.h>
 // #include "wayland.h"
 #include <cairo.h>
@@ -267,7 +268,8 @@ struct window {
   struct wl_shm *shm;
   struct wl_surface *surface;
   struct zwlr_layer_surface_v1 *layer_surface;
-  int scale, pending_scale;
+  int scale = 1;
+  int pending_scale = 1;
   int current_buffer;
   std::shared_ptr<cairo_surface_t> shm_surface[2];
   std::unique_ptr<uint8_t[]> private_buffer;
@@ -326,6 +328,7 @@ void output_scale(void *data, struct wl_output *wl_output, int32_t factor) {
   /* For now, assume we have one output and adopt its scale unconditionally. */
   /* We should also re-render immediately when scale changes. */
   global_window->pending_scale = factor;
+  LOG_TRACE_WITH(({"window->scale", global_window->scale}), "recieved scale event: {}", factor);
 }
 #endif
 
@@ -677,6 +680,9 @@ bool display_output_wayland::main_loop_wait(double t) {
     int fixed_size = 0;
 
     bool scale_changed = global_window->scale != global_window->pending_scale;
+    if (scale_changed)
+      LOG_TRACE("scale changed from {} to {}", global_window->scale,
+                global_window->pending_scale);
 
     /* resize window if it isn't right size */
     if ((fixed_size == 0) &&
@@ -1202,7 +1208,7 @@ static std::shared_ptr<conky::draw_surface> create_shm_surface_from_pool(
 void window_allocate_buffer(struct window *window) {
   assert(window->shm != nullptr);
 
-  int scale = window->pending_scale;
+  int scale = window->scale;
   struct shm_pool *pool;
   pool = shm_pool_create(
       window->shm, data_length_for_shm_surface(&window->rectangle, scale) * 2);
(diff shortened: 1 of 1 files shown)


Caellian (dismissed):
You're currently duplicating `cairo_surface_t` and `cairo_t` which won't work.

I see you have good understanding of browser APIs, so to bridge the gap: `cairo_t` needs to behave like `Canvas2D` context when interacted with on Lua side. If we swap 2 separate contexts internally, pre-draw and post-draw hooks will swap between two different states. e.g. user sets fg color to red once and doesn't reset it, they expect that color to persist across frames; current code will swap between black and red fg color when buffers get swapped.

Solution is to create a **third buffer** to which a single `cairo_surface_t` and `cairo_t` is attached and keep the `shm` stuff separate. You can use `shm_pool` to allocate this buffer if you want but it really doesn't matter much (I don't think), and it can just be a `new uint8_t[w*h*channel]` array. Then, before committing a buffer to Wayland compositor, you just `std::memcpy` this buffer onto the current front `shm` buffer and swap them. `display_output_wayland::drawing_surface` points to this third cairo buffer - that's what gets passed to Lua.

That's... intuitive explanation. The actual change is that you undo most of the changes made and store the two shm buffers in the `global_window` struct separately, create the `cairo_surface_t` that points to the private buffer (`new uint8_t[w*h*channel]` one), and then add logic for copying from the private buffer to front shm one. I expect diff to be much smaller - most of the existing code should stay the same (~ +25 -5).

satyvm:
hey gm @Caellian, can you check updated code?

Caellian:
After a second review I saw some bugs that were there from before the PR so I pushed fixes onto your branch.

7fea21f9e1ee070592c6a76c75f0a6c740db5e02 would be the only missing piece to your PR, but I just pushed the fix to reduce back and fourth.

I squashed your two commits into one given that the second was mostly rewriting the first one. 

Thank you for the PR, it was well written and completes a very important part of wayland support.

Caellian (approved):

PR #2387, opened 2026-06-02
fix(x11): preserve root button propagation

### LINKED ISSUE #2386: [Bug]: 1.24.0 disables root window right click in Openbox
### What happened?

In Openbox, right-clicking in the root window is supposed to show a menu.
With Conky 1.24.0, right clicking does nothing. Reverting to 1.23.0 fixes it.

### Version

1.24.0

Gentoo

### Conky config

```Lua

```

### Stack trace

```GDB

```

### Relevant log output

```Shell

```

brndnmtthws:
## Summary

Preserve X11 root-window button propagation for mouse events that are not over Conky. This fixes #2386.

## User Impact

Openbox users can right-click the root window to open the desktop menu again while Conky is running. Normal application-window clicks and scrolls still avoid duplicate synthetic propagation.

## Root Cause

Conky receives XI2 events from a passive selection on the root window. The previous propagation regression fix consumed every off-Conky event because child-window events are already delivered by the X server, but that also consumed root-targeted button events that window managers handle through the legacy root-click path.

## Fix

Classify XI2 button events whose `child` is `None` as root-button events and leave those eligible for the existing legacy propagation path. Continue consuming other off-Conky XI2 events to avoid duplicate delivery to child windows.

## Validation

- `clang-format --dry-run --Werror src/output/display-x11.cc`
- `mise run build`
- `mise run test`

### DIFF
--- a/src/output/display-x11.cc
+++ b/src/output/display-x11.cc
@@ -638,14 +638,15 @@ bool handle_event<x_event_handler::MOUSE_INPUT>(
     *consumed = llua_mouse_hook(mouse_button_event(
         type, data->pos, data->pos_absolute, button.value(), mods));
   }
-  if (!cursor_over_conky) {
-    *consumed = !should_replay_off_conky_xi_event(*data);
-  }
 #else  /* !BUILD_MOUSE_EVENTS */
   // Events over conky were intercepted by our window; propagate since we have
   // no handler. Events elsewhere are already delivered by the X server.
   if (cursor_over_conky) { *consumed = false; }
 #endif /* BUILD_MOUSE_EVENTS */
+
+  if (!cursor_over_conky) {
+    *consumed = !should_replay_off_conky_xi_event(*data);
+  }
 
   LOG_TRACE("xi event pre-window-type: consumed={}", *consumed);
   if (!own_window.get(*state)) return true;
--- a/src/output/display-x11.cc
+++ b/src/output/display-x11.cc
@@ -433,9 +433,16 @@ static bool is_xi_button_event(const conky::xi_event_data &data) {
   return data.evtype == XI_ButtonPress || data.evtype == XI_ButtonRelease;
 }
 
-static bool is_xi_root_button_event(const conky::xi_event_data &data) {
+static bool is_xi_root_button3_event(const conky::xi_event_data &data) {
   return data.event == data.root && data.child == None &&
-         is_xi_button_event(data);
+         is_xi_button_event(data) && data.detail == Button3;
+}
+
+static bool should_replay_off_conky_xi_event(const conky::xi_event_data &data) {
+  // Off-Conky child-window events are already delivered by the X server.
+  // Replaying them would duplicate input. Root Button3 events are the narrow
+  // compatibility case needed by WMs that use root right-click menus.
+  return is_xi_root_button3_event(data);
 }
 
 template <x_event_handler handler>
@@ -631,12 +638,9 @@ bool handle_event<x_event_handler::MOUSE_INPUT>(
     *consumed = llua_mouse_hook(mouse_button_event(
         type, data->pos, data->pos_absolute, button.value(), mods));
   }
-  // Events not over conky are received via passive XISelectEvents on the root
-  // window; the X server already delivers child-window events to their targets,
-  // so re-propagating those would create duplicates. Root-targeted button
-  // events still need the legacy propagation path for WMs that listen for root
-  // clicks.
-  if (!cursor_over_conky) { *consumed = !is_xi_root_button_event(*data); }
+  if (!cursor_over_conky) {
+    *consumed = !should_replay_off_conky_xi_event(*data);
+  }
 #else  /* !BUILD_MOUSE_EVENTS */
   // Events over conky were intercepted by our window; propagate since we have
   // no handler. Events elsewhere are already delivered by the X server.
--- a/src/output/display-x11.cc
+++ b/src/output/display-x11.cc
@@ -429,6 +429,14 @@ enum class x_event_handler {
   DAMAGE,
 };
 
+bool is_xi_button_event(const conky::xi_event_data &data) {
+  return data.evtype == XI_ButtonPress || data.evtype == XI_ButtonRelease;
+}
+
+bool is_xi_root_button_event(const conky::xi_event_data &data) {
+  return data.child == None && is_xi_button_event(data);
+}
+
 template <x_event_handler handler>
 bool handle_event(conky::display_output_x11 *surface, Display *display,
                   XEvent &ev, bool *consumed, void **cookie) {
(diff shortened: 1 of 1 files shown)


digital-freak:
I tested this on FreeBSD 15-S with Conky 1.23.0 (from ports).
Only 184d4e16a98a02d5f27acc5543b5ab37e9790e7d works correctly: right-clicking on the root window opens the Openbox menu; scrolling over the root window switches virtual desktops.

digital-freak:
> I tested this on FreeBSD 15-S with Conky 1.23.0 (from ports). Only [184d4e1](https://github.com/brndnmtthws/conky/commit/184d4e16a98a02d5f27acc5543b5ab37e9790e7d) works correctly: right-clicking on the root window opens the Openbox menu; scrolling over the root window switches virtual desktops.

However, sometimes, after calling the Openbox menu and left-clicking on a free space on the screen, the menu is displayed again in a new location.

And scrolling switches between the two (or more, depending on the number of Conky processes running) virtual desktops.

The same thing happens with this patch 0704457de281d1ec4b3942a8025e89871accdeea.

digital-freak:
> > I tested this on FreeBSD 15-S with Conky 1.23.0 (from ports). Only [184d4e1](https://github.com/brndnmtthws/conky/commit/184d4e16a98a02d5f27acc5543b5ab37e9790e7d) works correctly: right-clicking on the root window opens the Openbox menu; scrolling over the root window switches virtual desktops.
> 
> However, sometimes, after calling the Openbox menu and left-clicking on a free space on the screen, the menu is displayed again in a new location.
> 
> And scrolling switches between the two (or more, depending on the number of Conky processes running) virtual desktops.
> 
> The same thing happens with this patch [0704457](https://github.com/brndnmtthws/conky/commit/0704457de281d1ec4b3942a8025e89871accdeea).

Conky 1.24.0 - similar behavior.

brndnmtthws:
@digital-freak Thanks for testing this.

Try `6e64d3d0`, which tightens the policy a bit. Still kind of janky but should be less annoying.

src/output/display-x11.cc
```diff
@@ -622,11 +638,10 @@ bool handle_event<x_event_handler::MOUSE_INPUT>(
     *consumed = llua_mouse_hook(mouse_button_event(
         type, data->pos, data->pos_absolute, button.value(), mods));
   }
-  // Events not over conky are received via passive XISelectEvents on the root
-  // window; the X server already delivers them to their targets, so
-  // re-propagating via XSendEvent would create duplicates.
-  if (!cursor_over_conky) { *consumed = true; }
-#else /* !BUILD_MOUSE_EVENTS */
+  if (!cursor_over_conky) {
+    *consumed = !should_replay_off_conky_xi_event(*data);
```
Caellian: Move the check out of `BUILD_MOUSE_EVENTS`.

The `if (cursor_over_conky) { *consumed = true/false; }` here just lowers `BUILD_MOUSE_EVENTS` to runtime and adding checks inside the `#ifdef` here will make the correctness diverge based on build flag state.

Caellian (changes requested):
I glanced at this at the start of the week but didn't manage to review properly until now.

**WIP:** I'll push a commit if needed or approve if I see I'm wrong. The changes to the function are difficult to understand from GH web UI, so I have to clone and test out locally. Ignore comments until then.

## Before PR

|`BUILD_MOUSE_EVENTS`|`ON`|`OFF`|
|-|:-:|:-:|
|`override`; outside|ignored|correct|
|`override`; propagation|1|ignored|
|`normal`; outside|ignored|correct|
|`normal`; propagation|1|ignored|

1 - aggressive right click gets through; propagation is inverted (events outside ignored)

With `BUILD_MOUSE_EVENTS`, when `consumed = true`, `XISetMask` for Motion, Press/Release makes conky steal all events - propagation is needed outside of conky anyway?

digital-freak:
> @digital-freak Thanks for testing this.
> 
> Try `6e64d3d0`, which tightens the policy a bit. Still kind of janky but should be less annoying.

**Well, I tried the suggested patch**

The result is as follows:
The Openbox main menu (right-click) works (details below). Switching virtual desktops only works when hovering over a Conky window.

I'm using FreeBSD 15-stable on my home computer. My desktop environment is XOrg + Openbox. I have four virtual desktops configured in Openbox.

I also have four Conky processes running, each with different settings to display the information I need. The only thing all the configurations have in common is the following:
```
own_window = true,
own_window_class = 'conky_clock',
own_window_type = 'normal',
own_window_color = '#00000000',
```

**Now for the details**

After applying patch 6e64d3d0168c26de3e529bd92b255f88578c0e85 (along with the two previous patches), right-clicking on an empty space on the screen (not occupied by windows) brings up the Openbox menu. Right-clicking on a Conky window has the same effect. However, if only one instance of Conky is running, left-clicking outside the menu itself makes it disappear. If two or more instances of Conky are running, left-clicking (not always, but in most cases) moves the menu to a new location.

**Now about scrolling (which switches virtual desktops)**

If you scroll the mouse wheel over a Conky window, everything works as expected – it switches virtual desktops.

Scrolling in an area not occupied by windows (the root window) does nothing. Without the latest patch (but with the previous two), with a single instance of Conky running, desktops switched as expected—one virtual desktop for each mouse wheel click. With four instances of Conky running (and four virtual desktops), a single mouse wheel click switched between the four desktops, effectively returning to the current one.

(I haven't tried the most recent patch (1a8de4c5d2a38de2d310638889351339802a7a17))

Hope this isn't too long.

digital-freak:
Apologies, I realized I forgot to mention a crucial detail about my environment:

My FreeBSD 15 + Openbox setup runs on a dual-monitor configuration (HDMI-0 primary + DVI-D-0) driven by the NVIDIA proprietary driver. It is configured as a single combined screen with a total resolution of 3840x1080. 

All 4 running Conky windows are positioned exclusively on the second (right) monitor, shifted into the higher coordinate space (with one of them at absolute X: 3505, for example). I figured this specific multi-monitor layout and window grouping might be important for reproducing the scrolling and menu anomalies I described earlier.

Caellian:
I introduced the regression when making a hard switch to XInput from previous handling. This requires a proper sit-down and a lot of debugging between commits since then (~6 hours of work). **1.23.0 is not correct either**, even though the issue says that version works. The last correct version should be assumed to be 1.22.3 because that's prior to #2337.

> If you scroll the mouse wheel over a Conky window, everything works as expected – it switches virtual desktops.

This is what I indicated with the table, it's not just click handling. What's not visible is events like mouse movement that the WM doesn't report back. And the whole logic seems inverted (consumes events only _outside_ conky). So it likely isn't just a single check that needs to be added.

Monkey-patching different clicks will just produce spaghetti that we've been trying to get rid of from X11 code over time. I can't tell what is though without staring at this for a few hours.

I have 2 submissions due by monday, so I'm too short on time to properly fix the issue. I'd need to go through a lot of code I wrote 2 years ago to understand what's even happening. Sadly LLMs won't be much help with this because it requires systemic reasoning and exhaustiveness, it's too sensitive to allow much fuzz - I've attempted today for review and both Opus 3.8 and DeepSeek v4 pro sucked at understanding this code. I'll probably have to revert #2376 and #2337, and redo them more carefully. I relied solely on user confirmation for verification, which is the underlying cause of error (lack of testing).

It's messy, every line has a reason, it's hard to understand and tweak, can't be made any simpler without introducing regressions, and that's why I'm suggesting you hold off on merging this - a fix that asks for 10 more fixes will not make the Gentoo pkg maintainer much happier.

rpaciorek:
this seems like a intentional change in 6e64d3d0168c26de3e529bd92b255f88578c0e85, but as a result it is no longer possible to switch desktops using the mouse wheel on free desktop space in Openbox (on Linux) ... this is an issue in 77fe8192695aa4511eea01d31c717016c33ff99a also and was fixed in 184d4e16a98a02d5f27acc5543b5ab37e9790e7d

Caellian:
Resolved via #2391

PR #2391, opened 2026-06-10
fix(x11): fix cursor event stealing

### LINKED ISSUE #2359: Conky halts on mouse movement, goes on when movement stops
`System: Linux Mint Cinnamon 22.3  with kernel 6.17.0-23-generic`

I have created a conky that runs 2 lua scripts. One script creates a ring, the other creates 3 graph lines. These all react on a cava data stream (on sound).

When i run the conky and i move my mouse around it freezes, when i stop moving the mouse it goes on! 
Exceptions are moving over Telegram and the conky windows itself,, that seems to have no effect. All other windows or desktop it freezes the conky.
I searched in the code, but have no clue what happens here!?? I don't know if this is a bug or not.
Maybe one of you guys can see what happens here or test if the same happens on your system?

Conky version 1.22.4-pre-ea256ffd and 1.19.8 have the same result

Short clip
https://www.cobrasoft.nl/pics/conky-stops-on-mousemove.mp4

Download example
https://www.cobrasoft.nl/download/conky/spectrum-bass/spectrum-bass-v1.1.tar.gz

### LINKED ISSUE #2386: [Bug]: 1.24.0 disables root window right click in Openbox

Caellian:
This is a replacement PR for #2387, which aims to fix issues mentioned in #2386.

# Checklist
- [x] No longer stealing desktop events from the WM
- [x] Accepting and propagating event corretly
  - [x] Passthrough works
  - [x] Lua reporting works.
- [x] Verified with/without `BUILD_MOUSE_EVENTS` on OpenBox
- [x] Verified on another WM (FluxBox)

## LLM Use

LLMs were used for quick sanity checks like, docs/C++ quirks lookups, collaborative guesses what X11 fields mean :smile:, and lastly to verify I didn't swap types/fields somewhere in the `x11-event.cc` file.

I tried using Opus 4.8 and Fable 5 them for analysis of the problem but I'd have to over-specify the problem to get them to produce correct results, so that didn't pan out. All code is written by hand I think.


Caellian:
As of 4th commit, this addresses same concerns as #2387 did.

There's several other related regressions I want to address as well though that were introduced by me removing basic X11 event handling too eagerly (we still relied on some of those even with XI2 enabled). So what's left is just adding XI2 cookie parsing for Enter, Leave and Scroll events.

Tried Antrophic Fable on this just for fun. Identical cluelessness as Opus.

Caellian:
@brndnmtthws ready for review.

Summary is: #2294 broke stuff by removing `(!own)`, then in #2376 I made things weirder by trying to patch over it. #2337 moved to XI2 without checking all old events are still handled by XI2.

- [refactor(mouse-events): cleanup event handling](https://github.com/brndnmtthws/conky/pull/2391/changes/a72464cd109e4142d1977c1174d521fa4714e7a0) 
  - I added `x11-event.cc/h` which is mostly just conversion from runtime `XEvent.type` to discrete event types. It's large, but it's just mappings. I asked Claude Opus to verify mappings are correct here on two separate occasions so they _should_ be.
  - Point of this was to split out that one large mouse event handler into smaller ones so they're easier to reason about.
  - This one closes gaps left behind by #2337.
- [fix(mouse-events): avoid stealing root events if it can be avoided](https://github.com/brndnmtthws/conky/pull/2391/changes/a26e94ac62a5e3bbc40cd1ac9fd942fe3cf971f2)
  - Makes root event capture depend on whether mouse event hook was registered; makes any future breakage affect fewer people
- [fix(mouse-events): fix event edge cases](https://github.com/brndnmtthws/conky/pull/2391/changes/2cb72afb18152619cdf4fb8e1e6bb4d4a93728b6)
  - From what I can tell/recall, these edge cases were never covered and likely caused some bugs.

brndnmtthws (approved):

PR #2396, opened 2026-06-17
Add Nvidia NVML backend

kwillick:
# Checklist

## Description
This set of changes adds the ability to use Nvidia's NVML API instead of the XNVCtrl API. This enables usage of the `nvidia`, `nvidiabar`, `nvidiaguage`, and `nvidiagraph` variables from Wayland.

I was unable to add every feature from XNVCtrl. However, I was able to add features that are unsupported by XNVCtrl.
I documented this in the `nvidia` variable table.

I am not particularly thrilled with how I got `cmake` to download the ```"nvml.h"``` header from Nvidia's servers. See the changes to `ConkyBuildOptions.cmake` to see what I mean.

This fixes/implements #1615.

## Existing behavior
As far as I know, these changes don't modify existing behavior.

## Testing
Caveats are that this has only been tested on my system with my particular GPU, the version of the Nvidia driver I have installed, and the version of the NVML library I have installed. I have been using it for the last week.

Here is the `conky.conf` I used to test this:
```lua
--
-- configurations.
--
--
--
-- For more on Lua, see:
--

conky.config = {
  update_interval = 1, -- update interval in seconds
  double_buffer = true,

  alignment = 'top_right',
  gap_x = 0,
  gap_y = 0,

  minimum_width = 170,
  minimum_height = 1200,
  own_window = true,
  own_window_type = 'normal',

  draw_blended = false,
  own_window_colour = '#FFFFFF',
  draw_borders = false,
  border_width = 1,
  border_inner_margin = 0,
  border_outer_margin = 0,
  draw_graph_borders = false,
  use_xft = true,
  xftalpha = 0,
  draw_shades = false,
  draw_outline = false,

  default_color = 'white',
  font = 'Nimbus Mono PS Regular:size=8',
  -- graph colors
  template1 = [[6b91e5 546fb2]],
};

conky.text = [[
${nvidia driverversion 0}
${nvidia modelname 0}
${nvidia arch}
Mem ${nvidia memmax 0}
Mem free ${nvidia memfree 0}
Mem used ${nvidia memused 0}
Mem util ${nvidia memutil 0}%
GPU temp ${nvidia gputemp 0} °C
Max GPU temp ${nvidia gputempthreshold 0} °C
GPU freq ${nvidia gpufreqcur 0} MHz
GPU min freq ${nvidia gpufreqmin 0} MHz
GPU max freq ${nvidia gpufreqmax 0} MHz
Mem freq ${nvidia memfreqcur 0} MHz
Mem min freq ${nvidia memfreqmin 0} MHz
Mem max freq ${nvidia memfreqmax 0} MHz
PStates ${nvidia perflevelmin} <-> ${nvidia perflevelcur} <-> ${nvidia perflevelmax}
Fan speed ${nvidia fanspeed} rpm
Fan speed ${nvidia fanlevel}%
Power limit ${nvidia powerlimit} W
Power instant ${nvidia powerinstant} W
Power avg ${nvidia poweravg} W
GPU temp${voffset 5}${alignr}${nvidiagraph gputemp 20,170 ${template1} 0}
GPU freq ${voffset 5}${alignr}${nvidiagraph gpufreqcur 20,170 ${template1} 0}
Mem used ${voffset 5}${alignr}${nvidiagraph memused 20,170 ${template1} 0}
Mem free ${voffset 5}${alignr}${nvidiagraph memfree 20,170 ${template1} 0}
Mem util ${voffset 5}${alignr}${nvidiagraph memutil 20,170 ${template1} 0}
Mem freq ${voffset 5}${alignr}${nvidiagraph memfreqcur 20,170 ${template1} 0}
Fan level ${voffset 5}${alignr}${nvidiagraph fanlevel 20,170 ${template1} 0}
Power instant ${voffset 5}${alignr}${nvidiagraph powerinstant 20,170 ${template1} 0}
Power avg ${voffset 5}${alignr}${nvidiagraph poweravg 20,170 ${template1} 0}
GPU temp${voffset 5}${alignr}${nvidiabar gputemp 20,170 0}
GPU freq ${voffset 5}${alignr}${nvidiabar gpufreqcur 20,170 0}
Mem used ${voffset 5}${alignr}${nvidiabar memused 20,170 0}
Mem free ${voffset 5}${alignr}${nvidiabar memfree 20,170 0}
Mem util ${voffset 5}${alignr}${nvidiabar memutil 20,170 0}
Mem freq ${voffset 5}${alignr}${nvidiabar memfreqcur 20,170 0}
Fan level ${voffset 5}${alignr}${nvidiabar fanlevel 20,170 0}
Power instant ${voffset 5}${alignr}${nvidiabar powerinstant 20,170 0}
Power avg ${voffset 5}${alignr}${nvidiabar poweravg 20,170 0}
]]
```


cmake/ConkyBuildOptions.cmake
```diff
@@ -144,13 +144,16 @@ cmake_dependent_option(BUILD_IPV6 "Enable if you want IPv6 support" true
 if(OS_LINUX)
   # nvidia may also work on FreeBSD, not sure
   # NvCtrl requires X11. Should be modified to use NVML directly.
-  dependent_option(BUILD_NVIDIA "Enable Nvidia NvCtrl variables" false
+  dependent_option(BUILD_NVIDIA_X11 "Enable Nvidia variables with NvCtrl" false
```
Caellian: Can't rename this, pulls the rug on what gets built by conky-all packages on some distros (Arch I think).

Use the dependent_option messages and dependencies to communicate it. It's a bit annoying, but people will figure it out or use conky-wayland packages.

cmake/ConkyPlatformChecks.cmake
```diff
@@ -811,7 +812,50 @@ if(BUILD_NVIDIA)
...
+  # We just care about the header, is it different between platforms?
+  # No idea, so print a warning
+  if(NOT ("${CMAKE_SYSTEM_PROCESSOR}" STREQUAL "x86_64"))
+    message(WARNING "This build process is getting the header for x86_64 systems, but yours is ${CMAKE_SYSTEM_PROCESSOR}.")
+  endif()
+  
+  # Find latest version from base url
+  set(NVML_fetched_base_url "https://developer.download.nvidia.com/compute/cuda/redist/")
+  
+  # Execute this shell command to find the latest json file.
```
Caellian: Version has to be predetermined. You're trying to automate things that shouldn't be automated here - your package manager will get the latest version, but linking against an ancient one is fine because C has a stable ABI.

cmake/ConkyPlatformChecks.cmake
```diff
@@ -811,7 +812,50 @@ if(BUILD_NVIDIA)
...
+
+  # Get the json info of that version using another shell command
+  set(NVML_fetched_json_version_url "${NVML_fetched_base_url}${NVML_fetched_json_version}")
+  execute_process(COMMAND sh "-c" "curl -s '${NVML_fetched_json_version_url}' | jq -r '.\"cuda_nvml_dev\".\"linux-x86_64\".\"relative_path\" | tostring'"
+    COMMAND_ECHO STDOUT
+    OUTPUT_STRIP_TRAILING_WHITESPACE
+    OUTPUT_VARIABLE NVML_fetched_path)
+  
+  # Fetch the archive
+  FetchContent_Declare(
```
Caellian: But yes, we'd need this FetchContent thing to get the stub and the header. CMake makes downloading require two configure steps, because it was designed to make junior Bloomberg devs cry "why" into their pillows before sleep.

You can just download nvml stub library and the header into 3rdparty directory + CMakeLists that provides the target and notes the version.

NVIDIA won't mind, they're cool like that (read: they have a stub intermediate library almost exactly for this reason).

cmake/ConkyPlatformChecks.cmake
```diff
@@ -811,7 +812,50 @@ if(BUILD_NVIDIA)
   else(XNVCtrl_INCLUDE_PATH AND XNVCtrl_LIB)
     message(FATAL_ERROR "Unable to find XNVCtrl library")
   endif(XNVCtrl_INCLUDE_PATH AND XNVCtrl_LIB)
-endif(BUILD_NVIDIA)
+endif(BUILD_NVIDIA_X11)
+
+if(BUILD_NVIDIA_NVML)
+  find_library(NVML_LIB NAMES nvidia-ml)
```
Caellian: Then, if we provide/download our own thing, `nvidia-ml` isn't needed. It looks for system-wide libraries.

Given that we'll just provide a copy ourselves, it's better to just make the CMake target link that instead of asking the user to provide it. It leads to e.g. Arch users having to download 4.68GiB cuda package to satisfy this ~80KiB dependency (on Ubuntu).

src/data/hardware/nvidia_nvml.cc
```diff
@@ -0,0 +1,1164 @@
...
+
+  constexpr unsigned int dvsize = NVML_SYSTEM_DRIVER_VERSION_BUFFER_SIZE;
+  init.driver_version.resize(dvsize, 0);
+
+  ret = nvmlSystemGetDriverVersion(init.driver_version.data(), dvsize);
+  if (ret != NVML_SUCCESS) {
+    LOG_ERROR("Unable to get driver version: {}", nvmlErrorString(ret));
+    return;
+  }
+  LOG_INFO("nvidia driver version is {}", init.driver_version);
```
Caellian: debug level - users can access it with variables if they want it, not helpful with most issue reports

src/data/hardware/nvidia_nvml.cc
```diff
@@ -0,0 +1,1164 @@
...
+    return;
+  }
+  init.is_init = true;
+
+  constexpr unsigned int dvsize = NVML_SYSTEM_DRIVER_VERSION_BUFFER_SIZE;
+  init.driver_version.resize(dvsize, 0);
+
+  ret = nvmlSystemGetDriverVersion(init.driver_version.data(), dvsize);
+  if (ret != NVML_SUCCESS) {
+    LOG_ERROR("Unable to get driver version: {}", nvmlErrorString(ret));
```
Caellian: warning level? - specify "NVIDIA driver". Headless system could have an integrated gpu but conky could've been built with nvml, which isn't really an error (wrong use = warning)

src/data/hardware/nvidia_nvml.cc
```diff
@@ -0,0 +1,1164 @@
...
+  if (ret != NVML_SUCCESS) {
+    LOG_ERROR("Unable to get NVML version: {}", nvmlErrorString(ret));
+    return;
+  }
+  LOG_INFO("NVML version is {}", init.nvml_version);
+
+  unsigned int count = 0;
+  ret = nvmlDeviceGetCount_v2(&count);
+  if (ret != NVML_SUCCESS) {
+    LOG_ERROR("Unable to get number of GPUs: {}", nvmlErrorString(ret));
```
Caellian: warning level? - same as before, see [bumblebee setups](https://wiki.archlinux.org/title/Bumblebee) ([PRIME](https://wiki.archlinux.org/title/PRIME)/runtime D3 supersedes this, but ideally should treat all components (even CPUs) as hot-swappable)

src/data/hardware/nvidia_nvml.cc
```diff
@@ -0,0 +1,1164 @@
...
+    return;
+  }
+  LOG_INFO("Found {} GPUs", count);
```
Caellian: debug, "NVIDIA GPUs"

src/data/hardware/nvidia_nvml.cc
```diff
@@ -0,0 +1,1164 @@
...
+
+  // init.devices
+  for (unsigned int i = 0; i < count; i++) { init.devices.push_back(nullptr); }
+}
+
+std::unique_ptr<Device> Device::create(unsigned int gpu_index) {
+  nvmlDevice_t d = nullptr;
+  auto ret = nvmlDeviceGetHandleByIndex_v2(gpu_index, &d);
```
Caellian: Use `nvmlDeviceGetHandleByIndex` instead.

src/data/hardware/nvidia_nvml.cc
```diff
@@ -0,0 +1,1164 @@
...
+  Device* get_or_create_device(unsigned int gpu_index) {
+    init_nvml();
+
+    if (gpu_index >= this->devices.size()) { return nullptr; }
+
+    auto device_ptr = this->devices[gpu_index].get();
+    if (device_ptr) { return device_ptr; }
+
+    auto device = Device::create(gpu_index);
+    this->devices[gpu_index] = std::move(device);
```
Caellian, line 68: Not really a requirement for this PR, but it would be nice if you also stored a `std::map` with keys derived from `nvmlDeviceGetPciInfo` alongside `std::vector` with indices.

Indices aren't perfectly stable across reboots, and the `$nvidia` variable should support PCI bus path so that people with multiple GPUs can specify a more deterministic bus path.

src/data/hardware/nvidia_nvml.cc
```diff
@@ -0,0 +1,1164 @@
...
+  }
+}
+
+void Device::query_pstate_min_max() {
+  // Did we already call this?
+  if (this->pstate_min.was_queried) { return; }
+
+  constexpr unsigned int size = NVML_MAX_GPU_PERF_PSTATES;
+  nvmlPstates_t states[size];
+  memset(states, NVML_PSTATE_UNKNOWN, size);
```
Caellian: `memset` uses byte size as well.

Use <code>[std::array](https://en.cppreference.com/cpp/container/array)::[fill](https://en.cppreference.com/cpp/container/array/fill)</code> which avoids this pitfall:

```c++
std::array<nvmlPstates_t, NVML_MAX_GPU_PERF_PSTATES> states;
states.fill(NVML_PSTATE_UNKNOWN);
```

src/data/hardware/nvidia_nvml.cc
```diff
@@ -0,0 +1,1164 @@
...
+  auto ret =
+      nvmlDeviceGetSupportedPerformanceStates(this->device, states, byte_size);
+  if (ret != NVML_SUCCESS) {
+    this->pstate_min.set_unsupported();
+    this->pstate_max.set_unsupported();
+    return;
+  }
+
+  int max = 0;
+  for (int i = 0; i < size; i++) {
```
Caellian: and then you can use a ranged for loop here

src/data/hardware/nvidia_nvml.cc
```diff
@@ -0,0 +1,1164 @@
...
+  auto ret = nvmlDeviceGetTemperatureV(this->device, &query_temp);
+  if (ret != NVML_SUCCESS) {
+    this->query_max_temps_fail_with_info(ret,
+                                         "Unable to get current temperature");
+    return;
+  }
+  int current = query_temp.temperature;
+
+  // Next get the margin
+  nvmlMarginTemperature_t query_margin = {.version = nvmlMarginTemperature_v1};
```
Caellian: I can't see `nvmlMarginTemperature_v1` in the old header I'm referencing, but is there no `nvmlMarginTemperature`? You keep mixing in the versioned names in places. I'm honestly not sure what's ideal here, but it feels like if we're just passing it into an internal function then the latest version (version erased variant) is easier to maintain.

src/data/hardware/nvidia_nvml.cc
```diff
@@ -0,0 +1,1164 @@
...
+    this->query_max_temps_fail_with_info(
+        ret, "Unable to get current margin temperature");
+    return;
+  }
+  int margin = query_margin.marginTemperature;
+
+  // Now get all the TLIMITs
+  nvmlFieldValue_t queries[4] = {
+      // shutdown_tlimit
+      {.fieldId = NVML_FI_DEV_TEMPERATURE_SHUTDOWN_TLIMIT},
```
Caellian: I wish. [Designated initializers](https://en.cppreference.com/cpp/language/aggregate_initialization#Designated_initializers) (`{.fieldId = NVML_FI_DEV_TEMPERATURE_SHUTDOWN_TLIMIT}`) is a compiler extension until C++20. Won't repeat this comment on other structs, but if version is in the middle of the struct then set the fields after default initialization.

src/data/hardware/nvidia_nvml.cc
```diff
@@ -0,0 +1,1164 @@
...
+
+  // gpu_freq_throttle_temp
+  if (queries[3].nvmlReturn == NVML_SUCCESS) {
+    int gpu_max_tlimit = queries[3].value.siVal;
+    this->gpu_freq_throttle_temp.set(current + margin + (-gpu_max_tlimit));
+
+    LOG_INFO("gpu_freq_throttle_temp: {}", this->gpu_freq_throttle_temp.value);
+  } else {
+    this->gpu_freq_throttle_temp.set_unsupported();
+  }
```
Caellian, line 338: nit: You can dedupe this by iterating over a constexpr array of [pointers to members](https://en.cppreference.com/cpp/language/pointer#pointers_to_members).

src/data/hardware/nvidia_nvml.cc
```diff
@@ -0,0 +1,1164 @@
...
+  } else {
+    this->mem_throttle_temp.set_unsupported();
+  }
+
+  // gpu_freq_throttle_temp
+  if (queries[3].nvmlReturn == NVML_SUCCESS) {
+    int gpu_max_tlimit = queries[3].value.siVal;
+    this->gpu_freq_throttle_temp.set(current + margin + (-gpu_max_tlimit));
+
+    LOG_INFO("gpu_freq_throttle_temp: {}", this->gpu_freq_throttle_temp.value);
```
Caellian: Also, log level should be trace here as it's very detailed info that won't almost ever be needed for debugging and users see variables.

In general, logging levels are too high by a level on many things.

src/data/hardware/nvidia_nvml.cc
```diff
@@ -0,0 +1,1164 @@
...
+    return;
+  }
+
+  this->power_limit.set(query.value.uiVal);
+  LOG_INFO("power limit is: {} milliwatts", this->power_limit.value);
+}
+
+void Device::query_mem_total() {
+  // Did we already call this?
+  if (this->mem_total.was_queried) { return; }
```
Caellian, line 467: I should check whether there's a better way than internally tracking `was_queried` on this many fields. There should be some top-level once guard for variables. I'll let you know.

src/data/hardware/nvidia_nvml.cc
```diff
@@ -0,0 +1,1164 @@
...
+  auto ret = nvmlDeviceGetTemperatureV(this->device, &query_temp);
+  if (ret != NVML_SUCCESS) {
+    LOG_WARNING("Unable to get GPU temperature: {}", nvmlErrorString(ret));
+    return 0;
+  }
+  return query_temp.temperature;
+}
+
+unsigned int Device::get_clock_freq(nvmlClockType_t clock) const {
+  static const char* names[static_cast<size_t>(NVML_CLOCK_COUNT)] = {
```
Caellian: `std::array`

src/data/hardware/nvidia_nvml.cc
```diff
@@ -0,0 +1,1164 @@
...
+
+unsigned int Device::get_video_freq() const {
+  return this->get_clock_freq(NVML_CLOCK_VIDEO);
+}
+
+unsigned int Device::get_gpu_util() const {
+  nvmlUtilization_t util = {};
+  auto ret = nvmlDeviceGetUtilizationRates(this->device, &util);
+  if (ret != NVML_SUCCESS) {
+    LOG_WARNING("Unable to get GPU utilization: {}", nvmlErrorString(ret));
```
Caellian: Once guard warning then trace.

src/core.cc
```diff
@@ -1957,7 +1962,50 @@ struct text_object *construct_text_object(char *s, const char *arg, long line,
...
+#endif /* BUILD_NVIDIA_X11 */
+#ifdef BUILD_NVIDIA_NVML
+  END OBJ_ARG(
+      nvidia, 0,
+      "nvidia needs an argument") if (!create_nvml_query(obj, arg,
+                                                         NVMLQueryType::Text)) {
+    COMMAND_ARG_ERR("nvidia", "nvidia: invalid argument specified: '{}'", arg);
+  }
+  obj->callbacks.print = &print_nvml_value;
+  obj->callbacks.free = &free_nvml_query;
```
Caellian: Place these above X11.

`OBJ_ARG` macros are a huge if-else-if chain. If someone enables both X11 and NVML implementation, NVML (a more general) one will never be reached.

Though I'd handle this at CMake level as well by making X11 one depend on "NOT BUILD_NVIDIA_NVML".

Caellian (changes requested):
Your C++ code is really nice and only needs a few tweaks to make sure it works well everywhere. CMake dependency part needs to be reworked.

I don't think we compile with compiler extensions disabled, but that's an omission.

cmake/ConkyBuildOptions.cmake
kwillick: Oh I didn't realize so many others rely on the names of build options. I can undo that rename.

cmake/ConkyPlatformChecks.cmake
kwillick: I didn't think of just vendoring it in the 3rdparty directory. That will greatly simplify things.

src/data/hardware/nvidia_nvml.cc
kwillick: Oh good catch.

src/data/hardware/nvidia_nvml.cc
kwillick: Oh I guess I am used to newer C++ and gcc just let me get away with it.

src/data/hardware/nvidia_nvml.cc
kwillick: The reason I added `was_queried` across many different fields is that just about every call to NVML can return an error with `NVML_ERROR_NOT_SUPPORTED`.

src/data/hardware/nvidia_nvml.cc
kwillick: Good idea

cmake/ConkyBuildOptions.cmake
Caellian: I'm not sure how used it really is. But [PKGBUILD for Arch](https://gitlab.archlinux.org/archlinux/packaging/packages/conky/-/blob/main/PKGBUILD?ref_type=heads#L95) definitely uses it.

I tend to err on side of caution for these things because I've noticed when I do things like that, it ends up turning into 3/4 issues people open because there was a breaking change.

src/data/hardware/nvidia_nvml.cc
Caellian: I'm :pinching_hand: close to switching to C++20. It would make a lot of things **much** nicer.

kwillick:
@Caellian I think I addressed all your comments.

Caellian (approved):
I switched to .a files and added some minor tweaks to make the code behave well on my (presumably older) Turing GPU.

You did a really amazing job with this, thank you.

kwillick:
Thanks! Those build changes are far more comprehensive.

kwillick:
Oh also feel free to ping me if there are bug reports for this.

PR #2421, opened 2026-07-17
feat: add lua_http_response_hook for customizable HTTP responses

### LINKED ISSUE #2395: [Feature]: Add http_response Lua hook to allow customization of HTTP responses
When using HTTP backend, Conky currently hardcodes a somewhat crude HTML response.

#2394 fixes a deadlock with formatting that was reported in #1649 by introducing a mutex for the string passed to `microhttpd`. This mutex was coincidentally a missing feature that was needed to sync Lua to `sendanswer` used by `microhttpd`.

It's possible now to let users generate the response they want the HTTP backend producing.

It would be ideal if the hook return value was a table that also includes headers besides just the content body - that would allow e.g. JSON responses.

I'm assuming that's what @ctrlcctrlv was proposing in #1649,

sgtanvi:
# Checklist

## Description

Fixes #2395. This PR introduces a `lua_http_response_hook` that allows users to fully customize the HTTP backend's response (body, status, and headers) via a Lua table.

## Problem

Previously, the HTTP backend's response in `display-http.cc` was locked into a hardcoded HTML format. Users wanting to interface with it programmatically had no way to get structured output like JSON, or set custom headers. They'd have to parse HTML.

## What this PR does

**Response Hook**: Added `llua_http_response_hook()` to call Lua and extract body, status, and headers. If the hook isn't set, fails, or returns junk, it falls back gracefully to default responses.

**Thread Safety**: Integrated into `sendanswer()` under `builder_mutex` so Lua runs safely on microhttpd worker threads.

**Fixes & Testing**: Added unit tests covering edge cases. This caught a fatal panic caused by `lua_tostring()` corrupting `lua_next()` iteration on numeric header keys, which is now fixed by explicitly checking for string types.

**Docs**: Updated `config_settings.yaml`.

### DIFF
--- a/src/lua/llua.h
+++ b/src/lua/llua.h
@@ -48,6 +48,13 @@ void llua_init();
 void llua_startup_hook(void);
 void llua_shutdown_hook(void);
 
+// TODO(#2395): declare llua_http_response_hook() here. Should call the
+// configured `lua_http_response_hook` Lua function and read back a table
+// (body/status/headers) so display-http.cc's sendanswer() can build a
+// customized MHD_Response instead of the hardcoded HTML page. Needs a new
+// table-reading path since llua_do_call() callers so far only expect a
+// string (llua_getstring) or number (llua_getnumber) return.
+
 #ifdef BUILD_GUI
 void llua_draw_pre_hook(void);
 void llua_draw_post_hook(void);
--- a/src/lua/llua.h
+++ b/src/lua/llua.h
@@ -30,9 +30,7 @@ extern "C" {
 #include <lualib.h>
 }
 
-#include <string>
-#include <utility>
-#include <vector>
+#include <optional>
 
 #include <config.h>
 #include "../geometry.h"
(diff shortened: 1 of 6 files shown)


src/lua/llua.h
```diff
@@ -48,6 +52,11 @@ void llua_init();
 void llua_startup_hook(void);
 void llua_shutdown_hook(void);
 
+// Calls lua_http_response_hook, returns false if unset or unusable.
+bool llua_http_response_hook(
```
Caellian: I'd return `std::optional<http_hook_result>` here instead of making it write to pointers. Pointers to side, returning a struct makes the signature cleaner if/when more fields are added. The current fields are ok, but later the hook might send extra stuff like system fonts to provide, some sort of pre/postprocessing, etc.

If it's a struct, the llua.h declares it as opaque, and display-http.hh or a new http.hh can own the struct declaration which is only imported from llua.cc where it needs to be constructed. I'd probably place it in display-http.hh for now, it's small enough and can be separated later.

Another minor benefit then is that changing these fields won't change llua.h at all.

src/output/display-http.hh
```diff
@@ -63,6 +65,13 @@ class display_output_http : public display_output_base {
 
 std::string html_escape(const std::string &input);
 
+// Response built from the user's lua_http_response_hook, if configured.
+struct http_response {
```
Caellian, line 69: Oh, there it is. Yeah, make the hook return `std::option<http_response>` with `struct http_response;` predeclaration in llua.h.

Caellian (changes requested):

src/lua/llua.h
sgtanvi: Updated! llua_http_response_hook() now returns std::optional<conky::http_response> instead of writing through the three pointers. 

Struct stays in display-http.hh as you suggested.

src/output/display-http.hh
sgtanvi: Done! Predeclared in llua.h and full struct stays here! Thank you :)

sgtanvi:
@Caellian Hi Tin! Just wanted to check in and see if I need to make any other changes?

Issue #2306, opened 2026-01-15
[Bug]: On KDE wayland, conky hides when clicking on the desktop, with own_window_type desktop


MikeSesketcharba:
### What happened?

When I set own_window_type to desktop, the conky window disappears when I click on the desktop itself. It doesn't stay on.

This has been part of many problems that involve conky staying on the desktop when using SUPER + D. On desktop mode it works, but then it disappears when I click on the desktop. I have tried dock and panel with below in own_window_hints, and I can click on the desktop and the window stays, but on SUPER + D conky disappears with the rest of the windows.

I tried messing with out_to_wayland and out_to_x, toggling one, or toggling both, but I couldn't get it to work either.

Fastfetch output:

```
                  -`                     francisco@gigabyte
                 .o+`                    ------------------
                `ooo/                    OS: Arch Linux x86_64
               `+oooo:                   Kernel: Linux 6.18.4-lqx1-1-lqx
              `+oooooo:                  Packages: 1544 (pacman), 27 (flatpak), 2 (am)
              -+oooooo+:                 Shell: bash 5.3.9
            `/:-:++oooo+:                Display (GS1562): 1920x1080 in 16", 60 Hz [External]
           `/++++/+++++++:               Display (CMN1521): 1920x1080 in 16", 144 Hz [Built-*
          `/++++++++++++++:              DE: KDE Plasma 6.5.4
         `/+++ooooooooooooo/`            WM: KWin (Wayland)
        ./ooosssso++osssssso+`           CPU: 13th Gen Intel(R) Core(TM) i7-13620H (16) @ 4.z
       .oossssso-````/ossssss+`          GPU 1: NVIDIA GeForce RTX 4060 Max-Q / Mobile [Disc]
      -osssssso.      :ssssssso.         GPU 2: Intel UHD Graphics @ 1.50 GHz [Integrated]
     :osssssss/        osssso+++.        Memory: 3.01 GiB / 15.33 GiB (20%)
    /ossssssss/        +ssssooo/-        Swap: 0 B / 4.00 GiB (0%)
  `/ossssso+/:-        -:/+osssso+-      Disk (/): 155.67 GiB / 952.87 GiB (16%) - btrfs
 `+sso+:-`                 `.-/+oso:
`++:.                           `-/+/
.`                                 `/

```

And conky --version:

```
conky 1.22.1-pre-f64e1cf compiled for Linux x86_64

Compiled in features:

System config file: /etc/conky/conky.conf
[...]

PR #2435, opened 2026-08-14
feat: add own_window_namespace config setting for wlr-layer-shell

### LINKED ISSUE #2432: [Bug]: wayland: desktop widgets are hidden by "Show Desktop" on KDE Plasma
### What happened?

On KDE Plasma under Wayland, my conky desktop widgets (`own_window_type = 'desktop'`) disappear the moment I hit **Meta+D** (Show Desktop) — just like any normal window. On X11 the same config stays on screen like wallpaper. The widget *is* mounted on the background layer, but KWin still classifies it as a normal window.

### Reproduction

1. Run conky with `own_window_type = 'desktop'` on KDE Plasma Wayland.
2. Press **Meta+D**.
3. The widget vanishes with all the normal windows.

On X11 the same config produces `_NET_WM_WINDOW_TYPE_DESKTOP` and survives Show Desktop, so this defect is Wayland-only.

### Root cause

wlr-layer-shell's `get_layer_surface` takes a `namespace` argument, defined by the protocol as *"a name identifying the compositor role of the layer surface"*. conky hardcodes `"conky"`:

```cc
conky::create_shell_surface<conky::layer_shell_surface>({
    on_close,
    global_window->surface,
    wl_globals.layer_shell,
    static_cast<uint32_t>(layer_for_window()),
    "conky",
});
```

KWin derives the window type from exactly this field. In `src/wayland/layershellwindow.cpp`:

```cc
static WindowType scopeToType(const QString &scope)
{
    static const QHash<QString, WindowType> scopeToType{
        {QStringLiteral("desktop"), WindowType::Desktop},
        ...
    };
    return scopeToType.value(scope.toLower(), WindowType::Normal);
}
```

`"conky"` isn't in that map, so **every** conky layer surface gets classified `WindowType::Normal` — no matter which layer it sits on. And normal windows are exactly what `Workspace::setShowingDesktop()` hides. So even though the widget is on the `BACKGROUND` layer (selected correctly by `layer_for_window()` at `display-wayland.cc:473`), it still gets hidden.

### Expected behavior

`own_window_type = 'desktop'` on Wayland should behave like its X11 sibling: stay on screen during Show Desktop. The namespace should follow the same logic as the layer mapping.

### Proposed direction

At the layer-shell create site, pick the namespace from the configured window type:

```cc
own_window_type.get(*state) == window_type::DESKTOP ? "desktop" : "conky"
```

Non-desktop conky windows keep the `"conky"` namespace, so nothing else changes.

## References

- wlr-layer-shell spec — `zwlr_layer_shell_v1.get_layer_surface` namespace.
- KWin `layershellwindow.cpp` — `scopeToType()`: unknown scope defaults to `WindowType::Normal`; `"desktop"` → `WindowType::Desktop`.
- KWin `Workspace::setShowingDesktop()` — desktop-type windows are exempt.

## Related existing reports

- #2306 `[Bug]: On KDE wayland, conky hides when clicking on the desktop, with own_window_type desktop` (closed as not planned, `env: xwayland`) — the closest report. It is **not** this bug: the reporter runs the **X11 backend through XWayland** (`out_to_x = true`, `conky: drawing to created window (0x1400001)`), while this report is about the **native Wayland layer-shell backend**. Different path, different root cause, and it was closed without a fix — the native-Wayland case was never addressed.
- #2340 `[Bug]: own_window_type desktop rendering` (open) — same config on Wayland, but a different symptom (double-rendering/black background on XWayland), not the Show-Desktop hiding I'm reporting here.
- #2304 `[Bug]: Openbox hides normal conky window when ToggleShowDesktop is used; desktop window type doesn't work` (closed) — the X11/Openbox analogue of this report, marked wontfix.
- #145 (2015, closed) — KDE/X11 desktop-type behavior; historical context.

I didn't find any open report covering Wayland + KDE + Show-Desktop specifically.

### Version

conky git master (Wayland build, as of `1affe0297` / June 2026), compiled with `-DBUILD_WAYLAND=ON`.

Arch Linux

### Conky config

```Lua
conky.config = {
    own_window = true,
    own_window_type = 'desktop',
    -- no 'above'/'below' hint
}
```

### Stack trace

```GDB
N/A — not a crash; a behavior defect.
```

### Relevant log output

```Shell
Nothing relevant to log — the widget just hides with the windows. The layer/stacking is correct; only KWin's classification is wrong, and that comes from the namespace string. A patched build is needed until this is fixed upstream.
```

sniper1720:
# Checklist

## Description

**The changes:** a new `own_window_namespace` config setting for the `namespace` passed to `get_layer_surface()`. It defaults to `"conky"` (unchanged behaviour for every existing config); a user can opt in to another classification, e.g. `own_window_namespace = 'desktop'` for desktop-widget behaviour. The setting is guarded by `BUILD_WAYLAND` only.

**Why they were necessary:** on KDE Plasma (Wayland), KWin derives the window type from the layer-shell namespace via `scopeToType()`. `"conky"` is not in its map, so every conky surface is classified `WindowType::Normal` regardless of the layer it is on. Normal windows are what `Workspace::setShowingDesktop()` hides, so Show Desktop (Meta+D) hides conky desktop widgets even though they sit on the `BACKGROUND` layer. The setting also lets dock/normal users decouple classification from the layer, e.g. keep `own_window_type = 'normal'` on the bottom layer while KWin still treats the surface as a desktop window.

**How it affects existing behaviour:** nothing changes for existing configs: the namespace stays `"conky"` unless `own_window_namespace` is set explicitly. Configs that set `own_window_namespace = 'desktop'` stay visible during Show Desktop on KDE Plasma, matching the X11 backend.

**How it was tested:** built and shipped as a patched AppImage (`v1.24.3-patched.2`); verified interactively on KDE Plasma 6 (Wayland): desktop widgets stay on screen through repeated Meta+D cycles with `own_window_namespace = 'desktop'`. The X11 backend is untouched.

Closes #2432

### DIFF
--- a/doc/config_settings.yaml
+++ b/doc/config_settings.yaml
@@ -513,7 +513,9 @@ values:
       are treated as a normal window. When unset, 'conky' is used. To keep
       conky visible during Show Desktop while staying above the wallpaper,
       set this to 'desktop' and keep [`own_window_type`](#own_window_type)
-      at 'normal'.
+      at 'normal'. On KDE Plasma, 'dock' is elevated by the slide effect
+      (shouldElevate() in slide.cpp) while switching virtual desktops,
+      so 'desktop' keeps conky in the background throughout.
     default: 'conky'
   - name: pad_percents
     desc: Pad percentages to this many decimals (0 = no padding).
--- a/doc/config_settings.yaml
+++ b/doc/config_settings.yaml
@@ -456,6 +456,19 @@ values:
         decorated.
     args:
       - hint(,hint)*
+  - name: own_window_namespace
+    desc: |-
+      Set the wlr-layer-shell namespace of the Wayland window. Compositors
+      such as KWin classify the surface from this (lowercased); KWin
+      recognizes 'desktop', 'dock', 'notification', 'tooltip',
+      'on-screen-display', 'dialog', 'splash' and 'utility'. Other values
+      are treated as a normal window. When unset, 'conky' is used. To keep
+      conky visible during Show Desktop while staying above the wallpaper,
+      set this to 'desktop' and keep [`own_window_type`](#own_window_type)
+      at 'normal'. On KDE Plasma, 'dock' is elevated by the slide effect
+      (shouldElevate() in slide.cpp) while switching virtual desktops,
+      so 'desktop' keeps conky in the background throughout.
+    default: 'conky'
   - name: own_window_title
     desc: |-
       Allows overriding conky window name.
@@ -504,19 +517,6 @@ values:
     default: normal
     args:
       - (normal|desktop|dock|panel|utility|override)
-  - name: own_window_namespace
-    desc: |-
-      Set the wlr-layer-shell namespace of the Wayland window. Compositors
-      such as KWin classify the surface from this (lowercased); KWin
-      recognizes 'desktop', 'dock', 'notification', 'tooltip',
-      'on-screen-display', 'dialog', 'splash' and 'utility'. Other values
-      are treated as a normal window. When unset, 'conky' is used. To keep
-      conky visible during Show Desktop while staying above the wallpaper,
-      set this to 'desktop' and keep [`own_window_type`](#own_window_type)
-      at 'normal'. On KDE Plasma, 'dock' is elevated by the slide effect
-      (shouldElevate() in slide.cpp) while switching virtual desktops,
-      so 'desktop' keeps conky in the background throughout.
-    default: 'conky'
   - name: pad_percents
     desc: Pad percentages to this many decimals (0 = no padding).
   - name: pop3
--- a/doc/config_settings.yaml
+++ b/doc/config_settings.yaml
@@ -504,6 +504,17 @@ values:
     default: normal
     args:
       - (normal|desktop|dock|panel|utility|override)
+  - name: own_window_namespace
+    desc: |-
+      Set the wlr-layer-shell namespace of the Wayland window. Compositors
+      such as KWin classify the surface from this (lowercased); KWin
+      recognizes 'desktop', 'dock', 'notification', 'tooltip',
+      'on-screen-display', 'dialog', 'splash' and 'utility'. Other values
+      are treated as a normal window. When unset, 'conky' is used. To keep
+      conky visible during Show Desktop while staying above the wallpaper,
+      set this to 'desktop' and keep [`own_window_type`](#own_window_type)
+      at 'normal'.
+    default: 'conky'
   - name: pad_percents
     desc: Pad percentages to this many decimals (0 = no padding).
   - name: pop3
--- a/src/output/display-wayland.cc
+++ b/src/output/display-wayland.cc
@@ -696,6 +696,14 @@ bool display_output_wayland::initialize() {
       !hints_require_layer_shell();
   auto on_close = []() { g_sigterm_pending = 1; };
 
+  // KWin classifies the layer-shell surface from its namespace. Keep the
+  // classic namespace unless the user opts into another via
+  // own_window_namespace, so existing configs are unaffected.
+  std::string namespace_str = own_window_namespace.get(*state);
+  if (namespace_str.empty()) {
+    namespace_str = PACKAGE_NAME;
+  }
+
   if (!hints_layer_shell && wl_globals.layer_shell != nullptr) {
     global_window->shell =
         conky::create_shell_surface<conky::layer_shell_surface>({
@@ -703,9 +711,7 @@ bool display_output_wayland::initialize() {
             global_window->surface,
             wl_globals.layer_shell,
             static_cast<uint32_t>(layer_for_window()),
-            own_window_type.get(*state) == window_type::DESKTOP
-                ? "desktop"
-                : "conky",
+            namespace_str.c_str(),
         });
   } else {
     if (!hints_layer_shell) {
--- a/src/output/display-wayland.cc
+++ b/src/output/display-wayland.cc
@@ -703,7 +703,9 @@ bool display_output_wayland::initialize() {
             global_window->surface,
             wl_globals.layer_shell,
             static_cast<uint32_t>(layer_for_window()),
-            "conky",
+            own_window_type.get(*state) == window_type::DESKTOP
+                ? "desktop"
+                : "conky",
         });
   } else {
     if (!hints_layer_shell) {
--- a/src/output/display-wayland.cc
+++ b/src/output/display-wayland.cc
@@ -700,9 +700,6 @@ bool display_output_wayland::initialize() {
   // classic namespace unless the user opts into another via
   // own_window_namespace, so existing configs are unaffected.
   std::string namespace_str = own_window_namespace.get(*state);
-  if (namespace_str.empty()) {
-    namespace_str = PACKAGE_NAME;
-  }
 
   if (!hints_layer_shell && wl_globals.layer_shell != nullptr) {
     global_window->shell =
--- a/src/output/gui.cc
+++ b/src/output/gui.cc
@@ -189,6 +189,11 @@ conky::simple_config_setting<window_type> own_window_type("own_window_type",
                                                           false);
 #endif /* OWN_WINDOW || BUILD_WAYLAND */
 
+#if defined(BUILD_WAYLAND)
+conky::simple_config_setting<std::string> own_window_namespace(
+    "own_window_namespace", "", false);
+#endif /* BUILD_WAYLAND */
+
 #if defined(OWN_WINDOW) || defined(BUILD_WAYLAND)
 conky::simple_config_setting<uint16_t, window_hints_traits> own_window_hints(
     "own_window_hints", 0, false);
--- a/src/output/gui.cc
+++ b/src/output/gui.cc
@@ -191,7 +191,7 @@ conky::simple_config_setting<window_type> own_window_type("own_window_type",
 
 #if defined(BUILD_WAYLAND)
 conky::simple_config_setting<std::string> own_window_namespace(
-    "own_window_namespace", "", false);
+    "own_window_namespace", PACKAGE_NAME, false);
 #endif /* BUILD_WAYLAND */
 
 #if defined(OWN_WINDOW) || defined(BUILD_WAYLAND)
--- a/src/output/gui.h
+++ b/src/output/gui.h
@@ -224,7 +224,15 @@ extern conky::simple_config_setting<std::string> own_window_class;
 /// @see window_type
 extern conky::simple_config_setting<window_type> own_window_type;
 
+#ifdef BUILD_WAYLAND
+/// @brief Layer-shell namespace hint (Wayland only).
+///
+/// The namespace passed to the wlr-layer-shell protocol. KWin classifies
+/// the surface from this (see scopeToType() in layershellv1window.cpp);
+/// unlisted names like the default "conky" are treated as a normal window.
+/// Other compositors interpret the namespace per their own rules.
 extern conky::simple_config_setting<std::string> own_window_namespace;
+#endif /* BUILD_WAYLAND */
 
 extern priv::colour_setting background_colour;
 
--- a/src/output/gui.h
+++ b/src/output/gui.h
@@ -224,6 +224,8 @@ extern conky::simple_config_setting<std::string> own_window_class;
 /// @see window_type
 extern conky::simple_config_setting<window_type> own_window_type;
 
+extern conky::simple_config_setting<std::string> own_window_namespace;
+
 extern priv::colour_setting background_colour;
 
 Colour get_background_colour_preference(lua::state &l);


Uri-Tauber:
I can confirm the fix is working. 
I had this bug with Conky 1.24.2, when **Peek at Desktop** would remove all windows include Conky. I compiled this PR and it solve the bug. 
Thanks @sniper1720!

If it relevant: I'm using Debian 13 with KDE 6.3.6. Wayland (Obviously).

sniper1720:
> I can confirm the fix is working. I had this bug with Conky 1.24.2, when **Peek at Desktop** would remove all windows include Conky. I compiled this PR and it solve the bug. Thanks @sniper1720!
> 
> If it relevant: I'm using Debian 13 with KDE 6.3.6. Wayland (Obviously).

Thanks for testing and confirming! I hope this PR gets merged soon so I can replace my patched AppImage with the official build for my [custom Conky themes](https://github.com/sniper1720/solus-conky-themes).

Uri-Tauber:
@sniper1720 Your Pure theme looks very good!
Somewhat similar to my own
Conky

Great minds think alike. :smile:

rcspam:
@sniper1720 , @brndnmtthws
This also fixes a second, unrelated-looking bug: other windows snap to conky
while you drag them.

KWin's `canSnap()` (`workspace.cpp`) already skips desktops and docks as snap
targets. But with the hardcoded `"conky"` namespace, `scopeToType()` falls back
to `WindowType::Normal`, so conky is a regular snap target no matter which
layer it sits on. Dragging any window near it makes the window jump and stick
to conky's edges. No `own_window_type` value could work around it, since the
namespace was the same for all of them.

With this PR and `own_window_type = 'desktop'`, KWin classifies conky as
`WindowType::Desktop` and the snapping stops, without touching WindowSnapZone.

One thing I hit while testing further: on KWin the namespace and the layer are
two separate knobs, and deriving one from the other ties them together.

`own_window_type = 'desktop'` requests the `BACKGROUND` layer, which KWin maps
to `DesktopLayer`, the same layer as the plasmashell desktop. Clicking the
desktop raises it above conky and the opaque wallpaper hides it (#2306).
Avoiding that means leaving the `desktop` type, which also gives up the
namespace this PR grants, so Show Desktop breaks again. On KDE you currently
get one or the other.

Decoupling them is a few lines. In `src/output/gui.cc`:

```cc
#ifdef BUILD_WAYLAND
conky::simple_config_setting<std::string> own_window_namespace(
    "own_window_namespace", PACKAGE_NAME, false);
#endif
```

and at the `get_layer_surface` call site the hardcoded `"conky"` becomes the
setting's value. The default is `PACKAGE_NAME`, so nothing changes for anyone
who does not set it.

`own_window_type` then picks the layer, `own_window_namespace` picks the KWin
window type. Running here with `own_window = false` (bottom layer, no struts)
plus `own_window_namespace = 'dock'`: Show Desktop keeps conky on screen,
clicking the desktop keeps conky on screen, normal windows still cover it, and
the maximize area is unchanged.

Tested on Plasma/KWin 6.6.5 Wayland, conky 1.24.2 built from source. Happy to
open a follow-up PR with this if you want it.

sniper1720:
> @sniper1720 Your Pure theme looks very good! Somewhat similar to my own Conky
> 
> Great minds think alike. 😄

Thank you for the kind words, I appreciate it!

sniper1720:
> @sniper1720 , @brndnmtthws This also fixes a second, unrelated-looking bug: other windows snap to conky while you drag them.
> 
> KWin's `canSnap()` (`workspace.cpp`) already skips desktops and docks as snap targets. But with the hardcoded `"conky"` namespace, `scopeToType()` falls back to `WindowType::Normal`, so conky is a regular snap target no matter which layer it sits on. Dragging any window near it makes the window jump and stick to conky's edges. No `own_window_type` value could work around it, since the namespace was the same for all of them.
> 
> With this PR and `own_window_type = 'desktop'`, KWin classifies conky as `WindowType::Desktop` and the snapping stops, without touching WindowSnapZone.
> 
> One thing I hit while testing further: on KWin the namespace and the layer are two separate knobs, and deriving one from the other ties them together.
> 
> `own_window_type = 'desktop'` requests the `BACKGROUND` layer, which KWin maps to `DesktopLayer`, the same layer as the plasmashell desktop. Clicking the desktop raises it above conky and the opaque wallpaper hides it (#2306). Avoiding that means leaving the `desktop` type, which also gives up the namespace this PR grants, so Show Desktop breaks again. On KDE you currently get one or the other.
> 
> Decoupling them is a few lines. In `src/output/gui.cc`:
> 
> ```c++
> #ifdef BUILD_WAYLAND
> conky::simple_config_setting<std::string> own_window_namespace(
>     "own_window_namespace", PACKAGE_NAME, false);
> #endif
> ```
> 
> and at the `get_layer_surface` call site the hardcoded `"conky"` becomes the setting's value. The default is `PACKAGE_NAME`, so nothing changes for anyone who does not set it.
> 
> `own_window_type` then picks the layer, `own_window_namespace` picks the KWin window type. Running here with `own_window = false` (bottom layer, no struts) plus `own_window_namespace = 'dock'`: Show Desktop keeps conky on screen, clicking the desktop keeps conky on screen, normal windows still cover it, and the maximize area is unchanged.
> 
> Tested on Plasma/KWin 6.6.5 Wayland, conky 1.24.2 built from source. Happy to open a follow-up PR with this if you want it.

I'm glad this PR brought out all this potential and even revealed a hidden bug and better concept! Honestly, I didn't catch that issue while testing my Conky themes on KDE Plasma.

The `own_window_namespace` idea is solid and clean. I checked the documentation and codebase, and it doesn't exist in the official Conky config or source yet: the namespace is hardcoded to `"conky"` in `display-wayland.cc` (with no config setting to override it; the only variation is this PR's `desktop`/`conky` conditional derived from `own_window_type`). Since the namespace field is a standard required field in `zwlr-layer-shell-v1`, exposing it as a config option is definitely the right move.

The coupling problem you hit is real, and I ran into it as well. On KDE, `own_window_type = 'desktop'` gets the namespace right for Show Desktop, but it also places the surface on the BACKGROUND layer where the wallpaper hides it. Decoupling them with a separate `own_window_namespace` setting fixes that cleanly.

What I suggest (though I'm open to discussion) is to wait for this PR to land first, and then we can open a follow-up PR adding `own_window_namespace`, defaulting to `"conky"` for backward compatibility.

Thanks for the thorough testing and analysis. I'm happy to collaborate on the follow-up if you're up for it!

sniper1720:
I honestly couldn't be more patient after @rcspam's hint and the whole thread, I went and played with the code directly. I changed a few lines and found a recipe that actually works end to end, tested on real hardware.

**What I changed**

Replaced the hardcoded `desktop ? "desktop" : "conky"` ternary with a configurable `own_window_namespace` setting, defaulting from `own_window_type` so the common cases need no config:

```cpp
std::string namespace_str = own_window_namespace.get(*state);
if (namespace_str.empty()) {
  switch (own_window_type.get(*state)) {
    case window_type::DESKTOP: namespace_str = "desktop"; break;
    case window_type::DOCK:    namespace_str = "dock";    break;
    case window_type::PANEL:   namespace_str = "panel";   break;
    default:                   namespace_str = PACKAGE_NAME; break;
  }
}
```

- Setting declared in `gui.cc` / `gui.h` alongside `own_window_title` / `own_window_class`.
- Docs entry added in [`config_settings.yaml`](https://github.com/brndnmtthws/conky/pull/2435/changes#diff-eca449b4e14d9021c01c7d9db580a2c51d6f4ea4dd501944914d43283839c318) (`own_window_namespace`).

**What I tested, and with what**

Built my fork of conky as a `patched` AppImage and ran it on KDE Plasma Wayland:

- [AppImage build](https://github.com/sniper1720/conky/releases/tag/v1.24.3-patched.2)
- Commit: `247e1700` on the `fix/wayland-desktop-namespace` branch (this PR)

Verified on KWin:
- `own_window_type='normal'` + `own_window_namespace='desktop'` → persists through Show Desktop and stays above the wallpaper.
- `desktop`, `dock`, `panel` defaults derive to matching namespaces.

I'd love your feedback and any suggestions on the approach before calling this merge-ready, especially on whether the type-derived default feels right.

rcspam:
Hi @sniper1720 ,
The setting is exactly what I had in mind, thanks for doing it.

One thing though: I would keep conky as the default. Today every conky uses that namespace. With your default, someone running own_window_type = 'dock' suddenly gets the dock namespace, and KWin starts treating their conky as a dock. Their conky changes behaviour on its own, they did not ask for anything.

Keep conky as the default and nobody is affected. Whoever wants the fix adds one line: own_window_namespace = 'desktop'.

Details, for the record:

- KWin's scopeToType() (layershellv1window.cpp) only knows desktop, dock, critical-notification, notification, tooltip, on-screen-display, dialog, splash and utility. Evormal, so the PANEL case derives a namespace KWinignores.
- The derivation reads own_window_type even when owown_window_type = 'desktop' in an old config, whichX11 ignores in that case, would now pick up the desktop namespace on Wayland.
- The setting is declared under #if defined(OWN_WIN) but only does anything on Wayland, so an X11-onlybuild accepts and ignores it. BUILD_WAYLAND alone would be enough.
- The doc entry says KWin derives the window class  window type.

I have only read the diff, I have not rebuilt your

sniper1720:
> Hi @sniper1720 , The setting is exactly what I had in mind, thanks for doing it.
> 
> One thing though: I would keep conky as the default. Today every conky uses that namespace. With your default, someone running own_window_type = 'dock' suddenly gets the dock namespace, and KWin starts treating their conky as a dock. Their conky changes behaviour on its own, they did not ask for anything.
> 
> Keep conky as the default and nobody is affected. Whoever wants the fix adds one line: own_window_namespace = 'desktop'.
> 
> Details, for the record:
> 
>     * KWin's scopeToType() (layershellv1window.cpp) only knows desktop, dock, critical-notification, notification, tooltip, on-screen-display, dialog, splash and utility. Evormal, so the PANEL case derives a namespace KWinignores.
> 
>     * The derivation reads own_window_type even when owown_window_type = 'desktop' in an old config, whichX11 ignores in that case, would now pick up the desktop namespace on Wayland.
> 
>     * The setting is declared under #if defined(OWN_WIN) but only does anything on Wayland, so an X11-onlybuild accepts and ignores it. BUILD_WAYLAND alone would be enough.
> 
>     * The doc entry says KWin derives the window class  window type.
> 
> 
> I have only read the diff, I have not rebuilt your

Thanks for the review! Keeping `"conky"` as the default was spot on, so I reworked the branch:

* **Removed type-derived default:** Namespace stays `"conky"` unless explicitly set (e.g., `own_window_namespace = 'desktop'`), keeping existing configs unaffected.
* **Wayland-only:** Guarded the setting with `BUILD_WAYLAND`.
* **Docs updated:** Removed `panel` after checking KWin's `scopeToType()` (only 9 recognized namespaces, everything else falls back to Normal).

Validated on KDE Plasma 6 Wayland via a patched AppImage with our themes setting the namespace explicitly. 

Would love your take on the updated commit. Anything you'd change?

rcspam:
@sniper1720 

Looks good to me now, nothing blocking.

One data point worth putting in the docs: with own_window_namespace = 'dock', conky jumped in front of everything while switching virtual desktops, then dropped back. That is KWin's slide effect, shouldElevate() returns isDock() (src/plugins/slide/slide.cpp), so docks get elevated for the duration of the animation. Switching to 'desktop' fixed it here, conky stays in the background throughout. Another reason your doc entry is right to recommend it.

Two nits:

- In gui.h the extern sits inside the #if defined(OWN_WINDOW) || defined(BUILD_WAYLAND) block, while the definition in gui.cc is now under BUILD_WAYLAND alone. Harmless, but the guards should match.
- The extern has no doc comment, unlike own_window_class and own_window_type right above it.

Also, your namespace list is right as is. KWin has a typo in that table, the key reads "crititical-notification" from 6.4 through 6.7 and is fixed only on master, so leaving it out is the accurate call.

sniper1720:
> @sniper1720
> 
> Looks good to me now, nothing blocking.
> 
> One data point worth putting in the docs: with own_window_namespace = 'dock', conky jumped in front of everything while switching virtual desktops, then dropped back. That is KWin's slide effect, shouldElevate() returns isDock() (src/plugins/slide/slide.cpp), so docks get elevated for the duration of the animation. Switching to 'desktop' fixed it here, conky stays in the background throughout. Another reason your doc entry is right to recommend it.
> 
> Two nits:
> 
>     * In gui.h the extern sits inside the #if defined(OWN_WINDOW) || defined(BUILD_WAYLAND) block, while the definition in gui.cc is now under BUILD_WAYLAND alone. Harmless, but the guards should match.
> 
>     * The extern has no doc comment, unlike own_window_class and own_window_type right above it.
> 
> 
> Also, your namespace list is right as is. KWin has a typo in that table, the key reads "crititical-notification" from 6.4 through 6.7 and is fixed only on master, so leaving it out is the accurate call.

Applied all three:

- **Guard:** The `extern` in `gui.h` is now guarded under `#ifdef BUILD_WAYLAND`, matching the definition in `gui.cc`. Kept it Wayland-only since `own_window_namespace` has no X11 parallel (X11 window type comes from `own_window_type`; the namespace is only passed to `zwlr_layer_shell_v1_get_layer_surface`).
- **Doc comment:** Added a `///` comment above the `extern` in the same style as `own_window_title`/`own_window_class`, noting KWin classifies from the namespace (referencing `scopeToType()` in `layershellv1window.cpp`) and that unlisted names like the default `"conky"` fall back to a normal window.
- **Doc entry:** Extended `config_settings.yaml` to include your slide data point: on KDE, `dock` is elevated by `shouldElevate()` while switching virtual desktops, so `desktop` keeps Conky in the background throughout.

@rcspam Thanks again for the review and the working data point on the slide behavior!

rcspam:
Confirmed on my side, all three points are addressed. Nothing blocking.

Two leftovers, both cosmetic:

- The `own_window_namespace` entry in `config_settings.yaml` sits after
  `own_window_type`, but the list is alphabetical (own_window, _class,
  _colour, _hints, _title, _type). It belongs between `own_window_hints`
  and `own_window_title`.
- The default is spelled twice: `""` in `gui.cc`, then the
  `if (namespace_str.empty())` fallback to `PACKAGE_NAME` in
  `display-wayland.cc`, while the doc says `default: 'conky'`. Declaring
  the setting as `("own_window_namespace", PACKAGE_NAME, false)` drops the
  `if`. The only behaviour change is that an explicit
  `own_window_namespace = ''` would then reach the compositor as an empty
  scope instead of falling back to "conky", which is arguably what the
  user asked for anyway.

Also double-checked `scopeToType()` on kwin master: the `.toLower()` your
doc entry mentions is there, and the namespace list matches.

sniper1720:
> Confirmed on my side, all three points are addressed. Nothing blocking.
> 
> Two leftovers, both cosmetic:
> 
>     * The `own_window_namespace` entry in `config_settings.yaml` sits after
>       `own_window_type`, but the list is alphabetical (own_window, _class,
>       _colour, _hints, _title, _type). It belongs between `own_window_hints`
>       and `own_window_title`.
> 
>     * The default is spelled twice: `""` in `gui.cc`, then the
>       `if (namespace_str.empty())` fallback to `PACKAGE_NAME` in
>       `display-wayland.cc`, while the doc says `default: 'conky'`. Declaring
>       the setting as `("own_window_namespace", PACKAGE_NAME, false)` drops the
>       `if`. The only behaviour change is that an explicit
>       `own_window_namespace = ''` would then reach the compositor as an empty
>       scope instead of falling back to "conky", which is arguably what the
>       user asked for anyway.
> 
> 
> Also double-checked `scopeToType()` on kwin master: the `.toLower()` your doc entry mentions is there, and the namespace list matches.

Applied both, confirmed.

- Moved `own_window_namespace` to sit alphabetically between `own_window_hints` and `own_window_title`, and consolidated the default to a single `("own_window_namespace", PACKAGE_NAME, false)` in `gui.cc`, dropping the empty-string fallback in `display-wayland.cc`. The one behavior change matches your note: an explicit `''` now reaches the compositor empty; no theme sets it, so nothing changes in practice.
- Thanks for confirming `scopeToType()` on master; the doc entry matches.

This also confirms the whole feature is scoped correctly: it only affects compositors with `zwlr_layer_shell_v1` (KWin and the wlroots family), and the `conky` default keeps it behavior-neutral everywhere else.

